PR #3, opened 2018-06-21
Added userland files for analog comparator

dverhaert:


dverhaert:
Changing test suite for analog comparator and pulling from another branch.

PR #10, opened 2018-07-11
Add MPU test

cmcavity:
This pull request adds a simple test app to check that MPU regions are set up correctly. It consists of a number of different valid and invalid accesses to flash and RAM addresses, which should succeed or cause an MPU fault, respectively. Currently all of these are commented out, and the tester is able to uncomment the access they want to check. I'm not sure if this is an acceptable format for a test, so any feedback would be welcome.

### DIFF
--- a/examples/tests/mpu/Makefile
+++ b/examples/tests/mpu/Makefile
@@ -1,0 +1,11 @@
+# Makefile for user application
+
+# Specify this directory relative to the current application.
+TOCK_USERLAND_BASE_DIR = ../../..
+
+# Which files to compile.
+C_SRCS := $(wildcard *.c)
+
+# Include userland master makefile. Contains rules and flags for actually
+# building the application.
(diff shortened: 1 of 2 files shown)


brghena:
Could you explain what the difference is between this app and the [`mpu_stack_growth`](https://github.com/tock/libtock-c/tree/master/examples/tests/mpu_stack_growth) and [`mpu_walk_region`](https://github.com/tock/libtock-c/tree/master/examples/tests/mpu_walk_region) apps?

cmcavity:
This app is a way to test accesses to arbitrary locations. It basically just gives a few example valid and invalid acceses for the tester to use. As far as I understand, the other apps test more specific types of accesses: `mpu_stack_growth` tests that a process cannot access memory past the end of its stack, and `mpu_walk_region` tests that a process can make valid reads successfully.

PR #13, opened 2018-08-16
Added test application for the analog comparator

dverhaert:
Test application for the analog comparator which was merged in the main repository a few days ago.

### DIFF
--- a/examples/tests/analog_comparator/main.c
+++ b/examples/tests/analog_comparator/main.c
@@ -1,0 +1,74 @@
+#include <stdio.h>
+#include <stdlib.h>
+
+#include <analog_comparator.h>
+#include <timer.h>
+#include <tock.h>
+static int channel;
+
+static void ac_comparison_polling(uint8_t ac) {
+  uint count = 0;
+  while (1) {
+    count++;
+    bool result = ac_comparison(ac);
+    printf("Try %d. Result = %d.\n", count, result);
+    if (result == 1) {
+      printf("This implies Vinp > Vinn!\n\n");
+    }else {
+      printf("This implies Vinp < Vinn!\n\n");
+    }
+    delay_ms(1000);
+  }
+}
+
+// Callback for AC interrupts. Channel on which the interrupt is generated is
+// passed through here.
+static void ac_cb (__attribute__ ((unused)) int arg0,
+                   __attribute__ ((unused)) int arg1,
+                   __attribute__ ((unused)) int arg2,
+                   __attribute__ ((unused)) void* userdata) {
+  channel = arg0;
+}
+
+static void ac_comparison_interrupt(uint8_t ac) {
+  // Enable AC interrupts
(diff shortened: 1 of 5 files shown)


libtock/analog_comparator.h
```diff
@@ -0,0 +1,33 @@
...
+
+// Does the driver exist?
+int ac_exists(void);
+
+// Request the number of available ACs
+int ac_count(void);
+
+// Compare the voltages of two pins (if one is higher than the other) on the
+// corresponding AC, by polling every second.
+bool ac_comparison(uint8_t);
```
ppannuto: What does the argument mean here? The comment says compares two pins but this only takes on argument?

libtock/analog_comparator.h
```diff
@@ -0,0 +1,33 @@
...
+
+// Enable interrupt-based comparisons
+int ac_enable_interrupts(uint8_t);
+
+// Disable interrupt-based comparisons
+int ac_disable_interrupts(uint8_t);
```
ppannuto: Please document what the enable/disable argument means

libtock/analog_comparator.h
```diff
@@ -0,0 +1,33 @@
+#pragma once
```
ppannuto: General comments on this file:

 - While C doesn't require a variable name for prototype declarations, they are helpful, please add them
 - We generally try to err on the side of explicit / verbose / clear rather than terse. Please change all the `ac_` to `analog_comparator_`

examples/tests/analog_comparator/main.c
```diff
@@ -0,0 +1,74 @@
...
+  }
+  printf("Analog Comparator driver exists with %d channels\n", ac_count());
+
+  // Set mode according to which implementation you want.
+  // mode = 0 --> polling comparison
+  // mode = 1 --> interrupt-based comparison
+  uint8_t mode = 1;
+
+  // Choose a comparator, starting from index 0 and depending on the chip
+  uint8_t ac = 1;
```
ppannuto: Should this use `ac_count` and verify that it's a valid comparator?

dverhaert:
Thanks a lot for the feedback Pat! It has also let me find and correct some other inconsistencies. Ready for another review.

libtock/analog_comparator.h
```diff
@@ -0,0 +1,43 @@
...
+bool analog_comparator_comparison(uint8_t channel);
+
+// Enable interrupt-based comparisons. This will make the AC listen and send an
+// interrupt as soon as Vp > Vn.
+//
+// channel - index of the analog comparator channel, starting at 0.
+int analog_comparator_start_comparing(uint8_t channel);
+
+// Disable interrupt-based comparisons. This will make the AC stop listening,
+// and therewith stop sending interrupts.
```
ppannuto: looks like a typo here :)

PR #14, opened 2018-09-14
UDP Port Binding

hudson-ayers:
### Pull Request Overview

This pull request modifies libtock-c to use FCFS port binding added to Tock in #1165. It allows for userland apps to bind to ports on which they are receiving, giving that app exclusive access to receive on that port. Notably, this port binding is for userland apps only -- kernel apps can still bind to any port, regardless of if a userland port has bound to that port. Further, this PR enforces that apps must be bound to a port in order to send from that port.

This PR also improves the example code in the UDP send and receive apps to better demonstrate all of the functionality available.

In order to use or test this PR, the code must be run alongside the port binding PR in tock (#1165).

### Testing Strategy

I tested this PR by flashing multiple UDP rcv apps onto an Imix and testing that when one app binds to a port, the other cannot. I also tested that two apps can still bind to two different ports. 

Additionally, I tested that the driver call for checking the max possible UDP packet works.

I also tested that the driver works fine for sending packets longer than a single frame. This works now, though it did not before this PR.

I flashed multiple IP send apps onto a single device and tested sending from multiple ports on one device to multiple ports on another device. In order to aid others in replicating this test, I added two example apps - ip_sense2 and udp_rx2 which are near clones of their un-numbered counterparts. These apps merely send/receive to/on different local UDP endpoints. Flashing ip_sense + ip_sense2 on one board and flashing udp_rx + udp_rx2 on another board allows for tests of simultaneous tx/rx.

### Documentation Updated

- [x] Updated the documentation in udp.h to reflect how this library allows for use of the updated driver.

### Formatting

- [x] Ran `make formatall`.


libtock/udp.c
```diff
@@ -14,17 +14,52 @@ static const int SUBSCRIBE_TX = 1;
...
 static unsigned char BUF_TX_CFG[2 * sizeof(sock_addr_t)];
-static unsigned char BUF_RX_CFG[2 * sizeof(sock_addr_t)];
+static unsigned char zero_addr[2 * sizeof(sock_addr_t)];
 
 int udp_socket(sock_handle_t *handle, sock_addr_t *addr) {
   memcpy(&(handle->addr), addr, sizeof(sock_addr_t));
   return TOCK_SUCCESS;
 }
 
+int udp_bind(sock_handle_t *handle, unsigned char *buf_bind_cfg) {
```
cmcavity: Do we need both `udp_socket()` and `udp_bind()`?

hudson-ayers:
Addressed Conor's review. The corresponding PR in tock/tock has been merged, so I am going to merge this as well -- it has been tested on hardware.

PR #15, opened 2018-09-28
Adjust stack and memory break before initializing

bradjc:
We have to make sure there is enough accessible memory for the stack, bss, got and
data sections before trying to initialize them. Right now we're getting
away with it only because the kernel is liberal in letting us use memory
we haven't yet asked for.

### DIFF
--- a/libtock/crt0.c
+++ b/libtock/crt0.c
@@ -63,8 +63,19 @@ void _start(void* app_start,
   // bss_size from start of memory. Also make sure that the stack starts on an
   // 8 byte boundary per section 5.2.1.2 here:
   // http://infocenter.arm.com/help/topic/com.arm.doc.ihi0042f/IHI0042F_aapcs.pdf
-  struct hdr* myhdr = (struct hdr*)app_start;
-  uint32_t stacktop = (((uint32_t)mem_start + myhdr->stack_size + 7) & 0xfffffff8);
+
+  // Before adjusting the stack, make sure to keep all variables in registers.
+  register struct hdr* myhdr = (struct hdr*)app_start;
+  register uint32_t stacktop = (((uint32_t)mem_start + myhdr->stack_size + 7) & 0xfffffff8);
+
+  {
+    register uint32_t heap_size = myhdr->got_size + myhdr->data_size + myhdr->bss_size;
+    memop(0, stacktop + heap_size);
+    memop(11, stacktop + heap_size);
+    memop(10, stacktop);
+    asm volatile ("mov sp, %[stacktop]" :: [stacktop] "r" (stacktop) : "memory");
+    asm volatile ("mov r9, sp");
+  }
 
   // fix up GOT
   volatile uint32_t* got_start     = (uint32_t*)(myhdr->got_start + stacktop);
@@ -94,15 +105,6 @@ void _start(void* app_start,
     } else {
       *target = (*target ^ 0x80000000) + (uint32_t)app_start;
     }
-  }
-
-  {
-    uint32_t heap_size = myhdr->got_size + myhdr->data_size + myhdr->bss_size;
-    memop(0, stacktop + heap_size);
-    memop(11, stacktop + heap_size);
-    memop(10, stacktop);
-    asm volatile ("mov sp, %[stacktop]" :: [stacktop] "r" (stacktop) : "memory");
-    asm volatile ("mov r9, sp");
   }
 
   main();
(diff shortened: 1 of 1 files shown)


dverhaert:
Looks good to me.

libtock/crt0.c
```diff
@@ -63,8 +63,19 @@ void _start(void* app_start,
   // bss_size from start of memory. Also make sure that the stack starts on an
   // 8 byte boundary per section 5.2.1.2 here:
   // http://infocenter.arm.com/help/topic/com.arm.doc.ihi0042f/IHI0042F_aapcs.pdf
-  struct hdr* myhdr = (struct hdr*)app_start;
-  uint32_t stacktop = (((uint32_t)mem_start + myhdr->stack_size + 7) & 0xfffffff8);
+
+  // Before adjusting the stack, make sure to keep all variables in registers.
+  register struct hdr* myhdr = (struct hdr*)app_start;
+  register uint32_t stacktop = (((uint32_t)mem_start + myhdr->stack_size + 7) & 0xfffffff8);
```
ppannuto: I'm pretty sure the register keyword is just a hint (and inhibits taking the variable's memory address). I think if we _actually_ want to force this to live in registers only, we have to write assembly here

libtock/crt0.c
```diff
@@ -63,8 +63,19 @@ void _start(void* app_start,
...
+
+  {
+    register uint32_t heap_size = myhdr->got_size + myhdr->data_size + myhdr->bss_size;
+    memop(0, stacktop + heap_size);
+    memop(11, stacktop + heap_size);
+    memop(10, stacktop);
+    asm volatile ("mov sp, %[stacktop]" :: [stacktop] "r" (stacktop) : "memory");
```
ppannuto: Given that `memop` under the hood will trigger a syscall, and that sycall will cause and interrupt, and the interrupt handler will store things on the stack, I think we should really move this to happen before any function calls.

bradjc:
Someone want to update this?

libtock/crt0.c
alevy: We can't move it before the the call to sbrk (the first `memop`), since that memory isn't accessible to the process yet. We _anyway_ need to ensure that the kernel gives the process enough stack to do a system call (which it does), and that's the maximum we're using here.

That was my thought, anyway.

libtock/crt0.c
alevy: so should we just do an entire `_start0` routine in assembly that allocates the stack+heap and moves the stack, and then just to a `_start1` in C that loads the data section into memory and rewrites the GOT?

Seems safest given your point, right?

libtock/crt0.c
ppannuto: 👍 I'd forgotten about the (fairly obvious) need for a minimum initial stack size. Is that documented anywhere (I'd have guessed here https://github.com/tock/tock/blob/master/doc/Userland.md#application-entry-point, but it's missing)?

I'll look more seriously through the docs later and add it if it's missing.

libtock/crt0.c
ppannuto: Probably. I have some flights tomorrow afternoon/evening. I can take a stab at implementing that then.

ppannuto:
This is updated to do the very initial bit in assembly, to make sure the compiler doesn't try to do anything with the stack prematurely.

The assembly is written to target the m0 and above. It's small enough and rare enough that the complexity/performance of m4-specific code to save a few instructions isn't worth it.

Tested on hail+blink/c_hello/accel_leds

alevy:
OK, I suspect this is gonna be real annoying to port to different architectures, though (e.g. RISC-V). Maybe we'll just deal with it when we need to.

ppannuto:
A combination of curious and an extra 30 minutes before my flight prompted me to see what this would look like for RISC-V. It's more or less the same. There are some other architectural issues we'll need to address for this to be complete,* but this is still a useful start / stub.

*Most notably, RISC-V doesn't specify how to pass syscall number (or syscall arguments, though that's easy enough to keep the same). That is, on arm it's `svc 4`, but risc-v is just `ecall` (no `4`), which means that syscall number has to go into a register in practice. This is a question for another day, but an interesting difference.

alevy:
Um, @ppannuto this is awesome, except can we remove that commit for now. I was just making a note, we shouldn't piggyback untested risc-v assembly on this PR.

ppannuto:
Fair point, I mostly did it as an excuse to play with / learn RISC-V a bit :)

I left the other commit in a branch ( https://github.com/tock/libtock-c/tree/riscv-rt0 ) for whenever that thread picks up.

PR #16, opened 2018-10-17
Simple i2c_master driver. TI Sensors Boostpack example.

lthiery:
Connects to https://github.com/tock/tock/pull/1190

Pseudo-synchronous calls only for now.


lthiery:
Not sure if the TI Sensor BoosterPack example is desirable. I can pop that commit if not - it does add a lot of lines of code but is a useful example of how to hack some off-the-shelf drivers into libtock-c

lthiery:
Will do. I have an improved version that uses buttons and interrupts that I will complete and then write up in the next week or so.

PR #18, opened 2018-10-22
Sensys buggy app

hchiang:
Buggy app for sensys demo that busy waits for a button press to occur, upon which it sends "A button was pressed" message over UDP.

### DIFF
--- a/examples/sensys-buggy-app/main.c
+++ b/examples/sensys-buggy-app/main.c
@@ -2,13 +2,12 @@
 #include <stdio.h>
 
 #include <button.h>
-#include <timer.h>
 
 #include <ieee802154.h>
 #include <udp.h>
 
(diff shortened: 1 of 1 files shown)


bradjc:
👍 

Probably will need a different name specified in the makefile though.

examples/sensys-buggy-app/main.c
```diff
@@ -0,0 +1,100 @@
...
+    dest_addr,
+    16123
+  };
+
+  int len = snprintf(packet, sizeof(packet), "A button was pressed!\n");
+
+  while (1) {
+    // wait for gpio pin to be pressed
+    while (!button_press) {
+      delay_ms(1);
```
alevy: I think this might not do just what we want. `delay_ms` yields to the kernel under the hood, so in practice, the process won't actually exhaust it's timeslice (or consume that much CPU). Instead we probably want to use `button_read` in a tight loop to poll the button.

PR #20, opened 2018-11-05
Cleanups for userland after the SenSys tutorial

ppannuto:
This pulls all of the apps into a dedicated folder. It's possible/likely that at least the `sensys_udp_receive` app might be useful to bubble back up, but I think it's a bit better to leave a copy as-was in `courses/` and then copy it over to `examples/` proper when it's a bit cleaned up and not tied to the webserver idea in the future.

Coupled with this is a fix for the {build/clean/format}_all scripts that had been missing deeper directories, as it was noticed from the depth of directory created here.


ppannuto:
Sigh, as a silly bonus, the old compiler issued a benign/wrong(?) warning with the BLE example.

Rather than fix a non-warning, I just bumped the compiler version on travis to the latest arm-gcc release (it was the 2016 compiler..)

This should be gtg now, even if it is a bit larger in scope than originally planned.

PR #22, opened 2018-11-14
crt0: change data size calculation

bradjc:
Before we assumed that got, data, and bss would be back-to-back-to-back.
However, with compiler aligned buffers that is no longer the case. This
updates the math so that we put the initial brk point in the correct
spot.

I believe this fixes the issue with the rot_* apps.

### DIFF
--- a/libtock/crt0.c
+++ b/libtock/crt0.c
@@ -78,14 +78,17 @@ void _start(void* app_start __attribute__((unused)),
     "movs r5, #7\n"
     "bic  r4, r4, r5\n"         // r4 = (mem_start + myhdr->stack_size + 7) & ~0x7
     //
-    // Compute the heap size
+    // Compute the app data size and where initial app brk should go.
+    // This includes the GOT, data, and BSS sections. However, we can't be sure
+    // the linker puts them back-to-back, but we do assume that BSS is last
+    // (i.e. myhdr->got_start < myhdr->bss_start && myhdr->data_start <
+    // myhdr->bss_start). With all of that true, then the size is equivalent
+    // to the end of the BSS section.
     //
-    // uint32_t heap_size = myhdr->got_size + myhdr->data_size + myhdr->bss_size;
-    "ldr  r5, [r0, #8]\n"       // r5 = myhdr->got_size
-    "ldr  r6, [r0, #20]\n"      // r6 = myhdr->data_size
-    "ldr  r7, [r0, #28]\n"      // r6 = myhdr->bss_size
-    "add  r5, r5, r6\n"         // r5 = got_size + data_size
-    "add  r5, r5, r7\n"         // r5 = got_size + data_size + bss_size
+    // uint32_t appdata_size = myhdr->bss_start + myhdr->bss_size;
+    "ldr  r5, [r0, #24]\n"      // r6 = myhdr->bss_start
+    "ldr  r6, [r0, #28]\n"      // r6 = myhdr->bss_size
+    "add  r5, r5, r6\n"         // r5 = bss_start + bss_size
     //
     // Move registers we need to keep over to callee-saved locations
     "movs r6, r0\n"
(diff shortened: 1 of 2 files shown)


libtock/crt0.c
```diff
@@ -78,22 +78,25 @@ void _start(void* app_start __attribute__((unused)),
     "movs r5, #7\n"
     "bic  r4, r4, r5\n"         // r4 = (mem_start + myhdr->stack_size + 7) & ~0x7
     //
-    // Compute the heap size
+    // Compute the app data size and where initial app brk should go.
+    // This includes the GOT, data, and BSS sections. However, we can't be sure
+    // the linker puts them back-to-back, but we do assume they are in order:
+    // GOT then data then BSS (i.e. myhdr->got_start < myhdr->data_start <
+    // myhdr->bss_start). With all of that true, then the size is equivalent
```
ppannuto: Tracking these bugs down is painful. Can we codify this assumption somewhere and make it part of the TBF standard? Perhaps enforced in the TBF parsing in the kernel or maybe tockloader (or both)

bradjc: This file would be the place to do it, or perhaps wherever we document that 10 word header. TBF explicitly does _not_ care about these details as apps are free to do whatever they want with their bits.

ppannuto: Sigh.. I was afraid you were going to say that.

I think in this case we could probably get away with implementing this check in the `_c_start` function rather than in the assembly, that just needs any stack to be around to be safe, something a touch too small won't burn us there.

bradjc: Ok I added the check to the linker script, so if someone does try to move _got after _bss an error will show up.

alevy: I agree that having it in the linker script is better. In general, I think our contract should be that you take the linker script and crt0 as a unit, and if you change either, you're responsible for both.

So specifically, the linker script in this repo, is a linker script that works with `libtock.a`'s `_start` function. You don't have to use that `_start` symbol if you choose to write your own linker script.

libtock/crt0.c
```diff
@@ -78,22 +78,25 @@ void _start(void* app_start __attribute__((unused)),
...
-    // uint32_t heap_size = myhdr->got_size + myhdr->data_size + myhdr->bss_size;
-    "ldr  r5, [r0, #8]\n"       // r5 = myhdr->got_size
-    "ldr  r6, [r0, #20]\n"      // r6 = myhdr->data_size
-    "ldr  r7, [r0, #28]\n"      // r6 = myhdr->bss_size
-    "add  r5, r5, r6\n"         // r5 = got_size + data_size
-    "add  r5, r5, r7\n"         // r5 = got_size + data_size + bss_size
+    // uint32_t appdata_size = myhdr->bss_start + myhdr->bss_size;
+    "ldr  r5, [r0, #24]\n"      // r6 = myhdr->bss_start
+    "ldr  r6, [r0, #28]\n"      // r6 = myhdr->bss_size
+    "add  r5, r5, r6\n"         // r5 = bss_start + bss_size
```
alevy: As discussed (though not agreed upon) on Slack, I think we need to substract `got_start` here.

Both `bss_start` and `got_start` are defined at link time to be location in memories of the BSS and GOT sections, respectively, _relative to the memory start used for linking_.

Currently, our linker script uses `0x0` as the beginning of RAM, so, in practice, those addresses _will_ be an offset from the start of RAM, regardless of where it starts. However, I don't think this is guaranteed. If `got_start` were guaranteed to be 0, we don't need it in the header in the first place.

For example, the linker _could_ place the GOT at an offset from `0x0`, which would push _everything_ forward by the offset. In our loader (this file), we _do not_ take that into account. We simply take the GOT section and stick it immediately above the stack.

To see this, try sticking a garbage section _in RAM_ right before the got section in the linker script. Alternatively, change the `RAM_START` in the linker script to a non-zero value. If you do that, this will break.

PR #23, opened 2018-11-15
add malloc test app

bradjc:
Runs malloc a few times, and it panics on the buggy version of crt0.


examples/important-client/app2/main.c
```diff
@@ -67,7 +67,7 @@ int main(void) {
 
     // get randomness
     uint32_t rand = 0;
-    int err = rng_sync((uint8_t*)&rand, 4, 4);
+    int err       = rng_sync((uint8_t*)&rand, 4, 4);
```
ppannuto: this looks like it shouldn't be here

PR #25, opened 2018-11-16
crt0: Move the stack pointer before shrinking brk

bradjc:
Before we moved the app break (with memop 0) and then moved the stack pointer. This however means we could be using a stack frame that is outside of the process's available memory. Because of MPU alignment issues this wasn't biting us, until I found an app that was using a much smaller stack (and therefore was moving the app break location much lower, causing the old stack pointer to be way far from the app's available memory).

This attempts to check for the condition where we are moving the brk location down, and moves the stack pointer before doing that.

I'm no assembly expert, so improvements in my assembly changes would be good.

Also I added two simple tests.

### DIFF
--- a/libtock/crt0.c
+++ b/libtock/crt0.c
@@ -94,6 +94,22 @@ void _start(void* app_start __attribute__((unused)),
     "movs r6, r0\n"
     "movs r7, r1\n"
     //
+    // Now we may want to move the stack pointer. If the kernel set the
+    // `app_heap_break` larger than we need (and we are going to call `brk()`
+    // to reduce it) then our stack pointer will fit and we can move it now.
+    // Otherwise after the first syscall (the memop to set the brk), the return
+    // will use a stack that is outside of the process accessible memory.
+    //
+    "add r1, r4, r5\n"          // r1 = stacktop + appdata_size
+    "cmp r1, r3\n"              // Compare `app_heap_break` with new brk
+    "bgt skip_set_sp\n"         // If our current `app_heap_break` is larger
+                                // then we need to move the stack pointer
+                                // before we call the `brk` syscall.
+    "mov  sp, r4\n"             // Update the stack pointer
+    "mov  r9, sp\n"
+    //
+    "skip_set_sp:\n"            // Back to regularly scheduled programming.
+    //
     // Call `brk` to set to requested memory
     //
     // memop(0, stacktop + appdata_size);
--- a/examples/tests/stack_size_test01/Makefile
+++ b/examples/tests/stack_size_test01/Makefile
@@ -1,0 +1,13 @@
+# Makefile for user application
+
+# Specify this directory relative to the current application.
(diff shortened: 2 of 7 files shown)


libtock/crt0.c
```diff
@@ -94,6 +94,22 @@ void _start(void* app_start __attribute__((unused)),
...
     //
+    // Now we may want to move the stack pointer. If the kernel set the
+    // `app_heap_break` larger than we need (and we are going to call `brk()`
+    // to reduce it) then our stack pointer will fit and we can move it now.
+    // Otherwise after the first syscall (the memop to set the brk), the return
+    // will use a stack that is outside of the process accessible memory.
+    //
+    "add r1, r4, r5\n"          // r1 = stacktop + appdata_size
+    "cmp r1, r3\n"              // Compare `app_heap_break` with new brk
+    "bgt skip_set_sp\n"         // If our current `app_heap_break` is larger
```
alevy: i think there is some sort of `itt*` instruction that executes a certain number of instructions if the previous condition holds. It would accomplish the same goal, but save us from adding a label to the global namespace. Maybe @ppannuto knows the syntax and semantics?

ppannuto: Unfortunately the `it` instruction is part of ARM v7-m, aka not available to M0/M0+ cores (the latter being more relevant). I think we should just use a more verbose label `bgt _tock_crt0_skip_set_sp` or whatever

Generally the it instruction is `it_,_,_` where each blank can be {t}hen,{e}lse,{ }ommitted (i.e. any of ttt, tet, et -- can't do `itt t` though, no way to encode "blank in the middle"). In assembly, it looks like:
```
ittee gt
movgt r0 r1
movle r0 r2
suble r0 r0 r1
```
where the extensions on all the instructions covered by the it instruction (this group called the "it block") must match the t/e pattern specified (enforced by the assembler)

ppannuto: Alternatively, if you'd like to avoid a label, you can use the `bx` instruction instead of `b`, which branches to an address in a register (it's how you do long jumps as the encoding of the `b` instruction limits the range). Then you have to get the PC + N in to a register which requires a bit of care, but is do-able.

If you're doing that, it's generally wise to include the explicit width specifiers (aka `mov.n r2 r3` or `mov.w r2 r3`) so it's clear which instructions are 16 or 32 bits wide (they'll all be `.n` to support the M0's in practice)

alevy: OK, never mind :)

alevy:
This is particularly good because it includes a regression test, so even if we're not 100% happy with the style of the assembly, the test is there to help us rewrite it at some later point.

bors r+

libtock/crt0.c
ppannuto: I mean, I would still argue for renaming to something less likely to ever collide / clearly owned by Tock, but otherwise agreed

PR #33, opened 2018-12-05
Update shell.nix

alevy:
Upgrades tockloader, better handling of elf2tab

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -1,0 +1,58 @@
+# Shell expression for the Nix package manager
+#
+# This nix expression creates an environment with necessary packages installed:
+#
+#  * `tockloader`
+#  * arm-none-eabi toolchain
+#  * elf2tab
+#
+# To use:
+#
+#  $ nix-shell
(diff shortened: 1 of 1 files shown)


shell.nix
```diff
@@ -0,0 +1,58 @@
...
+
+with builtins;
+let
+  inherit (pkgs) stdenv;
+  pythonPackages = stdenv.lib.fix' (self: with self; pkgs.python3Packages //
+  {
+
+    tockloader = buildPythonPackage rec {
+      pname = "tockloader";
+      version = "1.3.1";
```
ppannuto: You're probably going to tell me this is antithetical to nix, but does this mean we'd have to bump this number every time tockloader updates? There's no way to express `version = latest`?

alevy: Not that I know of, and I _do_ think it's antithetical to nix  :)

The way to do that would be to have a nix expression in the nix package repo for tockloader and reference that.

PR #39, opened 2019-03-29
Correct elf2tab detection.

jrvanwhy:
libtock-c's app makefile would try to detect whether elf2tab exists, and install it if not. It detected whether elf2tab exists by attempting to feed it a specific set of arguments and looking at whether it displayed output.

Prior to https://github.com/tock/elf2tab/commit/ecbb6039b79397b7c6cab606d34ed8c77faf7b50, elf2tab would spit out a help page with this set of arguments. However, a newer elf2tab will not output anything, causing the Makefile to try to install elf2tab.

This patch attempts to detect elf2tab using the `which` utility rather than calling elf2tab, which will hopefully be a more robust detection mechanism.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ os:
   - linux
 
 rust:
-  - nightly-2018-06-26
+  - stable
 
 before_install:
   - ./.travis-install-gcc
--- a/Configuration.mk
+++ b/Configuration.mk
@@ -34,7 +34,7 @@ TOCK_ARCHS ?= cortex-m0 cortex-m3 cortex-m4
 
 # Check if elf2tab exists, if not, install it using cargo.
 ELF2TAB ?= elf2tab
-ELF2TAB_EXISTS := $(shell $(ELF2TAB) -o k --stack 1 --app-heap 1 --kernel-heap 1 2> /dev/null)
+ELF2TAB_EXISTS := $(shell $(SHELL) -c "command -v $(ELF2TAB)")
 ifndef ELF2TAB_EXISTS
   $(shell cargo install elf2tab)
 endif


jrvanwhy:
It looks like my fix doesn't work in Travis -- it seems to think that elf2tab is unavailable.

Any other ideas on how we can detect it?

brghena:
That's frustrating. I just tried and `which` works fine on MacOS.

brghena:
Actually, the problem in travis is that elf2tab fails to compile. "error[E0658]: use of unstable library feature 'proc_macro' (see issue #38356)", which isn't a problem on my machine. Is Travis using an old version of Rust?

jrvanwhy:
Is it expected for the Makefile to be installing elf2tab? I would've expected elf2tab to be part of the Travis setup, not something dynamically installed in CI.

ppannuto:
As a user convenience, I think it is expected for the build system to
install elf2tab if needed. There was a goal of minimal manual steps to get
going on Tock.

I’m not so surprised that the CI is leveraging / effectively testing that
mechanism rather than manually installing

ppannuto:
Re which on CI, be careful that the default shell used by make in Travis
environment (and some other Linux boxes) is sh not bash. I know on Mac and
I believe on Ubuntu it defaults to bash rather than sh, so things can seem
to work locally and break in CI

jrvanwhy:
Thanks Pat, I did some research and replaced it with a solution that Stack Overflow claims is more portable: https://unix.stackexchange.com/questions/85249/why-not-use-which-what-to-use-then

Regardless, the build is still failing, presumably because of the outdated compiler. Has this not been a persistent issue?

jrvanwhy:
I just altered the rust version in `.travis.yml` and now Travis seems happy with this PR.

jrvanwhy:
There is a rust-toolchain file in elf2tab (set to `stable`), but I don't see how that would impact what toolchain is used with `cargo install`. I don't think `rustup` could determine what toolchain it needs to invoke, and once it starts running `cargo` it's too late to switch toolchains.

brghena:
Yeah. I've run into this issue before. I don't use Rust for anything other than Tock, which manages which version is installed for its directory only. So when I'm in other repositories, which version of Rust I'm on is pretty random (2017-02-24 in this case). A `rustup default stable` seems to solve that. I wonder if we should suggest such things in a README somewhere.

alevy:
>  I don't think rustup could determine what toolchain it needs to invoke, and once it starts running cargo it's too late to switch toolchains.

I think we can start using the `edition` field in Cargo.toml to help? That's stable now

jrvanwhy:
> > I don't think rustup could determine what toolchain it needs to invoke, and once it starts running cargo it's too late to switch toolchains.
> 
> I think we can start using the `edition` field in Cargo.toml to help? That's stable now

`edition` is for Rust editions, which are independent of (orthogonal to) toolchain versions. Regardless, `cargo install` doesn't see the crate's source until well after `rustup` selects the toolchain to run.

As-is, I don't think `cargo install` can be reliable unless the user is using a recent `cargo` -- in general it cannot install binaries that use Rust features newer than itself.

PR #40, opened 2019-05-07
make: support multiple toolchains

bradjc:
Before we assumed that `arm-none-eabi` would be able to compile code for all Tock architectures. With riscv that is no longer the case. This makes the toolchain a variable as well.

I definitely made this quickly and wanted to get it into its own branch because we need something like it to support riscv in the future.

I'm also realizing that `riscv64-unknown-elf` isn't going to be a simple swap in, at least because of PIC. So that is going to take some hacking, but I wanted to at least get a first draft of a mechanism to support different toolchains for different architectures in a pr.

### DIFF
--- a/AppMakefile.mk
+++ b/AppMakefile.mk
@@ -47,7 +47,7 @@ ifneq "$$(wildcard $(1)/include)" ""
 endif
 
 # Add arch-specific rules for each library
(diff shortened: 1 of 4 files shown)


alevy:
CMake would/will help with this.... a lot :)

bradjc:
I'm sure Pat will not like my wholesale commenting out of old compiler checks!

PR #41, opened 2019-05-26
Remove use of yield() in test applications

phil-levis:
Calling yield() directly assuming a particular callback is bug-prone. If the runtime registers its own callbacks, then the application might assume a callback has occurred when it hasn't. This patch replaces all of the calls to yield() into calls to yield_for().

### DIFF
--- a/examples/tests/analog_comparator/main.c
+++ b/examples/tests/analog_comparator/main.c
@@ -27,7 +27,7 @@ static void analog_comparator_callback (__attribute__ ((unused)) int arg0,
                                         __attribute__ ((unused)) int arg1,
                                         __attribute__ ((unused)) int arg2,
                                         __attribute__ ((unused)) void* userdata) {
-  callback_channel = arg0;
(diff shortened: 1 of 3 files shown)


ppannuto:
We've been hit by various flavors of this `yield` vs `yield_for` bug before.. should we just remove `yield` completely in favor of `yield_for`? After this patch, the only remaining instances of `yield` (in code) in this repo are:

    examples/security_app/main.c:62:    yield();
    examples/tests/gpio_interrupt/main.c:24:    yield();
    libtock/alarm_timer.c:201:    yield();
    libtock/crt0.c:187:    yield();

alevy:
We should not remove `yield` completely, you need it to build lower layer runtimes. Could make sense (as I've suggested in the past) to separate the low level user interface from the synchronous appearing runtime, and only have `yield` in the former.

alevy:
(I meant from the library, not the example/test applications. we can remove completely from example apps)

PR #45, opened 2019-06-11
Remove legacy libs

bradjc:
We had the comment `# Single-arch libraries, to be phased out` in libtock-c makefile, I'm not sure exactly what the phase out was dependent on. I removed them and replaced them with `-lc -lgcc -lm -lstdc++` and all applications build.


bradjc:
Eh, ok `accel-leds` no longer works...

Is this a PIC issue?

ppannuto:
A touch late to the game on this, but the "single-arch libraries" comment
was a reference to the fact that we build newlib for cortex-m0 as that's
upward compatible through the m4. Ideally we'd build it for each
sub-architecture, as I think that gains some non-trivial space wins on the
m4, but clearly not the top priority

bradjc:
Gotcha. Well the RISC-V stuff starts the move toward specific arch libraries, and that could be fleshed out more to have different libraries for each cortex-m variant.

PR #49, opened 2019-07-25
riscv support

DieracDelta:
This is a couple of minor changes (before I forget) to get c_hello to run on qemu emulating the hifive1 board. The idea here is that if you clone this and compile, it'll just work. Most of this came out of a discussion with brad last week.

1) I couldn't get the cpp library to compile. Has anyone had success with that?
2) We need to save app_start in a0 (first argument) before overwriting s0 in crt0. This allows app_start to be passed to _c_start instead of the frame pointer. 
3) Due to issues with PIC, c_hello is finding an incorrect address for the char array. Need to move it inside main() to prevent this.

### DIFF
--- a/Configuration.mk
+++ b/Configuration.mk
@@ -88,9 +88,6 @@ override CPPFLAGS_cortex-m0 += $(CPPFLAGS_cortex-m) \
 override LEGACY_LIBS_cortex-m += \
       $(TOCK_USERLAND_BASE_DIR)/newlib/cortex-m/libc.a\
       $(TOCK_USERLAND_BASE_DIR)/newlib/cortex-m/libm.a\
-      $(TOCK_USERLAND_BASE_DIR)/libc++/cortex-m/libstdc++.a\
-      $(TOCK_USERLAND_BASE_DIR)/libc++/cortex-m/libsupc++.a\
-      $(TOCK_USERLAND_BASE_DIR)/libc++/cortex-m/libgcc.a
 
 override LEGACY_LIBS_cortex-m4 += $(LEGACY_LIBS_cortex-m)
 override LEGACY_LIBS_cortex-m3 += $(LEGACY_LIBS_cortex-m)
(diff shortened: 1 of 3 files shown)


ppannuto:
> I couldn't get the cpp library to compile. Has anyone had success with that?

Libtock grabs a snapshot of newlib, it looks like the current version is from April 2017 (see the Makefile in the newlib folder) -- it's quite possible that risc-v support was incomplete at that point. Maybe try updating that?

bradjc:
> Libtock grabs a snapshot of newlib, it looks like the current version is from April 2017 (see the Makefile in the newlib folder) -- it's quite possible that risc-v support was incomplete at that point. Maybe try updating that?

It's likely that we would need a version from one of the riscv teams.

Configuration.mk
```diff
@@ -88,9 +88,6 @@ override CPPFLAGS_cortex-m0 += $(CPPFLAGS_cortex-m) \
 override LEGACY_LIBS_cortex-m += \
       $(TOCK_USERLAND_BASE_DIR)/newlib/cortex-m/libc.a\
       $(TOCK_USERLAND_BASE_DIR)/newlib/cortex-m/libm.a\
-      $(TOCK_USERLAND_BASE_DIR)/libc++/cortex-m/libstdc++.a\
-      $(TOCK_USERLAND_BASE_DIR)/libc++/cortex-m/libsupc++.a\
-      $(TOCK_USERLAND_BASE_DIR)/libc++/cortex-m/libgcc.a
```
bradjc: Does this need to change for some reason?

DieracDelta: Oops, no that shouldn't be deleted.

libtock/crt0.c
```diff
@@ -215,11 +215,10 @@ void _start(void* app_start __attribute__((unused)),
     "ecall\n"                   // memop
     //
     // Setup initial stack pointer for normal execution
-    "mv   sp, s1\n"             // sp = stacktop
-    "mv   s0, sp\n"             // Set the frame pointer to sp.
-    //
     // Call into the rest of startup. This should never return.
+    "mv   sp, s1\n"             // sp = stacktop
     "mv   a0, s0\n"             // first arg is app_start
+    "mv   s0, sp\n"             // Set the frame pointer to sp.
```
bradjc: It would be great to add this change to `master`.

DieracDelta: I'm just going to manually submit a pr to master.

PR #53, opened 2019-09-25
add pre-built archival versions of apps

ppannuto:
Useful for backwards compatabililty testing, also a possibly a good place for
tutorials to grab stable images of apps from


alevy:
I don't think these are good to have _in the git repo/history_.

Instead, we host compiled `tab`s on the main website (https://github.com/tock/tock-www/tree/master/assets/tabs) where `tockloader` also knows where to find them.

We've also added them to a release for `tock` once, i.e.. https://github.com/tock/tock/releases/tag/release-1.1-2018-04

I'm personally partial to the website.

ppannuto:
I'm a little concerned about the discoverability, that's why I put them in the repo proper. It's not exactly massive repo bloat I don't think (2.2 MB total), but I also don't feel that strongly. Old 1.0 apps are really only useful during regression-testing for releases I think.

I made a 1.0 release and added them as assets: https://github.com/tock/libtock-c/releases/tag/release-1.0-2018-02

Not opposed to adding to the website as well if you think that's useful for anything.

PR #55, opened 2019-10-01
program.mk: update as tockloader has changed

### LINKED ISSUE #54: Program.mk out of sync with Tockloader
The Makefiles in each app still link to `Program.mk` which uses specifies methods of loading code onto boards. It looks like that file has gotten quite a bit out of sync with Tockloader.

It still uses the `--jtag` flag, which Tockloader warns is deprecated.

It also refers to `nrf52dk.py` for programming the nrf52dk, which doesn't seem to work. While `tockloader install --jlink --board nrf52dk` works just fine.

bradjc:
This updates the flashing commands in Program.mk since Tockloader has been updated.

Fixes #54

### DIFF
--- a/Program.mk
+++ b/Program.mk
@@ -16,22 +16,22 @@ endif
 #
 ifeq ("$(TOCK_BOARD)","hail")
 PROGRAM = $(TOCKLOADER) $(TOCKLOADER_GENERAL_FLAGS) install $<
-FLASH = $(TOCKLOADER) $(TOCKLOADER_GENERAL_FLAGS) install --jtag $<
+FLASH = $(TOCKLOADER) $(TOCKLOADER_GENERAL_FLAGS) install --jlink $<
 
 else ifeq ("$(TOCK_BOARD)","imix")
 # Change program region offset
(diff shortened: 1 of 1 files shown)


Program.mk
```diff
@@ -16,22 +16,22 @@ endif
 #
 ifeq ("$(TOCK_BOARD)","hail")
 PROGRAM = $(TOCKLOADER) $(TOCKLOADER_GENERAL_FLAGS) install $<
-FLASH = $(TOCKLOADER) $(TOCKLOADER_GENERAL_FLAGS) install --jtag $<
+FLASH = $(TOCKLOADER) $(TOCKLOADER_GENERAL_FLAGS) install --jlink $<
 
 else ifeq ("$(TOCK_BOARD)","imix")
 # Change program region offset
 TOCKLOADER_INSTALL_FLAGS += --app-address 0x40000
```
brghena: Does Imix still use a different address for apps too?

bradjc: Yes, although I think it most cases that flag is not required as the value is in the bootloader args.

PR #69, opened 2020-02-07
userland_generic.ld: Include .sdata in the data section

alistair23:
If an ELF has a .sdata section it won't be copied to the TBF file. To
ensure that it makes it in let's include it in the data section.

### DIFF
--- a/userland_generic.ld
+++ b/userland_generic.ld
@@ -108,6 +108,10 @@ SECTIONS {
         . = ALIGN(4); /* Make sure we're word-aligned here */
         _data = .;
         KEEP(*(.data*))
+	/* Include the "small data" in the data section. Otherwise it will be
+	 * dropped when the TBF is created.
+	 */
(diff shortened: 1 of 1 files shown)


alistair23:
@gkelly I'm not sure what you mean?

sdata seems to just be [small data](https://embarc.org/toolchain/gcc/small-data.html) it's not RISC-V specific. Do you want me to add a comment saying that it's small data?

ppannuto:
Huh, interesting; from a quick skim of that doc, looks like a new gcc thing
circa ~late 2017, TIL.

If you wouldn’t mind still adding a comment and pointer, we err pretty hard
on the side of doing anything possible to make the linker scripts easier to
read and maintain. Thanks!

gkelly:
Oh yeah this is generic, sorry @alistair23.

alistair23:
I added a comment

PR #72, opened 2020-03-04
LCD 16x2 library HD44780

teonaseverin:
Created an LCD 16x2 library for HD44780 based on the gpio library and an example using it.


bradjc:
This is really cool! Can you make two changes?
1. Move the lcd_1602.c/h files into the app folder. Libtock is just shims on top of syscalls, and so far we have been storing all higher-level library code in application folders.
2. Add a README.md describing the hardware you are using and a brief description of what should happen when the app runs.

Otherwise looks good!

examples/lcd_16x2/README.md
```diff
@@ -0,0 +1,24 @@
...
+Datasheet for the LCD: https://www.sparkfun.com/datasheets/LCD/HD44780.pdf 
+
+Supported Boards
+-----------------
+- It should support any board that has GPIO support. The communication is being done 
+only through GPIO pins. 
+
+Tested Boards
+-----------------
+- NUCLEO-F429ZI
```
bradjc: Can you add a link to the LCD model, and which GPIO pins you connected? I just want to make sure we record that to make it easier to test this in the future.

teonaseverin:
Hey guys!

Do you have any feedback here?

examples/lcd_16x2/lcd_1602.h
```diff
@@ -0,0 +1,116 @@
+/*
+	lcd_1602.h - C implementation of the LiquidCrystal class, providing basic
+	API for communicating with the 16x2 HD44780 LCD; 
+
+	The code was developed based on the basic Arduino Library for the LCD: 
+		https://github.com/arduino-libraries/LiquidCrystal.git
```
ppannuto: :(  :(

I don't think we can take this code unfortunately. My understanding is that this is a derivative work of a GPL-licensed driver, and pulling that in would require libtock-c to become GPL as well.

@alevy are actually going to confer with some software licensing professionals next week, so I'll try to ask about this specific case in case I'm wrong, but I think we would need to ask the upstream library to re-license as Apache or MIT or something equally permissive.

ppannuto: Unfortunately, I think it's confirmed that we need to either ask the upstream library to re-license a release for us to base a derivative work on or we'll have to reimplement this from scratch from datasheets.

ppannuto:
I'm going to close this because we unfortunately can't take this code as-is. If the upstream author is willing to re-license, or there's a clean implementation, we're certainly happy to take the driver.

For folks coming across this without licensing concerns, this will certainly still work compiled into your own applications, etc.

PR #74, opened 2020-03-11
Fix wrong header handling in ble passive scanning example

TobiasSchaffner:
* Only read the first 4 bits instead of the first byte to get the PDU type.
* Read out TxAdd and RxAdd bytes correctly.
* Some improvements on printouts and some additional constants.

### DIFF
--- a/examples/ble_passive_scanning/advertisement.cpp
+++ b/examples/ble_passive_scanning/advertisement.cpp
@@ -11,7 +11,9 @@ Advertisement::Advertisement(const unsigned char* buf, int len)
   std::memcpy(&header_, &buf[HEADER_START], HEADER_SIZE);
   std::memcpy(&address_, &buf[ADDRESS_START], ADDRESS_SIZE);
   unsigned char data_len = len - HEADER_SIZE - ADDRESS_SIZE;
-  if (data_len > DATA_MAX_SIZE) data_len = DATA_MAX_SIZE;
+  if (data_len > DATA_MAX_SIZE) {
+    data_len = DATA_MAX_SIZE;
+  }
   std::memcpy(&data_, &buf[DATA_START], data_len);
 }
 
@@ -37,8 +39,9 @@ void Advertisement::print() const
   printf("Address: %02x %02x %02x %02x %02x %02x\r\n", address_[5],
          address_[4], address_[3], address_[2], address_[1], address_[0]);
   printf("Data: ");
-  for (int i = 0; i < pduLength() - ADDRESS_SIZE; i++)
+  for (int i = 0; i < pduLength() - ADDRESS_SIZE; i++) {
     printf("%02x ", data_[i]);
+  }
   printf("\r\n\r\n");
 }
 
@@ -84,7 +87,7 @@ const char* Advertisement::pduTypeStr() const
   }
 }
 
-bool Advertisement::validAdvertisement(const unsigned char* buf, int len)
+bool Advertisement::checkScanResult(const unsigned char* buf, int len)
 {
   if (buf == nullptr) {
     printf("Malformed scan result: Buffer was null!\n");
--- a/examples/ble_passive_scanning/advertisement.cpp
+++ b/examples/ble_passive_scanning/advertisement.cpp
@@ -47,7 +47,7 @@ void Advertisement::print() const
 
 unsigned char Advertisement::pduLength() const
 {
-  return header_[1] & PDU_ADDRESS_HEADER_MASK;
+  return header_[1] & PDU_LEN_HEADER_MASK;
 }
 
 unsigned char Advertisement::pduType() const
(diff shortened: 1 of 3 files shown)


examples/ble_passive_scanning/advertisement.h
```diff
@@ -2,31 +2,41 @@
...
+const unsigned char HEADER_SIZE = 2;
+const unsigned char ADDRESS_START = 2;
+const unsigned char ADDRESS_SIZE = 6;
+const unsigned char DATA_START = 8;
+const unsigned char DATA_MAX_SIZE = 31;
+
+const unsigned char ADV_MIN_SIZE = 8;
+const unsigned char ADV_MAX_SIZE = HEADER_SIZE + ADDRESS_SIZE + DATA_MAX_SIZE;
+
+const unsigned char PDU_ADDRESS_HEADER_MASK = 0x3F;
```
niklasad1: I don't understand this, is it meant to get rid of the 2 RFU bits in the second byte of the header?

IIRC, the actual length in the header includes the address which is 6-37 bytes.

examples/ble_passive_scanning/advertisement.cpp
```diff
@@ -63,10 +86,18 @@ const char* Advertisement::advertisementTypeToStr() const
 
 bool Advertisement::validAdvertisement(const unsigned char* buf, int len)
 {
-  if (buf != nullptr && len >= ADDRESS_SIZE && buf[0] <= 6) {
```
niklasad1: Ok, I see if the `TxAdd` or `RxAdd` would be set, it be regarded as an invalid advertisement.

This function does just check lengths after this change, maybe rename this method or actually check that `buf[0] & 0xf <= 6`?

examples/ble_passive_scanning/advertisement.cpp
```diff
@@ -28,20 +30,41 @@ bool Advertisement::operator!=(const Advertisement& other) const {
...
+  printf("PDU TxAdd: %d\r\n", pduTxAddSet() ? 1 : 0);
+  printf("PDU RxAdd: %d\r\n", pduRxAddSet() ? 1 : 0);
+  printf("PDU Length: %d\r\n", pduLength());
+  printf("Address: %02x %02x %02x %02x %02x %02x\r\n", address_[5],
          address_[4], address_[3], address_[2], address_[1], address_[0]);
-  printf("BLE AD_Type: %s\r\n", advertisementTypeToStr());
-  printf("Packet Size (address + data): %d\r\n", len_);
   printf("Data: ");
-  for (int i = 0; i < len_ - ADDRESS_SIZE; i++) {
+  for (int i = 0; i < pduLength() - ADDRESS_SIZE; i++)
```
niklasad1: style grumble: I actually prefer braces but I'm no "real" C++ developer

examples/ble_passive_scanning/advertisement.cpp
```diff
@@ -1,16 +1,18 @@
...
+Advertisement::Advertisement(const unsigned char* buf, int len)
 {
-  std::memcpy(&address_, &buf[2], ADDRESS_SIZE);
-  if (len_ > DATA_SIZE) {
-    std::memcpy(&data_, &buf[8], len_ - DATA_SIZE);
-  }
+  std::memcpy(&header_, &buf[HEADER_START], HEADER_SIZE);
+  std::memcpy(&address_, &buf[ADDRESS_START], ADDRESS_SIZE);
+  unsigned char data_len = len - HEADER_SIZE - ADDRESS_SIZE;
+  if (data_len > DATA_MAX_SIZE) data_len = DATA_MAX_SIZE;
```
niklasad1: style grumble: I actually prefer braces but I'm no "real" C++ developer

brghena: Strongly agree. Non-brace if statements lead to future bugs.

examples/ble_passive_scanning/advertisement.h
TobiasSchaffner: I used a reference from texas instruments: http://www.ti.com/lit/an/swra475a/swra475a.pdf

PDU-Length is between 6 and 37 bytes including address. This leads to a data part between 0 and 31 bytes. My reference says that the PDULength field in the header has a length of 6 bits. Makes sense to me as the maximum of 37 perfectly fits into 6 bits.

-> I mask the length of the field.

Did I get something wrong here?

examples/ble_passive_scanning/advertisement.cpp
TobiasSchaffner: Aggree. I Renamed validAdvertisement to checkScanResult.

I only want this method to check if it is save to construct an Advertisement object from it.
Afterwards it is save to check all the fields of the object if they carry the information I expected.

examples/ble_passive_scanning/advertisement.h
niklasad1: No, it's correct but I was more referring to the name of constant. `PDU_ADDRESS_HEADER_MASK` is not clear to me I would suggest `PDU_LEN_HEADER_MASK` or something similar.

TobiasSchaffner: Ah now I see your point. You are of course totally right.

PR #75, opened 2020-03-20
set timer_at to use relativce tics

alexandruradovici:
This is the user space patch to try to fix https://github.com/tock/tock/issues/1691.

The timer_at function now uses the new alarm system call and sets the timer to a relative numbers of tics. It eliminates the need to two extra system calls.

Any feedback is welcome, I am not sure that repeating_cb is correctly implemented now.

### DIFF
--- a/libtock/alarm_timer.c
+++ b/libtock/alarm_timer.c
@@ -68,8 +68,8 @@ static void callback( __attribute__ ((unused)) int unused0,
     uint32_t now = alarm_read();
     // has the alarm not expired yet? (distance from `now` has to be larger or
     // equal to distance from current clock value.
-    if (alarm->expiration - alarm->t0 > now - alarm->t0) {
-      alarm_internal_set(alarm->expiration);
+    if (alarm->expiration > now - alarm->t0) {
+      alarm_internal_relative_set(alarm->expiration);
       break;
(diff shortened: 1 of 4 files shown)


alexandruradovici:
Thank you for the feedback, I made the changes. The PR supporting this has been already merged (https://github.com/tock/tock/pull/1709).

libtock/alarm_timer.c
```diff
@@ -23,7 +23,7 @@ static void root_insert(alarm_t* alarm) {
   alarm_t **cur = &root;
   alarm_t *prev = NULL;
   while (*cur != NULL) {
-    if (cmp_exp(alarm->t0, alarm->expiration, (*cur)->expiration) < 0) {
+    if (cmp_exp(alarm->t0, alarm->interval, (*cur)->interval) < 0) {
```
ppannuto: Should this be referenced from `alarm->t0` if comparing intervals now? I think this method of scheduling doesn't quite work anymore, this list ordering assumes a fixed time, but really we need to somehow understand what amount of an interval has expired when sorting the next one into the list.

phil-levis:
Should be fixed by https://github.com/tock/tock/pull/2089 ?

PR #76, opened 2020-03-25
User library for HD44780 LCD capsule

teonaseverin:
User library and example for HD44780 LCD capsule(https://github.com/tock/tock/pull/1715).


libtock/hd44780.h
```diff
@@ -0,0 +1,47 @@
+/*
+	hd44780.h - C implementation of the LiquidCrystal class, providing basic
+	API for communicating with the 16x2 HD44780 LCD; 
+
+	The code was developed based on the basic Arduino Library for the LCD: 
+		https://github.com/arduino-libraries/LiquidCrystal.git
```
ppannuto: Like #72, I believe the GPL derivation means we can't accept this :(

teonaseverin: That wasn't supposed to be here. Initially the library was in user-space, and when I modified it I forgot to remove the comment.

libtock/hd44780.c
```diff
@@ -0,0 +1,269 @@
...
+void hd44780_no_display(void)
+{
+  int ret = command(DRIVER_HD44780_NUM, 7, 1, 0);
+  if (ret)
+    return;
+}
+/* hd44780_display turns the display on very quickly
+ *
+ * Example:
+ *  hd44780_no_display(void);
```
ppannuto, line 63:
```suggestion
 *  hd44780_display(void);
```

libtock/hd44780.c
```diff
@@ -0,0 +1,269 @@
...
+ */
+void hd44780_no_display(void)
+{
+  int ret = command(DRIVER_HD44780_NUM, 7, 1, 0);
+  if (ret)
+    return;
+}
+/* hd44780_display turns the display on very quickly
```
ppannuto, line 60:
```suggestion

/* hd44780_display turns the display on very quickly
```

libtock/hd44780.c
```diff
@@ -0,0 +1,269 @@
...
+ * Example:
+ *  hd44780_no_blink(void);
+ */
+void hd44780_no_blink(void)
+{
+  int ret = command(DRIVER_HD44780_NUM, 8, 1, 0);
+  if (ret)
+    return;
+}
+/* hd44780_blink turns on the blinking cursor display
```
ppannuto, line 83:
```suggestion

/* hd44780_blink turns on the blinking cursor display
```

libtock/hd44780.c
```diff
@@ -0,0 +1,269 @@
...
+ * Example:
+ *  hd44780_scroll_display_left(void);
+ */
+void hd44780_scroll_display_left(void)
+{
+  int ret = command(DRIVER_HD44780_NUM, 9, 0, 0);
+  if (ret)
+    return;
+}
+/* hd44780_scroll_display_right scrolls the display to the right without
```
ppannuto, line 130:
```suggestion

/* hd44780_scroll_display_right scrolls the display to the right without
```

libtock/hd44780.c
```diff
@@ -0,0 +1,269 @@
...
+ * Example:
+ *  hd44780_left_to_right(void);
+ */
+void hd44780_left_to_right(void)
+{
+  int ret = command(DRIVER_HD44780_NUM, 4, 0, 0);
+  if (ret)
+    return;
+}
+/* hd44780_right_to_left should flow the text from right to left on the display
```
ppannuto, line 154:
```suggestion

/* hd44780_right_to_left should flow the text from right to left on the display
```

libtock/hd44780.c
```diff
@@ -0,0 +1,269 @@
...
+ * Example:
+ *	hd44780_autoscroll(void);
+ */
+void hd44780_autoscroll(void)
+{
+  int ret = command(DRIVER_HD44780_NUM, 5, 0, 0);
+  if (ret)
+    return;
+}
+/* hd44780_no_autoscroll will 'left justify' the text from the cursor
```
ppannuto, line 177:
```suggestion

/* hd44780_no_autoscroll will 'left justify' the text from the cursor
```

libtock/hd44780.c
```diff
@@ -0,0 +1,269 @@
...
+    n      += ret;
+    string += ret;
+  }
+  return n;
+}
+
+/* hd44780_set_cursor will set the cursor position to the row and col given as
+ * arguments
+ * As arguments, there are:
+ *	- the column
```
ppannuto, line 247:
```suggestion
 * - the column
```

libtock/hd44780.c
```diff
@@ -0,0 +1,269 @@
...
+ *  - the line
+ *
+ * Example:
+ *	hd44780_set_cursor(5, 1);
+ */
+
```
ppannuto, line 253:
```suggestion
```

libtock/hd44780.c
```diff
@@ -0,0 +1,269 @@
...
+{
+  int ret = command(DRIVER_HD44780_NUM, 1, col, row);
+  if (ret)
+    return;
+}
+
+/* get_min returns the lower number from the 2 arguments given
+ *
+ */
+uint32_t get_min(uint32_t a, uint32_t b)
```
ppannuto, line 264:
```suggestion
static uint32_t get_min(uint32_t a, uint32_t b)
```

We don't have LTO, so if you can tell the compiler this won't need external linkage it can inline it more aggressively

ppannuto: May also need to move this to the top of the file

libtock/hd44780.h
```diff
@@ -0,0 +1,43 @@
...
+void hd44780_left_to_right(void);
+void hd44780_right_to_left(void);
+void hd44780_autoscroll(void);
+void hd44780_no_autoscroll(void);
+void hd44780_set_cursor(uint8_t line, uint8_t position);
+
+uint8_t hd44780_print_string(char* str);
+uint8_t hd44780_print_full_string(char* str);
+
+uint32_t get_min(uint32_t a, uint32_t b);
```
ppannuto, line 39:
```suggestion
```

Don't think it really makes sense for this to export a generic min function (could add one to tock.h if really needed, but think it's better to leave local and let the compiler inline, as noted in the c file)

ppannuto:
The other issue: currently this library just suppresses all possible errors from the underlying driver (the `if (ret) return` pattern). I'd assumed at first that the underlying commands couldn't error, but it looks like they can (i.e. `EBUSY`).

What's the rationale for not bubbling the error back up to the caller? Silently dropping errors seems like a real nightmare to debug.

libtock/hd44780.h
teonaseverin: Hello and sorry for the delay. I updated the little format changes and I delete the get_min function. You are right, it didn't make any sense.

libtock/hd44780.c
```diff
@@ -0,0 +1,282 @@
...
+ *
+ * Example:
+ *  hd44780_clear(void);
+ */
+int hd44780_clear(void)
+{
+  int ret = command(DRIVER_HD44780_NUM, 3, 0, 0);
+  if (ret)
+    return TOCK_EBUSY;
+  return TOCK_SUCCESS;
```
ppannuto, line 37:
```suggestion
  return command(DRIVER_HD44780_NUM, 3, 0, 0);
```

Here and everywhere else, what is the purpose for masking all errors to `EBUSY`? Why not report the actual error?

teonaseverin: Hello, I updated!

PR #78, opened 2020-03-30
User space library for L3gd20 driver

alexandruradovici:
Added the user space for the l3gd20 capsule (https://github.com/tock/tock/pull/1723).

Updated the ninedof user space and added the gyroscope functions.

### DIFF
--- a/examples/l3gd20/main.c
+++ b/examples/l3gd20/main.c
@@ -1,0 +1,37 @@
+/* vim: set sw=2 expandtab tw=80: */
+
+#include <l3gd20.h>
+#include <math.h>
+#include <stdio.h>
+#include <timer.h>
+
+#define ABS(N) ((N < 0) ? (-N) : (N))
+
+static int integer (float f)
+{
+  return trunc (f);
+}
+
+static int decimals (float f, int n)
+{
+  return trunc((ABS(f - trunc (f))) * pow (10, n));
+}
+
+int main(void) {
+  L3GD20XYZ xyz;
+  int temp;
+  printf ("is present %d\r\n", l3gd20_is_present ());
+  printf ("power on %d\r\n", l3gd20_power_on ());
+  printf ("hpf set parameters %d\r\n", l3gd20_set_hpf_parameters (L3GD20_HPF_MODE_NORMAL, L3GD20_HPF_DIV_64));
+  printf ("enable hpf %d\r\n", l3gd20_enable_hpf (true));
+  while (1)
+  {
+    if (l3gd20_read_xyz (&xyz) == TOCK_SUCCESS && l3gd20_read_temperature (&temp) == TOCK_SUCCESS)
+    {
+      printf ("temperature %d x %d.%d y %d.%d z %d.%d\r\n", temp, integer(xyz.x), decimals(xyz.x, 3),
+              integer(xyz.y), decimals(xyz.y, 3), integer(xyz.z), decimals(xyz.z, 3));
+    }
+    delay_ms (10);
+  }
+  return 0;
+}
--- a/examples/l3gd20/main.c
+++ b/examples/l3gd20/main.c
@@ -26,8 +26,7 @@ int main(void) {
   printf ("enable hpf %d\r\n", l3gd20_enable_hpf (true));
   while (1)
   {
-    if (l3gd20_read_xyz (&xyz) == TOCK_SUCCESS && l3gd20_read_temperature (&temp) == TOCK_SUCCESS)
-    {
+    if (l3gd20_read_xyz (&xyz) == TOCK_SUCCESS && l3gd20_read_temperature (&temp) == TOCK_SUCCESS) {
       printf ("temperature %d x %d.%d y %d.%d z %d.%d\r\n", temp, integer(xyz.x), decimals(xyz.x, 3),
               integer(xyz.y), decimals(xyz.y, 3), integer(xyz.z), decimals(xyz.z, 3));
     }
--- a/examples/l3gd20/main.c
+++ b/examples/l3gd20/main.c
@@ -5,32 +5,41 @@
 #include <stdio.h>
 #include <timer.h>
 
-#define ABS(N) ((N < 0) ? (-N) : (N))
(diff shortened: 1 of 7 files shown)


examples/sensors/main.c
```diff
@@ -42,23 +42,23 @@ static void timer_fired(__attribute__ ((unused)) int arg0,
...
+  if (lps25hb)        printf("LPS25HB:    Pressure:        %d\r\n", lps25hb_pressure);
+  if (temp)           printf("Temperature:                 %d deg C\r\n", temp/100);
+  if (humi)           printf("Humidity:                    %u%%\r\n", humi/100);
+  if (ninedof)        printf("FXOS8700CQ: X:               %d\r\n", ninedof_x);
+  if (ninedof)        printf("FXOS8700CQ: Y:               %d\r\n", ninedof_y);
+  if (ninedof)        printf("FXOS8700CQ: Z:               %d\r\n", ninedof_z);
   /* *INDENT-ON* */
 
-  printf("\n");
+  printf("\r\n");
```
ppannuto, line 56: Can you pull the `\r` changes out of this PR? I don't think they have anything to do with this driver.

I understand some serial terminals require CRLF line endings, and maybe all the tock examples could use them, but right now we pretty consistently use just `\n` endings (excepting one BLE example, apparently)

examples/l3gd20/main.c
```diff
@@ -0,0 +1,36 @@
+/* vim: set sw=2 expandtab tw=80: */
+
+#include <l3gd20.h>
+#include <math.h>
+#include <stdio.h>
+#include <timer.h>
+
+#define ABS(N) ((N < 0) ? (-N) : (N))
```
ppannuto, line 8: This already pulls in math.h, why not just use `fabs`?

examples/l3gd20/main.c
```diff
@@ -0,0 +1,36 @@
...
+
+static int integer (float f)
```
ppannuto, line 10: why have a function for this instead of just `(int)` cast?

examples/l3gd20/main.c
```diff
@@ -0,0 +1,36 @@
...
+  printf ("is present %d\r\n", l3gd20_is_present ());
+  printf ("power on %d\r\n", l3gd20_power_on ());
+  printf ("hpf set parameters %d\r\n", l3gd20_set_hpf_parameters (L3GD20_HPF_MODE_NORMAL, L3GD20_HPF_DIV_64));
+  printf ("enable hpf %d\r\n", l3gd20_enable_hpf (true));
+  while (1)
+  {
+    if (l3gd20_read_xyz (&xyz) == TOCK_SUCCESS && l3gd20_read_temperature (&temp) == TOCK_SUCCESS) {
+      printf ("temperature %d x %d.%d y %d.%d z %d.%d\r\n", temp, integer(xyz.x), decimals(xyz.x, 3),
+              integer(xyz.y), decimals(xyz.y, 3), integer(xyz.z), decimals(xyz.z, 3));
+    }
```
ppannuto, line 32: As a test / example app, it's very helpful to report errors here if they occur; please add the other half of this case

libtock/l3gd20.c
```diff
@@ -0,0 +1,140 @@
+#include "l3gd20.h"
+
+typedef struct l3gd20_response {
+  int data1;
+  int data2;
+  int data3;
+  bool done;
+} L3GD20Response;
+
+const float SENZIVITY_FACTOR[3] = {8.75, 17.5, 70.0};
```
ppannuto, line 10:
```suggestion
const float SENSITIVITY FACTOR[3] = {8.75, 17.5, 70.0};
```

Should this be this instead?

libtock/l3gd20.c
```diff
@@ -0,0 +1,140 @@
+#include "l3gd20.h"
+
+typedef struct l3gd20_response {
+  int data1;
+  int data2;
+  int data3;
+  bool done;
+} L3GD20Response;
+
```
ppannuto, line 9: Could you add a comment here about where these constants come from?

libtock/l3gd20.c
```diff
@@ -0,0 +1,140 @@
...
+  int evalue;
+  L3GD20Response response;
+  response.done = false;
+  // subscribe
+  l3gd20_subscribe (command_callback_yield, &response);
+  evalue = command (DRIVER_NUM_L3GD20, 2, 0, 0);
+  if (evalue == TOCK_SUCCESS) {
+    yield_for (&(response.done));
+  }
+  // unsubscribe
```
ppannuto, line 52: I don't think you should need to unsubscribe here or in the other functions -- it doesn't really hurt, but it doesn't do anything meaningful either

examples/sensors/main.c
alexandruradovici: My shell was one of the ones that needs the \r\n. I'll delete it.

examples/l3gd20/main.c
alexandruradovici: Done

examples/l3gd20/main.c
alexandruradovici: Thank you for the feedback.

alexandruradovici:
Thank you for the feedback, I made the changes.

examples/l3gd20/main.c
```diff
@@ -0,0 +1,37 @@
+/* vim: set sw=2 expandtab tw=80: */
+
+#include <l3gd20.h>
+#include <math.h>
+#include <stdio.h>
+#include <timer.h>
+
+static int decimals (float f, int n)
+{
+  return (int)((fabs(f - (int)(f))) * pow (10, n));
```
ppannuto, line 10: Sigh. I know I literally asked for this, but wow is C hard to read sometimes.

alevy:
It's good that we've merged this, but I think ideally we'd use the kernel's `ninedof` and `temperature` interfaces and use the generic userspace libraries, instead of a peripheral specific one.

If there is a reason we'd prefer peripheral-specific in this case, it would be great to articulate.

PR #79, opened 2020-04-02
RISC-V: Remove prebuilt newlibs and improve RISC-V support

alistair23:
This PR aims to do a few things.

First we add the options for building RISC-V without atomics, this is what OpenTitan does.

Second we remove newlib. There isn't a compelling reason for including newlib in the libtock-c source. Most baremetal toolchains include newlib already and if they don't a user can build their own. The current newlib doesn't work without atomics (OpenTitan) and apparently doesn't work at all (due to GOT issues).

This is another step towards supporting RISC-V on libtock-c.


brghena:
To the best of my memory, the only reason we compiled newlib ourselves was to provide PIC support. So using the toolchain's version for RISC-V, which isn't using PIC, seems like a good idea to me.

AppMakefile.mk
```diff
@@ -128,6 +128,7 @@ $$(BUILDDIR)/$(1)/$(1).elf: $$(OBJS_$(1)) $$(LIBS_$(1)) $$(LEGACY_LIBS_$(1)) $$(
 	    -Xlinker --defsym=KERNEL_HEAP_SIZE=$$(KERNEL_HEAP_SIZE)\
 	    -T $$(LAYOUT)\
 	    -nostdlib\
+	    $(LINK_LIBS)\
```
brghena, line 131: This should be `$$(LINK_LIBS_$(1))` so that the make process appends the correct architecture to the end of it

Configuration.mk
```diff
@@ -11,6 +11,7 @@ MAKEFLAGS += -r
 MAKEFLAGS += -R
 
 # Toolchain programs
+ifndef CC
```
brghena, line 14: If you have `CC` defined in your own environment, would it define it as `-gcc`, `-llvm`, etc.? I want to make sure we were using it the same way people's environment variables would.

alistair23: I'm not sure what you mean.

My `CC` is set as: `riscv32-oe-elf-gcc -mabi=ilp32 --sysroot=/usr/local/oecore-x86_64/sysroots/riscv32-oe-elf`

brghena: If you look, we currently define `CC` as `-gcc` (https://github.com/tock/libtock-c/blob/0d925bf30c2a68c6143f45372cb2f010fa5fd3c6/Configuration.mk#L17)

And a later step  combines that `CC` with a platform specifier such as "arm-none-eabi" (https://github.com/tock/libtock-c/blob/0d925bf30c2a68c6143f45372cb2f010fa5fd3c6/AppMakefile.mk#L98 ) In this case the platform specifier is `$(2)`.

So, just replacing `CC` with `riscv32-oe-elf-gcc -mabi=ilp32 --sysroot=/usr/local/oecore-x86_64/sysroots/riscv32-oe-elf` is not how the build system is set up to work right now for libtock-c. (It might still technically work, because it'll replace the platform prefix with nothing depending on the input architecture).

brghena: An argument could be made that we are using `CC` _wrong_ and should fix it. But I think just replacing it with the user's environment variable is not currently the correct choice.

brghena: Also, it's come up in calls, but wow I had forgotten how convoluted the libtock-c build system has gotten. I think I introduced the foreach $$ stuff, and I regret it. It's definitely past time to investigate different build systems for libtock-c to support building for multiple architectures with a simpler setup than Make provides.

alistair23: Ok, I'm going to drop the `CC` change and we can revisit in another PR as I think you are right we will need more `Makefile` work.

alistair23:
I build it with this command:

```shell
TOCK_ARCHS=rv32imc make
```

I don't think we should enable it by default. Lots of people don't have a RV32 toolchain and RV32 support is still really bad so it shouldn't be a default option.

alistair23:
Thinking about it a bit more.

Without relocation support RV32 support is really bad. It's fragile and takes some detailed knowledge of libtock-c and Tock to get working. Until that is fixed RV32 support shouldn't be enabled by default or even well documented/advertised. For people who know what they are doing we should include the best support we can upstream, to avoid forks and fragmentation, but it's not for everyday use.

PR #80, opened 2020-04-12
Lsm303dlhc userspace

alexandruradovici:
This adds the lsm303dlhc user space for https://github.com/tock/tock/pull/1754.

The lsm303dlhc api may be used to fine tune the sensor, getting data may be used thorught the ninedof.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,7 @@
+{
+	"files.associations": {
+		"*.l": "lex",
+		"cstdio": "cpp",
+		"math.h": "c"
+	}
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,7 +1,0 @@
-{
-	"files.associations": {
-		"*.l": "lex",
-		"cstdio": "cpp",
-		"math.h": "c"
-	}
-}
--- a/examples/sensors/main.c
+++ b/examples/sensors/main.c
@@ -42,11 +42,6 @@ static void timer_fired(__attribute__ ((unused)) int arg0,
   if (lps25hb)      lps25hb_pressure = lps25hb_get_pressure_sync();
   if (temperature)  temperature_read_sync(&temp);
   if (humidity)     humidity_read_sync(&humi);
-  if (ninedof)      { 
-    ninedof_read_acceleration_sync(&ninedof_accel_x, &ninedof_accel_y, &ninedof_accel_z);
-    ninedof_read_magnetometer_sync(&ninedof_magneto_x, &ninedof_magneto_y, &ninedof_magneto_z);
-    ninedof_read_gyroscope_sync(&ninedof_gyro_x, &ninedof_gyro_y, &ninedof_gyro_z);
-  }
 
   if (isl29035)       printf("ISL29035:   Light Intensity: %d\n", light);
   if (tmp006)         printf("TMP006:     Temperature:     %d\n", tmp006_temp);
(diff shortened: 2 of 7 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,7 @@
+{
```
ppannuto, line 1: Is this file meant to be included here?

(I don't use vscode, but we do have tool/editor config files elsewhere in Tock, so not inherently opposed to adding, but if we want to pull in some 'sane editor defaults', probably should be its own PR so people with opinions can take note more easily)

examples/sensors/main.c
```diff
@@ -68,7 +81,7 @@ int main(void) {
   humidity    = driver_exists(DRIVER_NUM_HUMIDITY);
   ninedof     = driver_exists(DRIVER_NUM_NINEDOF);
 
-  // Setup periodic timer to sample the sensors.
+  // // Setup periodic timer to sample the sensors.
```
ppannuto, line 84:
```suggestion
  // Setup periodic timer to sample the sensors.
```

examples/sensors/main.c
```diff
@@ -40,17 +42,28 @@ static void timer_fired(__attribute__ ((unused)) int arg0,
...
   if (isl29035)       printf("ISL29035:   Light Intensity: %d\n", light);
   if (tmp006)         printf("TMP006:     Temperature:     %d\n", tmp006_temp);
   if (tsl2561)        printf("TSL2561:    Light:           %d lux\n", tsl2561_lux);
   if (lps25hb)        printf("LPS25HB:    Pressure:        %d\n", lps25hb_pressure);
   if (temp)           printf("Temperature:                 %d deg C\n", temp/100);
   if (humi)           printf("Humidity:                    %u%%\n", humi/100);
-  if (ninedof)        printf("FXOS8700CQ: X:               %d\n", ninedof_x);
-  if (ninedof)        printf("FXOS8700CQ: Y:               %d\n", ninedof_y);
-  if (ninedof)        printf("FXOS8700CQ: Z:               %d\n", ninedof_z);
+  if (ninedof)   {
```
ppannuto, line 52: Hmm.., I see why you restructured this, but I think it would be better to handle all the sensors the same way. I think this app was written with hail/imix in mind, and kind of assumes all the sensors will work, but not that we have more boards with sensors, they really all need the `if .. == TOCK_SUCCESS` treatment.

---

The other issue right now, I think the spacing / alignment on the new ones don't line up (e.g. `Gryo: ` definitely needs more spaces).

examples/sensors/main.c
alexandruradovici: Done

examples/sensors/main.c
alexandruradovici: I fixed the gyro spaces, I think it lines up now. If we remove the ifs, if there is no sensor present, it will just display 0s. Is that ok?

.vscode/settings.json
alexandruradovici: No, my mistake. I deleted it and added .vscode to gitignore, may others use vscode.

examples/sensors/main.c
ppannuto: I think my question is actually the opposite: should we add more if's, i.e. is the light sensor line just printing 0's on the STM board now?

alexandruradovici: It is not printing anything, as there is no light sensor capsule (or actual sensor). It does not pass the "capsule present" if.

ppannuto: OH. Right, I remember how this app works. Sorry.

However, with that in mind, should we expand to `ninedof_accel` `ninedof_mag` and `ninedof_gyro` booleans to match the other sensors?

PR #83, opened 2020-04-21
hail test: add more features

bradjc:
This makes the hail test app use 12/13 grant regions (all but IPC).


ppannuto:
Does this work for you? It hangs for me, while current master works fine.

ppannuto:
I just pushed a second commit that I think makes sense to add to this eventually, but the hang happens even without that commit.

bradjc:
Works just fine, and I tried on two different boards.

PR #84, opened 2020-04-22
AppMakefile.mk: Fix the linker ordering

alistair23:


alistair23:
When I removed the RV32 newlib I started linking with the toolchains newlib. After building more test cases for the 1.5 release I noticed that linking would fail based on the order, this fixes that up.

PR #87, opened 2020-05-27
Various improvements to Make in anticipation of RISC-V support

bradjc:
This pulls the changes from the riscv branch that are useful updates for all platforms. There are some lingering cortex-m-only things in Make that this puts into cortex-m variables.

It's also very convenient to be able to generate a .lst file for apps compiled for a fixed address without having to set the make variables to the correct addresses.


AppMakefile.mk
```diff
@@ -251,12 +251,15 @@ $(BUILDDIR)/$(PACKAGE_NAME).tab: $(foreach platform, $(TOCK_ARCHS), $(BUILDDIR)/
...
-# The size target accumlates dependencies in the platform build rule creation
+# The size target accumulates dependencies in the platform build rule creation
 .PHONY: size
 
 .PHONY: debug
 debug:	$(foreach platform, $(TOCK_ARCHS), $(BUILDDIR)/$(call ARCH_FN,$(platform))/$(call ARCH_FN,$(platform)).userland_debug.lst)
 
+.PHONY: lst
+lst:	$(foreach platform, $(TOCK_ARCHS), $(BUILDDIR)/$(call ARCH_FN,$(platform))/$(call ARCH_FN,$(platform)).lst)
+
```
brghena, line 262: Can you add some comments above this rule and the `make debug` rule to explain which  one you use when?

AppMakefile.mk
```diff
@@ -208,7 +208,7 @@ endif
 # Step 2: Create a new ELF with the layout that matches what's loaded
 $$(BUILDDIR)/$(1)/$(1).userland_debug.elf: $$(OBJS_$(1)) $$(LIBS_$(1)) $$(LEGACY_LIBS_$(1)) $$(BUILDDIR)/$(1)/$(1).userland_debug.ld | $$(BUILDDIR)/$(1)
 	$$(TRACE_LD)
-	$$(Q)$(2)$$(CC) $$(CFLAGS) -mcpu=$(1) $$(CPPFLAGS)\
+	$$(Q)$(2)$$(CC) $$(CFLAGS) $$(CPPFLAGS) $$(CPPFLAGS_$(1))\
```
brghena, line 211: Was the `mcpu` specification not necessary?

bradjc: It gets thrown back in there through `$$(CPPFLAGS_$(1))`, I think I just missed removing it here by default it in the first round of these changes.

bradjc:
This PR was a good thought, but I forgot about it and it's easier to just use #88.

Issue #44, opened 2019-06-11
How to support RISC-V apps?


bradjc:
Do we want to expect libtock-c users to have the riscv compiler installed and have apps be built for RISC-V as well as Thumb?

ppannuto:
I mean, you could ask the equivalent question of should we require RISC-V users to have the ARM compiler installed and building unused Thumb binaries. There's a toolchain / fragmentation tradeoff here.

My lean is to say that adding one more package to the "install this" list of copy-pasted `apt`/`brew`/whatever isn't so painful -- this strongly assumes that installing the toolchain is effectively a no-op (which I think is true). We could certainly configure the build system to do runtime detection (e.g. guarding all of the arch-specific build rules on the result of a `if $$(which $(CC))`), but I don't think that's worth the maintenance

bradjc:
Update: no real progress on this issue. If you are here because you are looking for RISC-V support, look for the `riscv` [branch](https://github.com/tock/libtock-c/tree/riscv) which should be able to build apps for RISC-V platforms.

Actual progress is probably blocked on being able to support some sort of position-independent applications. Right now compiling apps for RISC-V is very cludgy since they have to be compiled for the address they are eventually placed at.

hudson-ayers:
In general I agree with Pat on this, but if the install process for Linux is much more complicated / distro-dependent than an apt install, I would push back. I don't think that asking all linux users to build the toolchain from source and install it is particularly reasonable, for example (which seems like what most Linux users have been doing up until now)

ppannuto:
I think this is closed by #88.

PR #88, opened 2020-06-05
Compile apps for RISC-V with make flag

bradjc:
This adds `rv32imac` binaries to TABs from libtock-c. These are compiled for specific flash and RAM addresses.

The main change for doing this is automatically generating the linker script when building for each architecture. This copies the default linker script and sets the addresses if needed. Different compilation versions are set in the Configuration.mk makefile.

This also compiles libraries for rv32imac, and updates some apps that use assembly.

TODO
------

- [x] fix crt0 header for riscv, need to take the stack size into account
- [x] make got just a copy
- [x] fix lst build for cortex-m


examples/tests/stack_size_test02/main.c
```diff
@@ -2,7 +2,12 @@ int main(void) {
   printf("Stack Test App\n");
 
   register uint32_t stack_pointer;
+
+#if defined(__thumb__)
   asm volatile ("MOV %0, sp\n" : "=r" (stack_pointer));
+#elif defined(__riscv)
+  asm volatile ("mv %0, sp\n" : "=r" (stack_pointer));
+#endif
```
ppannuto, line 10:
```suggestion
#elif defined(__riscv)
  asm volatile ("mv %0, sp\n" : "=r" (stack_pointer));
#else
#error Unknown architecture
#endif
```

Here and elsewhere, I think once we're doing multi-arch stuff like this we should add the last case to catch anything that might get missed in the future.

examples/lua-hello/Makefile
```diff
@@ -3,6 +3,10 @@
 # Specify this directory relative to the current application.
 TOCK_USERLAND_BASE_DIR = ../..
 
+# We only compile this for Cortex-M platforms because of compiler
+# errors when building lua on risc-v.
+TOCK_TARGETS = cortex-m0 cortex-m3 cortex-m4
```
ppannuto, line 8:
```suggestion
TOCK_TARGETS := cortex-m0 cortex-m3 cortex-m4
```

Generally avoiding lazy evaluation makes all of the rest of the machinery more reliable.

examples/ble_passive_scanning/Makefile
```diff
@@ -3,6 +3,9 @@
 # Specify this directory relative to the current application.
 TOCK_USERLAND_BASE_DIR = ../..
 
+# This app only compiles (due to C++) for Cortex-M platforms.
+TOCK_TARGETS = cortex-m0 cortex-m3 cortex-m4
```
ppannuto, line 7:
```suggestion
TOCK_TARGETS := cortex-m0 cortex-m3 cortex-m4
```

AppMakefile.mk
```diff
@@ -95,65 +88,98 @@ $$(BUILDDIR)/$(1):
...
+
+# Force this to be built every time so we always generate the needed ld file
+# with the correct addresses.
+.PHONY: _FORCE_USERLAND_LD_CUSTOM
+
+$$(BUILDDIR)/$(1)/$(2).custom.ld: $$(LAYOUT) _FORCE_USERLAND_LD_CUSTOM | $$(BUILDDIR)/$(1)
+	@# Start with a copy of the template / generic ld script
+	$$(Q)cp $$< $$@
+	@# And with apologies to future readers, this is easier as one shell
+	@# command/script so we can set intervening variables. We want to replace
```
ppannuto, line 138: What intervening variables are set? Was this from an older version?

AppMakefile.mk
```diff
@@ -95,65 +88,98 @@ $$(BUILDDIR)/$(1):
...
+	@# the flash and RAM addresses with specific versions in the generic file.
+	@#
+	@# #616 #635: sed is not cross-platform
+	@# https://stackoverflow.com/a/22247781/358675 <-- Use perl in place of sed
+	$$(Q)set -e ;\
+	  perl -pi -e "s/(FLASH.*ORIGIN[ =]*)([x0-9]*)(,.*LENGTH)/\$$$${1}$(3)\$$$$3/" $$@ ;\
+	  perl -pi -e "s/(SRAM.*ORIGIN[ =]*)([x0-9]*)(,.*LENGTH)/\$$$${1}$(4)\$$$$3/" $$@
```
ppannuto, line 145:
```suggestion
	$$(Q)\
	  perl -pi -e "s/(FLASH.*ORIGIN[ =]*)([x0-9]*)(,.*LENGTH)/\$$$${1}$(3)\$$$$3/" $$@ &&\
	  perl -pi -e "s/(SRAM.*ORIGIN[ =]*)([x0-9]*)(,.*LENGTH)/\$$$${1}$(4)\$$$$3/" $$@
```

is slightly simpler chaining. Thought it's not clear to me why these need to be chained at all?

---

Aside: What beautiful dollar-sign density: `\$$$${1}$(3)\$$$$3/` -- now you're really getting the full joy of Make

README.md
```diff
@@ -33,14 +33,15 @@ Prerequisites
...
-   `arm-none-eabi` toolchain.
+1. The main requirement to build the C applications in this repository is having
+   cross compilers for embedded targets. You will need an `arm-none-eabi`
+   toolchain for Cortex-M targets, and `riscv64-unknown-elf` toolchain for
+   RISC-V targets.
 
    MacOS:
    ```
    $ brew tap ARMmbed/homebrew-formulae && brew update && brew install arm-none-eabi-gcc
+   $ brew install riscv-gnu-toolchain --with-multilib
```
ppannuto, line 44:
```shell
$ brew install riscv-gnu-toolchain --with-multilib
[...]
Error: invalid option: --with-multilib

$ brew --version
Homebrew 2.3.0
Homebrew/homebrew-core (git revision 91f59; last commit 2020-06-05)
Homebrew/homebrew-cask (git revision 20cbf; last commit 2020-06-05)
```

?

README.md
```diff
@@ -33,14 +33,15 @@ Prerequisites
...
    ```
 
    Ubuntu (18.04LTS or later):
```
hudson-ayers, line 47: I can't build any apps on this branch because there aren't instructions for installing the riscv toolchain for Linux. 

I tried `sudo apt install gcc-riscv64-unknown-elf`, which got me past my initial error about the command not being found, but led to:

```
(.venv) hudson: ~/libtock-c/examples/blink (riscv-for-all) $ make
  CC        ../../libtock/internal/alarm_internal.c
In file included from <command-line>:31:
./../../support/warning_header.h:6:10: fatal error: stdio.h: No such file or directory
    6 | #include <stdio.h>
      |          ^~~~~~~~~
compilation terminated.
make: *** [../../TockLibrary.mk:149: ../../libtock/build/rv32imac/alarm_internal.o] Error 1
```

Do we have to use the siFive toolchain?

bradjc: I'm not sure, I need help here.

AppMakefile.mk
bradjc: I didn't want to delete that in case it became correct.

AppMakefile.mk
bradjc: I mean...you wrote that originally. I didn't want to change things and then break things.

README.md
bradjc: Arg that should be

`brew tap riscv/riscv` then install.

AppMakefile.mk
ppannuto: Hahaha, well, that sounds right. I'm older and wiser now. I think those can just be two individual commands.

bradjc:
Ok it seems like the main issue is how to install toolchains, and make sure they work on both mac and linux. There was no pushback on https://github.com/tock/libtock-c/issues/44, so I'm assuming there is no reservations about supporting both, we just need reasonable install steps.

If anyone knows of a trick for installing on Linux, please let me know.

bradjc:
I don't even know where to post an issue (or comment on one) asking the toolchain maintainers to make this easier. But, I'm kind of thinking we just host our own build for ubuntu.

.travis-install-gcc
```diff
@@ -10,3 +10,7 @@ if [ ! -x gcc-arm-none-eabi-9-2019-q4-major/bin/arm-none-eabi-gcc ]; then
   tar -xjvf gcc.tar.bz2
 fi
 
+if [ ! -x riscv64-unknown-elf-gcc-8.3.0-2019.08.0-x86_64-linux-ubuntu14/bin/riscv64-unknown-elf-gcc ]; then
+  wget https://static.dev.sifive.com/dev-tools/riscv64-unknown-elf-gcc-8.3.0-2019.08.0-x86_64-linux-ubuntu14.tar.gz -O gcc-riscv.tar.bz2
```
alistair23, line 14: Don't do this, this is not a supported toolchain

.travis.yml
```diff
@@ -28,6 +28,7 @@ cache:
   cargo: true
   directories:
     - $HOME/gcc-arm-none-eabi-9-2019-q4-major
+    - $HOME/riscv64-unknown-elf-gcc-8.3.0-2019.08.0-x86_64-linux-ubuntu14
```
alistair23, line 31: Debian10 has a toolchain, does that not work here?

README.md
```diff
@@ -53,6 +54,8 @@ Prerequisites
    $ sudo add-apt-repository ppa:team-gcc-arm-embedded/ppa && sudo apt update && sudo apt install gcc-arm-embedded
    ```
 
+   The RISC-V binaries are pre-compiled from [SiFive](https://www.sifive.com/boards).
```
alistair23, line 57: We shouldn't do this

README.md
```diff
@@ -33,14 +33,15 @@ Prerequisites
     $ cd libtock-c
     ```
 
-1. The main requirement to build the C applications in this repository
-   is a cross compiler for embedded targets. Currently, the only
-   supportetd target is ARM Cortex M, so you will need an
-   `arm-none-eabi` toolchain.
+1. The main requirement to build the C applications in this repository is having
+   cross compilers for embedded targets. You will need an `arm-none-eabi`
+   toolchain for Cortex-M targets, and `riscv64-unknown-elf` toolchain for
```
alistair23, line 38: Some distros package them separately, it could be a `riscv32-unknown-elf` toolchain.

Configuration.mk
```diff
@@ -28,8 +28,33 @@ KERNEL_HEAP_SIZE ?= 1024
 # PACKAGE_NAME is used to identify the application for IPC and for error reporting
 PACKAGE_NAME ?= $(shell basename "$(shell pwd)")
 
-# Tock supported architectures
-TOCK_ARCHS ?= cortex-m0|arm-none-eabi cortex-m3|arm-none-eabi cortex-m4|arm-none-eabi
+# Tock supported architectures.
+#
+# Each app will be compiled for each of the architectures.
+TOCK_ARCHS ?= cortex-m0 cortex-m3 cortex-m4 rv32imac
```
alistair23, line 34: Can we support OT as well?

alistair23: OT is `rv32imc`

Configuration.mk
```diff
@@ -28,8 +28,33 @@ KERNEL_HEAP_SIZE ?= 1024
...
+# 3. (Optional) The address to use as the fixed start of flash.
+# 4. (Optional) The address to use as the fixed start of RAM.
+#
+#  rv32imac|rv32imac.0x20040040.0x80002400|0x20040040|0x80002400 # RISC-V for HiFive1b
+#  rv32imac|rv32imac.0x40430060.0x80004000|0x40430060|0x80004000 # RISC-V for arty-e21
+TOCK_TARGETS ?= cortex-m0\
+                cortex-m3\
+                cortex-m4\
+                rv32imac|rv32imac.0x20040040.0x80002400|0x20040040|0x80002400\
+                rv32imac|rv32imac.0x40430060.0x80004000|0x40430060|0x80004000
```
alistair23, line 57: Can OT be added as well?

Configuration.mk
```diff
@@ -40,6 +65,12 @@ endif
 ELF2TAB_ARGS += -n $(PACKAGE_NAME)
 ELF2TAB_ARGS += --stack $(STACK_SIZE) --app-heap $(APP_HEAP_SIZE) --kernel-heap $(KERNEL_HEAP_SIZE)
 
+# Setup the correct toolchain for each architecture.
+override TOOLCHAIN_cortex-m0 = arm-none-eabi
+override TOOLCHAIN_cortex-m3 = arm-none-eabi
+override TOOLCHAIN_cortex-m4 = arm-none-eabi
+override TOOLCHAIN_rv32imac = riscv64-unknown-elf
```
alistair23, line 72: It looks like the Debian toolchain insteads 32-bit support in the 64-bit toolchain, but not all setups would do that. It might be worth documenting that this need to be edited if that's the case.

bradjc: We can't have users edit makefiles, we either need to have instructions on how to install the "correct" binaries, or have the makefiles support both.

Configuration.mk
bradjc: Yes of course.

bradjc:
We pre-compile binaries (nrf serialization for example) to make it easier for users.

Configuration.mk
alistair23: Ok, it probably is easiest to default to `riscv32-unknown-elf` and fallback to `riscv64-unknown-elf`. That should cover all bases.

alistair23:
How do you handle different RISC-V ISA strings in the pre-compiled binaries? For example supporting both OT and the HiFive1?

bradjc:
It's no different than cortex-m3 vs m4 (at least it shouldn't be).

alistair23:
Good point.

Just watch out as there are a lot of different combinations that will have to be built and maintained as RISC-V support grows.

README.md
```diff
@@ -33,30 +33,44 @@ Prerequisites
...
    ```
    $ sudo apt install gcc-arm-none-eabi
    ```
 
-   Ubuntu (before 18.04):
+   You will also need a `riscv64` toolchain. You can download a pre-built
+   toolchain
+   [here](http://cs.virginia.edu/~bjc8c/archive/gcc-riscv64-unknown-elf-8.3.0-linux.zip),
+   or follow [these
+   instructions](https://github.com/riscv/riscv-gnu-toolchain#installation-newlib)
```
alistair23, line 59: People shouldn't follow these steps. This is an out of date fork that we shouldn't be using.

README.md
```diff
@@ -33,30 +33,44 @@ Prerequisites
...
+
+   **Ubuntu (18.04LTS or later)**:
    ```
    $ sudo apt install gcc-arm-none-eabi
    ```
 
-   Ubuntu (before 18.04):
+   You will also need a `riscv64` toolchain. You can download a pre-built
+   toolchain
+   [here](http://cs.virginia.edu/~bjc8c/archive/gcc-riscv64-unknown-elf-8.3.0-linux.zip),
```
alistair23, line 57: Can we also document the install guide for people who don't want to download this?

I'm unable to download it as well

README.md
bradjc: Maybe on a call this week, can you explain this ecosystem? You seem to be in tune with it, but to me it's like black magic. How do you know it is out of date? There is a commit from 20 days ago. Where is the real version? How is anyone supposed to track this?

alistair23: Before RISC-V support was in upstream GCC, binutils, libc's ect. then then RISC-V toolchain fork was the best option.

Now that RISC-V support is supported in all the upstream projects that is what should be used. The upstream projects are heavily tested and shipped in distros. They are reviewed by more people and follow the standard release processes.

The RISC-V fork is pretty much exclusively maintained by SiFive for their them. For example the RISC-V binutils fork description on GitHub says: "RISC-V backports for binutils-gdb. Development is done upstream at the FSF". The fork is not as heavily tested and doesn't have formal release process.

Everyone should be using the upstream projects for toolchains at this point.

Configuration.mk
bradjc: What addresses should I add?

README.md
```diff
@@ -33,30 +33,60 @@ Prerequisites
...
    ```
    $ brew tap ARMmbed/homebrew-formulae && brew update && brew install arm-none-eabi-gcc
+   $ brew tap riscv/riscv && brew update && brew install riscv-gnu-toolchain --with-multilib
    ```
 
-   Ubuntu (18.04LTS or later):
+   You can also download a pre-built RISC-V toolchain
+   [here](http://cs.virginia.edu/~bjc8c/archive/gcc-riscv64-unknown-elf-8.3.0-mac.zip).
+
+   **Ubuntu (18.04LTS or later)**:
```
alistair23, line 50: This will only work on 20.04 LTS. Maybe some earlier non-LTS versions work as well, but 18.04 doesn't

README.md
```diff
@@ -33,30 +33,60 @@ Prerequisites
...
    ```
    $ sudo apt install gcc-arm-none-eabi
+   $ sudo apt install gcc-riscv64-unknown-elf
+   ```
+
+   Unfortunately, Ubuntu does not currently (June 2020) provide a package for
```
alistair23, line 56: It's probably best to specify the version as earlier then 20.10

Configuration.mk
alistair23: FLASH (rx) : ORIGIN = 0x20030040, LENGTH = 32M
  SRAM (rwx) : ORIGIN = 0x10002D00, LENGTH = 512K

README.md
bradjc: What happens if this isn't fixed in 20.10?

alistair23: Then we will have to update it, but a date has the same problem. What if 20.10 works? It's then not clear to people that it should work for them.

Either way though we will need to update the steps.

userland_generic.ld
```diff
@@ -89,6 +95,21 @@ SECTIONS {
         . = ALIGN(4); /* Make sure we're word-aligned here */
     } > FLASH =0xFF
 
+    /* Need to reserve room for the stack in the linker file. This makes the
+     * _got addresses used by the compiler match what they will be when the
+     * app is loaded into memory. This is not necessary for full PIC supported
+     * platforms (like Cortex-M), but is needed when an app is compiled for a
+     * fixed address.
```
ppannuto, line 102: I don't quite understand this comment. Why are there GOT address on non-PIC platforms?

README.md
```diff
@@ -33,30 +33,74 @@ Prerequisites
...
+1. The main requirement to build the C applications in this repository is having
+   cross compilers for embedded targets. You will need an `arm-none-eabi`
+   toolchain for Cortex-M targets, and `riscv64-unknown-elf` toolchain for
+   RISC-V targets.
 
-   MacOS:
+   **MacOS**:
    ```
    $ brew tap ARMmbed/homebrew-formulae && brew update && brew install arm-none-eabi-gcc
+   $ brew tap riscv/riscv && brew update && brew install riscv-gnu-toolchain --with-multilib
```
ppannuto, line 44: How important is `--with-multilib` / what is it doing? That causes brew to install from source, which fails on my machine:

(relevant bits of the logs):

```
checking for ANSI C header files... yes
checking for the correct version of mpc.h... rm: conftest.dSYM: is a directory
no
configure: error: Building GCC requires GMP 4.2+, MPFR 3.1.0+ and MPC 0.8.0+.
Try the --with-gmp, --with-mpfr and/or --with-mpc options to specify
their locations.  Source code for these libraries can be found at
their respective hosting sites as well as at
https://gcc.gnu.org/pub/gcc/infrastructure/.  See also
http://gcc.gnu.org/install/prerequisites.html for additional info.  If
you obtained GMP, MPFR and/or MPC from a vendor distribution package,
make sure that you have installed both the libraries and the header
files.  They may be located in separate packages.
rm: conftest.dSYM: is a directory
make: *** [stamps/build-gcc-newlib-stage1] Error 1
make: *** Waiting for unfinished jobs....
```

It's *probably* a missing dependency in the brew formula?

ppannuto:
I don't this `TOCK_TARGETS` is working as expected? Since I don't currently have a working riscv toolchain, I tried building the hail app to see if other stuff still worked, but that resulted in:

```
[-bash] Sun 14 Jun 16:10 [[riscv-for-all] ~/code/helena-project/libtock-c/examples/tests/hail]
$ make
  CC        ../../../libtock/internal/alarm_internal.c
/bin/sh: riscv64-unknown-elf-gcc: command not found
make: *** [../../../libtock/build/rv32imac/alarm_internal.o] Error 127
```

If riscv isn't in the targets list, then its toolchain shouldn't be being invoked at all.

README.md
```diff
@@ -33,30 +33,74 @@ Prerequisites
...
    ```
 
-   Ubuntu (18.04LTS or later):
+   You can also download a pre-built RISC-V toolchain
+   [here](http://cs.virginia.edu/~bjc8c/archive/gcc-riscv64-unknown-elf-8.3.0-mac.zip).
```
ppannuto, line 48: This unfortunately doesn't work unless you have all of the dependent libraries also installed. Running this on my machine resulted in:

```
$ make
  CC        ../../libtock/crt0.c
  AR        ../../libtock/build/cortex-m0/libtock.a
  CC        ../../libtock/crt0.c
  AR        ../../libtock/build/cortex-m3/libtock.a
  CC        ../../libtock/crt0.c
  AR        ../../libtock/build/cortex-m4/libtock.a
  CC        ../../libtock/internal/alarm_internal.c
dyld: Library not loaded: /usr/local/opt/isl/lib/libisl.21.dylib
  Referenced from: /private/tmp/g/gcc-riscv64-unknown-elf-8.3.0/bin/../libexec/gcc/riscv64-unknown-elf/8.3.0/cc1
  Reason: image not found
riscv64-unknown-elf-gcc: internal compiler error: Abort trap: 6 signal terminated program cc1
Please submit a full bug report,
with preprocessed source if appropriate.
See <https://gcc.gnu.org/bugs/> for instructions.
make: *** [../../libtock/build/rv32imac/alarm_internal.o] Error 4
```

ppannuto:
```
$ ls /usr/local/opt/isl/lib/
libisl.22.dylib libisl.a        libisl.dylib    pkgconfig
```

Or, apparently I've just updated something else more recently than you?

In either case, to ship a pre-built compiler image, we'd need to figure out how to statically link everything in the compiler build (if we can?) to be reliable across machines.

README.md
bradjc: Very important, it gets you a 32 bit compiler (in addition to the 64 bit standard version).

userland_generic.ld
bradjc: I don't know? But there are.

libtock/crt0.c
```diff
@@ -208,57 +218,102 @@ void _start(void* app_start __attribute__((unused)),
...
 
-  // fix up GOT
-  volatile uint32_t* got_start     = (uint32_t*)(myhdr->got_start + stacktop);
-  volatile uint32_t* got_sym_start = (uint32_t*)(myhdr->got_sym_start + (uint32_t)app_start);
+  // Fix up the Global Offset Table (GOT).
+
+  // Get the address in memory of where the table should go.
+  volatile uint32_t* got_start     = (uint32_t*)(myhdr->got_start + mem_start);
+  // Get the address in flash of where the table currently is.
+  volatile uint32_t* got_sym_start = (uint32_t*)(myhdr->got_sym_start + app_start);
```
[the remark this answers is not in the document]
bradjc: I don't know.

README.md
bradjc: Fair enough, probably better to just remove this link then.

bradjc:
> I don't this `TOCK_TARGETS` is working as expected? Since I don't currently have a working riscv toolchain, I tried building the hail app to see if other stuff still worked, but that resulted in:

TARGETS works together with the `TOCK_ARCHS`, you have to remove it there as well.

ppannuto:
> TARGETS works together with the TOCK_ARCHS, you have to remove it there as well.

I guess I don't understand why these should be different things?

ppannuto:
Filed https://github.com/riscv/riscv-gcc/issues/190 for ICE

userland_generic.ld
brghena: Can you provide a dump of the symbols that are in the GOT section for a RISC-V application? I'm worried that the existence of the GOT section is a sign that we're doing something wrong.

README.md
ppannuto: As a future reference point, if people are having issues, the inexplicable `brew reinstall gmp mpfr libmpc` fixed this for me (really just `libmpc` in this case, but let's be thorough).

Yes, it was already installed. Yes, `brew info libmpc` confirmed that. No, `brew doctor` reported no errors. Yes, brew was confused that I wanted to reinstall. Yes, that somehow fixed the problem. Yes, I'm mildly+ annoyed.

userland_generic.ld
bradjc: OK I moved the `-fPIC` flag to its own category, and don't include it with the RISC-V builds. The GOT size is now 0.

bradjc:
> > TARGETS works together with the TOCK_ARCHS, you have to remove it there as well.
> 
> I guess I don't understand why these should be different things?

Oh yeah, I remember now. They do different tasks. `TOCK_ARCHS` specifies all of the architectures to compile the app for. `TOCK_TARGETS` specifies all of the specific targets to link the apps for.

ppannuto:
Got it. I pushed a commit so that `TOCK_ARCHS` populates automatically from `TOCK_TARGETS`, that way if there's no riscv target (e.g. lua), only the arm toolchains are used.

bradjc:
So I think the only issue left is installing a toolchain?

- Ubuntu: It would be great if someone could test the ubuntu steps.
- Mac: The steps work generally, but since the brew formula isn't pegged to a version asynchronous updates to the riscv/riscv-gnu-toolchain repo can lead to a buggy compiler. It also takes forever to compile. I asked about maybe fixing that second one: https://github.com/riscv/homebrew-riscv/issues/22. One option is we could create a `homebrew-tock` repo and provide our own bottles with multilib support that are pegged to a version (or commit). We could also then provide formulas for all of our tools, and the entire getting started install becomes `brew tap tock/tock && brew install tock-tools`.

bradjc:
I think I'm concluding that the thing to do at the moment is retain the `riscv` branch, but have it _only_ be a single commit to modify `TOCK_TARGETS`. I would also update the readme to clarify that RISC-V compilation is optional, and that if you want to compile for riscv you need to install the toolchain and checkout the riscv branch.

I think it would be better to get these changes merged, while giving us more time to work on the toolchain problems.

ppannuto:
I agree that these changes are good and shouldn't sit needlessly out-of-tree while we wait on toolchain stuff.

We could even be a bit smarter and just have TOCK_TARGETS conditionally include riscv targets based on the presence of the toolchain (roughly

```make
if $(shell command -v riscv64-unknown-elf-gcc)
TOCK_TARGETS ?= ...all
else
TOCK_TARGETS ?= ...arm-only
endif
```

should work).

hudson-ayers:
> I agree that these changes are good and shouldn't sit needlessly out-of-tree while we wait on toolchain stuff.
> 
> We could even be a bit smarter and just have TOCK_TARGETS conditionally include riscv targets based on the presence of the toolchain (roughly
> 
> ```makefile
> ifeq ($(shell command -v riscv64-unknown-elf-gcc),)
> TOCK_TARGETS ?= ...arm-only
> else
> TOCK_TARGETS ?= ...all
> endif
> ```
> 
> should work).

I agree that making risc-v compilation optional in the master branch is better than having it in a second branch that will have to be kept up to date with master somehow. But having the toolchain installed is not sufficient for it to actually work, and I don't think we should force users to uninstall their risc-v toolchain just to be able to build arm apps, all because their toolchain does not include newlib.

bradjc:
I think it's important the out-of-master change is _very_ small (and never grows). After this PR it would just be changing one make variable (which people will need to do anyway to support new risc-v boards with different flash and ram addresses).

But I think that is preferable to having a dynamic system where supporting users is first establishing what toolchains they have just to understand what our build system should be doing.

hudson-ayers:
Why not just require users to pass `--riscv` to any call to `make` to explicitly build apps for riscv. That seems a lot easier than checking out another branch. Or set an environment variable, so that it persists once you do it once.

ppannuto:
Oy. Didn't realize there was a newlib wildcard. Yeah, the single TARGETS patch feels good.

Or a guard on an environment variable, like `if defined TOCK_LIBTOCKC_ENABLE_RISCV` so that people doing development work don't have to keep this lingering one-line change around while they're working.

alistair23:
A single line Make change works for me, same with a command line argument of environment variable.

It kind of looks like Ubuntu is the only recent distro that doesn't support RV32 embedded libc, so this is really useful and simple for lots of people. Hopefully Ubuntu 20.10 fixes this and then we at least has a working version (although non-LTS versions don't make it to Travis).

bradjc:
I added `make RISCV=1`, which I think is better than the branch idea. It would be nice to include the build on travis, however, but I'm not sure how to best do that.

hudson-ayers:
> I added `make RISCV=1`, which I think is better than the branch idea. It would be nice to include the build on travis, however, but I'm not sure how to best do that.

For CI, the options I see:

1. transition travis to macos so we get the easy toolchain install via brew
2. build newlib from source for ubuntu (this takes like 30 minutes, so don't love it as an option, but maybe we can figure out how to cache it?)
3. add a github actions runner just to run macos tests that build a couple apps for risc-v

bradjc:
The riscv toolchain isn't the problem, the problem is the build all script doesn't include `RISCV=1`.

ppannuto:
Make picks up variables from the environment, i.e. `$ RISCV=1 ./build_all.sh` does the right thing.

---

For CI, probably a separate thing, but should we kick this repo over to github actions as well? That makes mac way less painful as a target.

PR #89, opened 2020-06-11
crt0 and .ld: Cleanup & add stack as section

bradjc:
The motivation for this change is to make sure that applications are
compiled with the GOT, data, and BSS sections at the relative offset in
memory that they will be when an app actually executes. Currently, these
sections are at an offset from the start of memory, when they are
actually used after the stack (and the stack is at the start of memory).

This works fine for PIC apps, since the GOT, data, and BSS sections can
be relocated both to where the app is in flash/memory, and to move them
past the stack. But, for apps compiled for fixed addresses, the GOT,
data, and BSS addresses cannot be arbitrarily moved, they must end up in
memory at the addresses the compiler expects them to be.

The key fix here is to add a `.stack` section to the linker file, so
that the GOT, data, and BSS sections are at the actual addresses they
will be at when the app actually executes. This has no real effect on
PIC apps, but allows fixed address apps to work. However, it does mean
that c_start needs to change to match the new offsets.

The main change to c_start is that now everything is relative to the
start of app flash and start of app memory, rather than the start of app
flash and the end of the ap stack. This mirrors the new linker script,
where everything is now at the correct address relative to the start of
app memory.

I also added comments everywhere since I was changing the lines of code
anyway, and it's not immediately obvious to me why everything works the
way it does.

This works with the hail app on hail.

### DIFF
--- a/libtock/crt0.c
+++ b/libtock/crt0.c
@@ -17,37 +17,47 @@ extern int main(void);
 #pragma GCC diagnostic ignored "-Wmissing-declarations"
 #pragma GCC diagnostic ignored "-Wmissing-prototypes"
 
-/*
- * The structure populated by the linker script at the very beginning of the
- * text segment. It represents sizes and offsets from the text segment of
- * sections that need some sort of loading and/or relocation.
- */
+// The structure populated by the linker script at the very beginning of the
+// text segment. It represents sizes and offsets from the text segment of
+// sections that need some sort of loading and/or relocation.
 struct hdr {
-  //  0: Offset of GOT symbols in flash
+  //  0: Offset of GOT symbols in flash from the start of the application
+  //     binary.
   uint32_t got_sym_start;
-  //  4: Offset of GOT section in memory
+  //  4: Offset of where the GOT section needs to be placed in memory from the
+  //     start of the application's memory region.
   uint32_t got_start;
-  //  8: Size of GOT section
+  //  8: Size of GOT section.
   uint32_t got_size;
-  // 12: Offset of data symbols in flash
+  // 12: Offset of data symbols in flash from the start of the application
+  //     binary.
   uint32_t data_sym_start;
-  // 16: Offset of data section in memory
+  // 16: Offset of where the data section needs to be placed in memory from the
+  //     start of the application's memory region.
   uint32_t data_start;
-  // 20: Size of data section
+  // 20: Size of data section.
   uint32_t data_size;
-  // 24: Offset of BSS section in memory
+  // 24: Offset of where the BSS section needs to be placed in memory from the
+  //     start of the application's memory region.
   uint32_t bss_start;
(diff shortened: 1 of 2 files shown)


libtock/crt0.c
```diff
@@ -228,37 +238,82 @@ void _start(void* app_start __attribute__((unused)),
...
+//   app.
 __attribute__((noreturn))
-void _c_start(uint32_t* app_start, uint32_t stacktop) {
+void _c_start(uint32_t app_start, uint32_t mem_start) {
   struct hdr* myhdr = (struct hdr*)app_start;
 
-  // fix up GOT
-  volatile uint32_t* got_start     = (uint32_t*)(myhdr->got_start + stacktop);
-  volatile uint32_t* got_sym_start = (uint32_t*)(myhdr->got_sym_start + (uint32_t)app_start);
+  // Fix up the Global Offset Table (GOT).
```
brghena, line 253: I'm confused. What's even in the GOT for non-PIC applications? Is `got_size` just set to zero?

libtock/crt0.c
```diff
@@ -228,37 +238,82 @@ void _start(void* app_start __attribute__((unused)),
...
 
-  // zero BSS
-  char* bss_start = (char*)(myhdr->bss_start + stacktop);
+  // Zero BSS segment. Again, we know where this should be in the process RAM
+  // based on the crt0 header.
+  char* bss_start = (char*)(myhdr->bss_start + mem_start);
   memset(bss_start, 0, myhdr->bss_size);
 
+  // Do relative data address fixups. We know these entries are stored at the end
+  // of flash and can be located using the crt0 header.
```
brghena, line 295: `rd->len` is hopefully also zero for non-PIC apps?

libtock/crt0.c
bradjc: No. I wouldn't have found this issue otherwise :)

brghena: But all the fixup in the for loop only works for PIC code compiled with the 0x80000000 and 0x00000000 trick. So either `myhdr->got_size` is zero in practice, the GOT is being fixed up improperly for non-PIC code, or I'm misunderstanding something.

bradjc: Ok yeah I'm getting confused about what is in what PR. Yes, RISC-V apps have entries in the GOT. It seems like it might have something to do with global variables. No, they do not currently work in libtock-c, for exactly the reason you mentioned. The riscv-for-all PR includes two implementations for c_start, one with PIC, one without. However, that change makes the diff worse, so I didn't include it in this PR.

libtock/crt0.c
bradjc: So far, yes.

libtock/crt0.c
brghena: Alright, I'll prod over there about this issue then.

PR #91, opened 2020-06-24
Mark ORIGIN(SRAM) in the linker script

bradjc:
This generates a symbol that elf2tab uses to check for fixed addresses.

### DIFF
--- a/userland_generic.ld
+++ b/userland_generic.ld
@@ -110,7 +110,7 @@ SECTIONS {
 		 * elf2tab. elf2tab will use it to add a fixed address header in the
 		 * TBF header if needed.
 		 */
-	    _SRAM_ORIGIN = .;
+	    _sram_origin = .;
 
         /* Be conservative about our alignment for the stack. Different
          * architectures require different values (8 for ARM, 16 for RISC-V),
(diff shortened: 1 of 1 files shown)


userland_generic.ld
```diff
@@ -103,6 +103,15 @@ SECTIONS {
...
     .stack :
     {
+        /* elf2tab requires that the `_SRAM_ORIGIN` symbol be present to
+		 * mark the first address in the SRAM memory. Since ELF files do
+		 * not really need to specify this address as they only care about
+		 * loading into flash, we need to manually mark this address for
+		 * elf2tab. elf2tab will use it to add a fixed address header in the
+		 * TBF header if needed.
+		 */
+	    _SRAM_ORIGIN = .;
```
ppannuto, line 113:
```suggestion
	    _sram_origin = .;
```

I don't necessarily feel strongly about this, but all of the ALL_CAPS variables in this file are used by the linker to control its operations, and all of the external symbols are `_underscore_lowercase`; better to match?

bradjc: Ha good point. I don't know why I was thinking caps. Probably because `ORIGIN(SRAM)` is all caps above.

PR #102, opened 2020-08-14
Add a note about size optimised newlib-nano

alistair23:
Add a note for the RISC-V setup about a size optimsed newlib-nano.

The reason this is added as a note and not the default is because newlib packaged by most distros ([Debian](https://salsa.debian.org/debian/newlib/-/blob/master/debian/rules#L62) and OE at least) does not include these nano flags. For this reason we are telling users to use the more standard flags and giving them the option for the size reduced version. The hope for RISC-V is to eventually just point users to their distro packaged newlib, so let's stick as close to that as we can.

Tested on OpenTitan by running c_hello. In order to run on current Tock I need a OT app address change which is included in this PR.


README.md
```diff
@@ -111,6 +111,29 @@ Prerequisites
    $ sudo pacman -Syu riscv64-elf-gcc riscv32-elf-newlib
    ```
 
+   **newlib-nano**:
```
hudson-ayers, line 114: In the section 30 lines earlier that reads "If you would rather compile your own newlib-based libc, follow these steps:", could we add a pointer to this section so that people know to add these flags if they want the size optimized version?

PR #103, opened 2020-08-15
Basic C interface to the Rubble capsule

daboross:
This adds an interface to the rubble capsule added in https://github.com/tock/tock/pull/2036.

The rubble capsule currently exposes very little functionality, and this interface is thus equally minimalistic. I've based it, and the test app, on a reduced version of the existing ble_advertising interface and test app.

### DIFF
--- a/examples/rubble_ble_advertising/main.c
+++ b/examples/rubble_ble_advertising/main.c
@@ -1,0 +1,47 @@
+#include <gap.h>
+#include <rubble_ble.h>
+#include <stdbool.h>
+#include <stdio.h>
+#include <tock.h>
+
+// Sizes in bytes
+#define DEVICE_NAME_SIZE 10
+
+/*******************************************************************************
+ * MAIN
+ ******************************************************************************/
+
+int main(void) {
+  int err;
+  printf("[Tutorial] BLE Advertising\n");
+
+  // declarations of variables to be used in this BLE example application
+  uint16_t advertising_interval_ms = 300;
+  uint8_t device_name[]            = "Rubble App";
+
+  static uint8_t adv_data_buf[ADV_DATA_MAX_SIZE];
+
+  // configure advertisement interval to 300ms
+  // configure LE only and discoverable
+  printf(" - Initializing BLE... %s\n", device_name);
+  AdvData_t adv_data = gap_adv_data_new(adv_data_buf, sizeof(adv_data_buf));
+
+  gap_add_flags(&adv_data, LE_GENERAL_DISCOVERABLE | BREDR_NOT_SUPPORTED);
+
+  // configure device name as TockOS
+  printf(" - Setting the device name... %s\n", device_name);
+  err = gap_add_device_name(&adv_data, device_name, DEVICE_NAME_SIZE);
+  if (err < TOCK_SUCCESS)
+    printf("rubble_advertise_name, error: %s\r\n", tock_strerror(err));
+
+  // start advertising
+  printf(" - Begin advertising! %s\n", device_name);
+  err = rubble_start_advertising(adv_data.buf, adv_data.offset, advertising_interval_ms);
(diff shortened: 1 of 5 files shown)


libtock/rubble_ble.h
```diff
@@ -0,0 +1,33 @@
...
+
+#define RUBBLE_DRIVER_NUMBER 0x30003
+
+#define RUBBLE_ADV_START_CMD 0
+
+#define RUBBLE_ADV_BUF_ALLOW 0
+
+// start advertising
+//
+// advd               - The advertising data. Must be formatted properly
```
ppannuto, line 26: Should this be some form of struct rather than a byte array to help guide/enforce this?

examples/rubble_ble_advertising/main.c
```diff
@@ -0,0 +1,47 @@
...
+ * MAIN
+ ******************************************************************************/
+
+int main(void) {
+  int err;
+  printf("[Tutorial] BLE Advertising\n");
+
+  // declarations of variables to be used in this BLE example application
+  uint16_t advertising_interval_ms = 300;
+  uint8_t device_name[]            = "Rubble App";
```
ppannuto, line 20:
```suggestion
  uint8_t device_name[]            = "Rubble on Tock";
```

Maybe?

examples/rubble_ble_advertising/main.c
```diff
@@ -0,0 +1,47 @@
...
+// Sizes in bytes
+#define DEVICE_NAME_SIZE 10
+
+/*******************************************************************************
+ * MAIN
+ ******************************************************************************/
+
+int main(void) {
+  int err;
+  printf("[Tutorial] BLE Advertising\n");
```
ppannuto, line 16:
```suggestion
  printf("[Example] BLE Advertising\n");
```

Maybe?

libtock/rubble_ble.h
daboross: I'd definitely be interested in adding this in one form or another! But I'm also unsure if it makes sense to right now?

The rubble capsule's minimal interface is mostly in  MVP. I'd like to expose more - maybe even an interface matching rubble's Rust one - but that adds a lot of parsing code that it doesn't otherwise need, so I was thinking it could make more sense as a followup improvement.

The other alternative I know would be to make a struct-based interface in `libtock-c` which simply encodes the response back into bytes. If adding this, I'm not sure if it actually adds any value compared to just using and highly recommending `simple-ble` functions? As I understand it `gap_*` functions should always create valid data, so it seems like we'd just be duplicating that interface. With that said, I'm not a C paradigm expert, so if you think it'd make sense to enforce that with structs at this interface (and possibly just call into `simple-ble`, or duplicate its code?), I'd be down to do that.

ppannuto: Mhmm, in the short term, leaving it flexible probably makes the most sense. I'll confess that I don't know the simple_ble stack as well as @brghena -- Branden, is there some form of `ble_adv_data` type that we should consider putting in the interface here, or is just a byte-array appropriate?

phil-levis:
Sorry, catching up with this now.

Why do we want a Rubble BLE interface *and* a BLE interface? I.e., why doesn't Rubble just implement the API in `ble.h`? Is this so we can have two implementations in parallel?

alistair23:
I would say the main benefit of having two interfaces is that we can then support both implementations in parallel.

I think we should work on removing the current BLE implementation and replace it with Rubble, but that won't happen straight away even if at all.

ppannuto:
I think @phil-levis's question is in the same vein as the discussion currently going on at https://github.com/tock/tock/pull/2233. We've historically held the libtock-c merges until the kernel merges are ready/done -- but I think this is fine once the other half is ready.

ppannuto:
Rubble project EOL'd mid-2022 :(.

PR #104, opened 2020-09-02
Userland test app and libs for new proximity sensor interface on Tock

arjundeopujari:
Tested sensors/main.c and test/proximity/main.c on a Nano33ble board with tock kernel here --> https://github.com/arjundeopujari/tock-1.git (currently in pull request mode and will be on official tock github page in a while).

### DIFF
--- a/.travis-install-gcc
+++ b/.travis-install-gcc
@@ -10,3 +10,7 @@ if [ ! -x gcc-arm-none-eabi-9-2019-q4-major/bin/arm-none-eabi-gcc ]; then
   tar -xjvf gcc.tar.bz2
 fi
 
+if [ ! -x gcc-riscv64-unknown-elf-8.3.0-ubuntu/bin/riscv64-unknown-elf-gcc ]; then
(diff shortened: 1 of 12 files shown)


examples/tests/proximity/main.c
```diff
@@ -0,0 +1,57 @@
...
+    
+    while (true){
+
+        for (int led = 0; led < numLeds; led++){
+            led_on(led);
+            delay_ms(period / frequency);
+            led_off(led);
+        }
+
+        proximity_read_sync(&frequency);
```
ppannuto, line 49: This should probably check the return value here

ppannuto (changes requested):
I think the logic looks mostly good here.

There's a bunch of unrelated whitespace changes that need to be removed (they're also probably why CI is failing).

Can you please also remove the changes to the `simple-ble` library? I don't think they belong in this PR.

---

With those fixes, I think this should be good to go once the kernel stuff is in.

PR #107, opened 2020-09-16
Adds tests for the new QDEC Peripheral in tockOS

mmurray22:
### Pull Request Overview

This pull request contains a new userspace library and two new userspace test apps to test and demonstrate to features of the new QDEC Peripheral being added to tockOS.

### Testing Strategy

This pull request was tested and verified by Michaela Murray (author) and Hudson Ayers.

### TODO

N/a

### Documentation Updated

- Any documentation related to listing all available peripherals may need to be updated.

### Formatting

- [ ] Ran `make formatall`.


libtock/qdec.h
```diff
@@ -0,0 +1,19 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#define DRIVER_NUM_QDEC 0x90003
+
+bool qdec_exists(void);
+bool qdec_enable(void);
+bool qdec_interrupt_enable(void);
```
ppannuto, line 13: This may be more of an issue to address on the kernel side, but from a userspace perspective, this is a pretty low-level interface -- how often do interrupts fire (every partial rotation? after some displacement or interval?).

More generally, in the header file here, some documentation on how each of these functions expect to be called / what they expect to return would be helpful.

examples/tests/qdec_callback/main.c
```diff
@@ -0,0 +1,46 @@
...
+    led_off(1);
+    led_off(2);
+  }  else if (qdec_num == 0) {
+    led_on(1);
+    led_off(0);
+    led_off(2);
+  } else if (qdec_num < 0) {
+    led_on(2);
+    led_off(0);
+    led_off(1);
```
ppannuto, line 25: Generally, Tock only requires/expects platforms to have 1 LED (though many do have more). It would be nice if this could be agnostic to the number of LEDs as a test (i.e. by varying blink rate).

ppannuto (changes requested):
Always cool to get a new driver!

---

For each of the test cases, can you add a `README.md` that explains the purpose of the test (i.e. what functionality of the kernel/driver is it testing) as well as some explanation of how to evaluate whether the test is running correctly and example output from "good" test run?

examples/tests/qdec_callback/main.c
mmurray22: This was a point of concern when I first wrote this, so this makes sense. I will update it for the next round of review.

libtock/qdec.h
mmurray22: Interrupts fire every partial rotation. Each `tick` of the rotary encoder is equivalent to 4 clicks, and the interrupt comes at each of those clicks. I will make sure to add some helpful documentation for the next round of review.

libtock/qdec.h
```diff
@@ -0,0 +1,19 @@
+#pragma once
```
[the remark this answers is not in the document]
mmurray22: Will do-thanks for pointing this out.

examples/tests/qdec_callback/main.c
mmurray22: Ended up removing the leds from the code altogether since it did not seem relevant to the overall goal of the tests.

mmurray22:
Added in the READMEs. Will probably continue expanding on them as this PR progresses!

PR #109, opened 2020-10-06
Small fixes for running on OpenTitan

alistair23:


Configuration.mk
```diff
@@ -100,7 +100,8 @@ override CPPFLAGS += \
       -Wall\
       -Wextra\
       -Wl,--warn-common\
-      -Wl,--gc-sections
+      -Wl,--gc-sections\
+      -Wl,--build-id=none
```
ppannuto, line 104: Huh, I hadn't heard of build-id before, but can see that it could be useful -- should we consider following this approach instead ( https://stackoverflow.com/a/55960670 ) and moving it?

alistair23: I can't think of a reason to keep it around, it just takes up space in the binary.

hudson-ayers (approved):

PR #110, opened 2020-10-09
sensors app: actually read ninedof

bradjc:
Actually read the ninedof sensors.


hudson-ayers:
This is probably good, but worth keeping in mind this will make this app fail on all the new Imixes that don't have the accelerometer

bradjc:
If you look further down in the file the app is already calling the ninedof read functions.

hudson-ayers (approved):

ppannuto (approved):
bors r+

PR #112, opened 2020-10-10
hail app: adc samples in upper bits

bradjc:
The ADC always returns samples as 16 bits, even though the ADC may use
fewer bits. This fixes the calculation since the 12 bits of SAM4L adc
will always be the upper 12 bits.

### DIFF
--- a/examples/tests/hail/main.c
+++ b/examples/tests/hail/main.c
@@ -97,17 +97,17 @@ static void sample_sensors (void) {
   // Analog inputs: A0-A5
   uint16_t val;
   adc_sample_sync(0, &val);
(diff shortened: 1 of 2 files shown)


hudson-ayers:
This looks correct, but we should make this change for Imix as well. 

As an aside, it makes very little sense to me that we store the raw ADC values in the upper bits rather than the lower, but I am probably missing something?

hudson-ayers (approved):

bradjc:
The benefit is you can treat all ADC samples as 16 bit no matter what the actual ADC resolution is, and only loose some resolution in the less significant bits.

PR #115, opened 2020-10-20
Allow toolchains to bet set

### LINKED ISSUE #114: Toolchain configuration
Is there a reason why the `TOOLCHAIN_*` variables are set with the `override` prefix in `Configuration.mk`?

The values for `TOOLCHAIN_rv32im[a]c` are currently `riscv64-unknown-elf`.
Some packages (e.g. Arch, which is also referenced in the README.md) have binaries which have a prefix `riscv64-elf`.
It should also be possible to compile with `riscv32-unknown-elf-gcc`.
Not using `override` would help to configure which compiler is used locally.

The configuration dump of `Configuration.mk` uses the variable `TOOLCHAIN` but as far as I can see it is never used, to this output is a bit confusing.

towoe:
Remove keyword `override` so the variables are changeable with make
arguments.

Fixes #114

### DIFF
--- a/Configuration.mk
+++ b/Configuration.mk
@@ -80,11 +80,11 @@ ELF2TAB_ARGS += -n $(PACKAGE_NAME)
 ELF2TAB_ARGS += --stack $(STACK_SIZE) --app-heap $(APP_HEAP_SIZE) --kernel-heap $(KERNEL_HEAP_SIZE)
 
 # Setup the correct toolchain for each architecture.
-override TOOLCHAIN_cortex-m0 = arm-none-eabi
-override TOOLCHAIN_cortex-m3 = arm-none-eabi
-override TOOLCHAIN_cortex-m4 = arm-none-eabi
-override TOOLCHAIN_rv32imac = riscv64-unknown-elf
-override TOOLCHAIN_rv32imc = riscv64-unknown-elf
+TOOLCHAIN_cortex-m0 = arm-none-eabi
+TOOLCHAIN_cortex-m3 = arm-none-eabi
+TOOLCHAIN_cortex-m4 = arm-none-eabi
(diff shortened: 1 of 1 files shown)


Configuration.mk
```diff
@@ -80,11 +80,11 @@ ELF2TAB_ARGS += -n $(PACKAGE_NAME)
...
-override TOOLCHAIN_cortex-m0 = arm-none-eabi
-override TOOLCHAIN_cortex-m3 = arm-none-eabi
-override TOOLCHAIN_cortex-m4 = arm-none-eabi
-override TOOLCHAIN_rv32imac = riscv64-unknown-elf
-override TOOLCHAIN_rv32imc = riscv64-unknown-elf
+TOOLCHAIN_cortex-m0 = arm-none-eabi
+TOOLCHAIN_cortex-m3 = arm-none-eabi
+TOOLCHAIN_cortex-m4 = arm-none-eabi
+TOOLCHAIN_rv32imac = riscv64-unknown-elf
+TOOLCHAIN_rv32imc = riscv64-unknown-elf
```
ppannuto, line 87:
```suggestion
TOOLCHAIN_cortex-m0 := arm-none-eabi
TOOLCHAIN_cortex-m3 := arm-none-eabi
TOOLCHAIN_cortex-m4 := arm-none-eabi
TOOLCHAIN_rv32imac := riscv64-unknown-elf
TOOLCHAIN_rv32imc := riscv64-unknown-elf
```

While we're fixing things, we generally try avoid lazy binding in our makefiles... not really sure what happened to TOOLCHAIN :/

ppannuto (changes requested):
image
Well this is about to be a silly cycle

ppannuto (approved):
bors r+

towoe:
Thanks for the follow up commit @ppannuto

PR #116, opened 2020-10-21
fix race condition in UDP virtualization tests

hudson-ayers:
These tests worked originally, but depended on one app executing certain lines of code first, for the portion where both apps attempt to bind to the same port. I used a 10ms delay to achieve this. However, because the time to send a multi-fragment UDP packet is longer than 10ms, this test only worked originally because app1 happened to always be scheduled first on the old priority scheduler. For a 10ms delay to be enough, it must be placed before the packet is sent, ensuring that app1 sends its packet first.

Kinda a confusing explanation, but basically this test assumed an order of execution which was not exposed until we switched to a round robin scheduler on Imix. This PR fixes it.


ppannuto (approved):
bors d+

Maybe a comment around the delay explaining this?

PR #120, opened 2020-11-20
Add Cortex-M7 target

mciantyre:
The PR adds **Cortex-M7** support for libtock-c. As of this PR, we can build Tock applications in C for M7 targets. See tock/tock#1918 and tock/tock#2200 for more information on M7 architecture support in Tock.

I tested the `blink` and `console` examples on a Cortex-M7 system as part of tock/tock#2200. I showed that all of the examples now include M7 outputs:

```
cd examples && ./build_all.sh
```


hudson-ayers (changes requested):
Can you also add `cortex-m7` to `README.md` in the line where it mentions all architectures that examples will be built for, and add `cortex-m7` to `TOCK_TARGETS` in `lua-hello/Makefile`?

hudson-ayers (approved):

PR #129, opened 2020-12-14
UDP driver to 2.0 syscalls

hudson-ayers:
Updates the UDP driver to use the 2.0 system calls. Mostly works, but matching on error codes does not seem to work, I think this will need to be updated once the new error types are inserted in libtock-c

### DIFF
--- a/examples/tests/udp/udp_rx/main.c
+++ b/examples/tests/udp/udp_rx/main.c
@@ -65,7 +65,11 @@ int main(void) {
   printf(" : %d, and binding to that socket.\n", addr.port);
   sock_handle_t h;
   handle = &h;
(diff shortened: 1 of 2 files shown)


phil-levis (approved):

hudson-ayers:
What is the strategy now that the full 2.0 interface is merged in libtock-c? Should I update this to use command2 / subscribe2 / etc. ?

phil-levis:
> What is the strategy now that the full 2.0 interface is merged in libtock-c? Should I update this to use command2 / subscribe2 / etc. ?

Probably a good idea.

hudson-ayers:
updated to use the new system calls, and tested using the example apps, which work

PR #134, opened 2020-12-18
Port DAC library to Tock 2.0 system call interface

alevy:
This pull request ports the DAC driver to the new Tock 2.0 system call interface. Tested using the `dac` test on Hail with the related changes in to the Tock kernel: tock/tock#2285.

### DIFF
--- a/examples/tests/dac/main.c
+++ b/examples/tests/dac/main.c
@@ -18,17 +18,20 @@ uint16_t sine_samples[100] = {
 };
 
 int main(void) {
-  int ret;
+  syscall_return_t ret;
 
   printf("[DAC] Sine test app\n");
 
   ret = dac_initialize();
-  if (ret != 0) printf("ERROR initializing DAC\n");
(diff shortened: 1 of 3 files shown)


hudson-ayers (approved):

examples/tests/dac/main.c
```diff
@@ -18,17 +18,20 @@ uint16_t sine_samples[100] = {
...
 
 int main(void) {
-  int ret;
+  syscall_return_t ret;
 
   printf("[DAC] Sine test app\n");
 
   ret = dac_initialize();
-  if (ret != 0) printf("ERROR initializing DAC\n");
+  if (ret.type < TOCK_SYSCALL_SUCCESS) {
```
lschuermann, line 26: I think you'd rather match on the precise variant here. _Less than_ comparison will probably work given the enum values, but that seems more like a coincidence in how the values are chosen. Also, each command is only supposed to have one failure and success variant, which would make matching on the exact value more expressive.

alevy: Fixed

PR #142, opened 2020-12-20
Fix erroneous use of syscall_return_t.data[1] for fetching error code (should be data[0])

phil-levis:
@hudson-ayers Caught a bunch of places where code incorrectly grabbed the second value (data[1]) instead of the first (data[0]) for the error code of a 2.0 command invocation. This fixes them.


hudson-ayers (approved):
Thanks for submitting this!

bors r+

PR #143, opened 2020-12-21
Update Ieee802154 driver to Tock 2.0 syscalls

hudson-ayers:
This has been tested using the radio_rx and radio_tx example apps, which successfully exchange frames between an imix and an nrf52840dk.

One notable change as part of this is the need to explicitly unallow the receive buffer before accessing it again -- a task which previously was left up to the kernel capsule.


phil-levis:
Poke -- can you get this to pass Travis?

hudson-ayers:
done, sorry about that

PR #145, opened 2020-12-22
Tock 2.0: update led driver to Tock 2.0 API

alexandruradovici:
The pull request updates the `led` driver to the Tock 2.0 API.

### DIFF
--- a/libtock/led.c
+++ b/libtock/led.c
@@ -1,17 +1,37 @@
 #include "led.h"
 
 int led_count(void) {
-  return command(DRIVER_NUM_LEDS, 0, 0, 0);
+  syscall_return_t ret = command2(DRIVER_NUM_LEDS, 0, 0, 0);
+  if (ret.type == TOCK_SYSCALL_SUCCESS_U32) {
+    return ret.data[0];
+  } else {
+    return 0;
+  }
 }
 
 int led_on(int led_num) {
-  return command(DRIVER_NUM_LEDS, 1, led_num, 0);
+  syscall_return_t ret = command2(DRIVER_NUM_LEDS, 1, led_num, 0);
+  if (ret.type == TOCK_SYSCALL_SUCCESS) {
+    return TOCK_SUCCESS;
+  } else {
+    return ret.data[0];
+  }
 }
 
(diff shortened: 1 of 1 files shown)


libtock/led.c
```diff
@@ -1,17 +1,37 @@
...
+  }
 }
 
 int led_on(int led_num) {
-  return command(DRIVER_NUM_LEDS, 1, led_num, 0);
+  syscall_return_t ret = command2(DRIVER_NUM_LEDS, 1, led_num, 0);
+  if (ret.type == TOCK_SYSCALL_SUCCESS) {
+    return TOCK_SUCCESS;
+  } else {
+    return ret.data[0];
```
hudson-ayers, line 17: if the LED capsule mistakenly returned a `success_u32()`, rather than sucess, this would just return some arbitrary value rather than a valid failure return code. @phil-levis do you consider it acceptable for userspace drivers to assume the correct success / failure types rather than explicitly checking for them?

alexandruradovici: Just a thought, but what about having a macro IS_SUCCESS that would test for all the SUCCESS cases?

hudson-ayers: After looking at the alarm porting PR this seems like the expected approach, so marking as resolved

hudson-ayers: Oh sorry didn't see your comment. The issue is that a given subcommand should only return a single success type, so I think you would want multiple macros (one expecting `TOCK_SYSCALL_SUCCESS`, one expecting `TOCK_SUCCESS_U32`, and so forth). But I am not sure if the ability to write slightly terser code is worth the added indirection

lschuermann: > a macro IS_SUCCESS that would test for all the SUCCESS cases?

We explicitly don't want that. A command is only allowed to have a single success and failure variant, and as such all other success variants must be treated as an error, ideally a `BADRVAL` in userspace (as done in `libtock-rs`, as far as I know).

hudson-ayers:
This currently overlaps with changes in #147 , I'm gonna let @phil-levis decide which to use

phil-levis:
Subsumed by #147

PR #146, opened 2020-12-22
Update RNG to Tock 2.0 system call interface

lschuermann:
### Pull Request Overview

This pull request migrates the RNG userspace driver to use the Tock 2.0 system call interface, as introduced for the RNG capsule [here](https://github.com/tock/tock/pull/2297).

In contrast to the kernel side, I don't know the userspace Tock 2.0 type infrastructure by heart, so there might be some errors.

Probably noteworthy is that I explicitly match on both the success and error syscall return variant. A capsule _can technically_ return more than a single success and failure variant (although per definition it must not). Simply checking for equality in the success case might cause UB as a result of accessing fields which are believed to have been set by the kernel.

Furthermore, I use `exit(-1)` for those unrecoverable cases where the kernel behaves not according to the specification.

### Testing Strategy

This pull request was tested using a Hail, printing obtained randomness through the accompanying [kernel PR](https://github.com/tock/tock/pull/2297).

### TODO or Help Wanted

N/A

### DIFF
--- a/libtock/rng.c
+++ b/libtock/rng.c
@@ -1,12 +1,15 @@
-#include <rng.h>
+#include <stdlib.h>
 #include <tock.h>
+#include <rng.h>
 
 struct rng_data {
   bool fired;
   int received;
 };
 
-static struct rng_data result = { .fired = false, .received = 0};
+// Global state for faking synchronous reads using a callback and
+// yield
+static struct rng_data result = { .fired = false, .received = 0 };
 
 // Internal callback for faking synchronous reads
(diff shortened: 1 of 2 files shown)


hudson-ayers (approved):

phil-levis:
This is just a formatting problem. libtock-c needs a `make prepush`.

lschuermann:
> This is just a formatting problem. libtock-c needs a `make prepush`.

Thanks for merging nonetheless! I actually didn't notice the CI failure and didn't know we had a formatter in `libtock-c`. :)

PR #147, opened 2020-12-24
Updates alarm to use 2.0 system calls.

phil-levis:
This PR makes three changes:
  - It updates the alarm/timer library to use the 2.0 system calls and their return values
  - It updates the alarm/timer library to use the system call that passes a reference and a dt, rather than an absolute time for an alarm; this prevents bugs where the absolute time has fallen behind now and the kernel interprets this as an alarm in the far far future
  - It also (trivially) updates the LED library routines to use the new system calls. This is important for led_count(), which the alarm test applications depend on; without this fix they crash by malloc'ing too many alarms.

### DIFF
--- a/libtock/led.c
+++ b/libtock/led.c
@@ -5,33 +5,33 @@ int led_count(void) {
   if (rval.type == TOCK_SYSCALL_SUCCESS_U32) {
     return rval.data[0];
   } else {
-    return 0;
+    return tock_error_to_rcode(rval.data[0]);
   }
 }
 
 int led_on(int led_num) {
   syscall_return_t rval = command2(DRIVER_NUM_LEDS, 1, led_num, 0);
   if (rval.type == TOCK_SYSCALL_SUCCESS) {
-    return 0;
+    return TOCK_SUCCESS;
   } else {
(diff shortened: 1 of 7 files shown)


libtock/led.c
```diff
@@ -1,17 +1,37 @@
...
 }
 
 int led_toggle(int led_num) {
-  return command(DRIVER_NUM_LEDS, 3, led_num, 0);
+  syscall_return_t rval = command2(DRIVER_NUM_LEDS, 3, led_num, 0);
+  if (rval.type == TOCK_SYSCALL_SUCCESS) {
+    return 0;
+  } else {
+    return 1;
+  }
```
hudson-ayers, line 36: These just return 1 rather than the actual error codes in the case of failure, so are different than the proposed changes in #145 . Not sure which I think is better, just pointing that out

hudson-ayers (approved):

libtock/led.c
phil-levis: Good point. I might have encountered a bug when running a long-term test; I am trying to reproduce it now. Once that resolves I'll fix this.

PR #126, opened 2020-12-11
Touch API

alexandruradovici:
The PR adds the user space API for the touch driver (https://github.com/tock/tock/pull/2060).

PR #151, opened 2020-12-30
Tock 2.0: touch api and example app for 2.0 API

alexandruradovici:
This pull request adds the touch driver API and example.

It should close #126

### DIFF
--- a/libtock/touch.c
+++ b/libtock/touch.c
@@ -1,8 +1,6 @@
 #include "touch.h"
 #include <stdlib.h>
 
-#define MULTI_TOUCH_LENGTH 8
-
 static int touch_subscribe(int subscribe_num, subscribe_cb callback, void* callback_args) {
   subscribe_return_t sv = subscribe2(DRIVER_NUM_TOUCH, subscribe_num, callback, callback_args);
   if (sv.success == 0) {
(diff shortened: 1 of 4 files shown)


libtock/touch.c
```diff
@@ -0,0 +1,130 @@
...
+          free (multi_touch_buffer);
+          multi_touch_buffer = NULL;
+        }
+      }else {
+        err = TOCK_ENOMEM;
+      }
+    }else {
+      err = TOCK_EALREADY;
+    }
+  }else {
```
phil-levis, line 74: Can you run `format_all.sh` in `examples/` to format the code?

libtock/touch.h
```diff
@@ -0,0 +1,41 @@
...
+#define GESTURE_SWIPE_RIGHT 4
+#define GESTURE_ZOOM_IN 5
+#define GESTURE_ZOOM_OUT 6
+
+typedef void (touch_callback)(int, int, int, void*);
+typedef void (gesture_callback)(int, void*);
+
+// buffer data format
+//  0         1           2                  4                  6           7             8         ...
+// +---------+-----------+------------------+------------------+-----------+---------------+--------- ...
```
phil-levis, line 26: Can you define a structure with this format? Hardcoded offsets are buggy.

phil-levis (changes requested):

libtock/touch.h
alexandruradovici: Your suggestion would be to use a packed structure instead of the buffer?

libtock/touch.c
alexandruradovici: I did, this is what it does. I actually ran `make format`. I fixed the other `}else  {` manually, but I missed this one :)

phil-levis (approved):

PR #154, opened 2021-01-08
Tock 2.0/rv32i: fix memop calls in _start

lschuermann:
### Pull Request Overview

This pull request
- fixes documentation for the initial memop calls as part of the `_start` function on RISC-V userland applications.
- fixes a memop system call number (in register `a4`) which changed as part of the Tock 2.0 transition
- adjusts the memop wrapper function to follow the Tock 2.0 system call semantics
- fixes the Tock 1.x system call wrapper functions to work while we're still porting

### Testing Strategy

This pull request was tested by running a userspace app on Tock `tock-2.0-dev` and this patch on RISC-V.

### TODO or Help Wanted

N/A

### DIFF
--- a/libtock/tock.c
+++ b/libtock/tock.c
@@ -441,16 +441,22 @@ allow_ro_return_t allow_readonly(uint32_t driver, uint32_t allow, const void* pt
 }
 
 void* memop(uint32_t op_type, int arg1) {
-  register uint32_t a0  asm ("a0") = op_type;
-  register uint32_t a1  asm ("a1") = arg1;
-  register void*    ret asm ("a0");
+  register uint32_t a0    asm ("a0") = op_type;
+  register int a1         asm ("a1") = arg1;
+  register void* val      asm ("a1");
+  register uint32_t code  asm ("a0");
   asm volatile (
     "li    a4, 5\n"
     "ecall\n"
-    : "=r" (ret)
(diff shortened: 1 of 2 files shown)


hudson-ayers (approved):

lschuermann:
Actually, I just tested `c_hello` and assumed things were fine otherwise. Turns out, there's some thing to update for apps to work correctly with RISC-V, so I'll add some commits.

hudson-ayers (approved):

libtock/tock.c
```diff
@@ -441,16 +441,22 @@ allow_ro_return_t allow_readonly(uint32_t driver, uint32_t allow, const void* pt
...
+    : "=r" (code), "=r" (val)
     : "r" (a0), "r" (a1)
-    : "memory");
-  return ret;
+    : "memory"
+    );
+  if (code == TOCK_SYSCALL_SUCCESS_U32) {
+    return val;
+  } else {
+    return 0;
```
alistair23, line 458: Why not `return NULL`?

lschuermann: Good question! I copied the code from ARM. I don't believe it makes a difference on RISC-V or ARM, but you're right in that it'd fit the type system better.

alistair23 (approved):

PR #159, opened 2021-01-21
Tock 2.0: nonvolatile storage & app flash driver

lschuermann:
### Pull Request Overview

This pull request ports both the nonvolatile storage driver and the app flash driver to the Tock 2.0 system call interface. It's the counterpart to https://github.com/tock/tock/pull/2362.

It furthermore adds a `BADRVAL` error, as indicated in the Tock 2.0 TRD. Drivers can return this error code (with an assigned value of `-1024`) in case a kernel driver returns an unexpected variant which they do not know how to handle.

### Testing Strategy

This pull request was tested by running the `nonvolatile_storage` and `app_state` tests on the accompanying [kernel PR](https://github.com/tock/tock/pull/2362).

### TODO or Help Wanted

N/A

### Formatting

- [x] Ran uncrustify.

### DIFF
--- a/libtock/internal/nonvolatile_storage_internal.c
+++ b/libtock/internal/nonvolatile_storage_internal.c
@@ -1,31 +1,52 @@
 #include "internal/nonvolatile_storage.h"
 
-int nonvolatile_storage_internal_read_done_subscribe(subscribe_cb cb, void *userdata) {
-  return subscribe(DRIVER_NUM_NONVOLATILE_STORAGE, 0, cb, userdata);
+subscribe_return_t nonvolatile_storage_internal_read_done_subscribe(subscribe_cb cb, void *userdata) {
+  return subscribe2(DRIVER_NUM_NONVOLATILE_STORAGE, 0, cb, userdata);
 }
 
-int nonvolatile_storage_internal_write_done_subscribe(subscribe_cb cb, void *userdata) {
-  return subscribe(DRIVER_NUM_NONVOLATILE_STORAGE, 1, cb, userdata);
+subscribe_return_t nonvolatile_storage_internal_write_done_subscribe(subscribe_cb cb, void *userdata) {
+  return subscribe2(DRIVER_NUM_NONVOLATILE_STORAGE, 1, cb, userdata);
 }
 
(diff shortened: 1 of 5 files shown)


alistair23 (approved):

libtock/tock.h
```diff
@@ -36,7 +36,7 @@ typedef enum {
   TOCK_ERROR_NOSUPPORT   = 9,
   TOCK_ERROR_NODEVICE    = 10,
   TOCK_ERROR_UNINSTALLED = 11,
-  TOCK_ERROR_NOACK       = 12
+  TOCK_ERROR_NOACK       = 12,
```
hudson-ayers, line 39: Why not also include `ERROR_BADRVAL` here?

libtock/internal/nonvolatile_storage_internal.c
```diff
@@ -1,31 +1,52 @@
...
+  } else {
+    return TOCK_EBADRVAL;
+  }
 }
 
 int nonvolatile_storage_internal_read(uint32_t offset, uint32_t length) {
-  uint32_t arg0 = (length << 8) | 2;
-  return command(DRIVER_NUM_NONVOLATILE_STORAGE, (int) arg0, (int) offset, 0);
+  syscall_return_t res =
+    command2(DRIVER_NUM_NONVOLATILE_STORAGE, 2, (int) offset, (int) length);
```
hudson-ayers, line 32: what is this change? I may have just missed a change in the kernel driver

libtock/tock.h
lschuermann: BADRVAL is not an error that can be returned by the kernel. It's just used in userspace that the kernel has returned an unexpected return value.

libtock/internal/nonvolatile_storage_internal.c
lschuermann: Yes. The ABI was changed slightly, as we had sufficient arguments available such that we didn't need to pack with the length and offset into a single register. Is there a good reason not to do this?

hudson-ayers: oh I see. Nope that makes sense!

hudson-ayers (approved):

PR #161, opened 2021-01-22
Yield 2.0

phil-levis:
This updates yield to the 2.0 kernel. It adds new system call, `yield_no_wait` and also adds a new test app in `tests`.

### DIFF
--- a/libtock/tock.c
+++ b/libtock/tock.c
@@ -128,11 +128,11 @@ void yield2(void) {
     // registers r4-r8, r10, r11 and SP (and r9 in PCS variants that designate
     // r9 as v6) As our compilation flags mark r9 as the PIC base register, it
     // does not need to be saved. Thus we must clobber r0-3, r12, and LR
-    register uint32_t wait asm ("r0") = 1; // yield-wait
+    register uint32_t wait asm ("r0")       = 1; // yield-wait
     register uint32_t wait_field asm ("r1") = 0; // yield result ptr
     asm volatile (
       "svc 0       \n"
-      : 
+      :
       : "r" (wait), "r" (wait_field)
       : "memory", "r1", "r2", "r3", "r12", "lr"
       );
@@ -148,7 +148,7 @@ void yield2(void) {
 
 int yield_no_wait() {
   uint8_t result = 0;
-   if (task_cur != task_last) {
+  if (task_cur != task_last) {
     tock_task_t task = task_queue[task_cur];
     task_cur = (task_cur + 1) % TASK_QUEUE_SIZE;
     task.cb(task.arg0, task.arg1, task.arg2, task.ud);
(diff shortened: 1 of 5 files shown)


libtock/tock.c
```diff
@@ -89,12 +90,56 @@ void yield(void) {
...
       "svc 0       \n"
       :
+      : "r" (wait), "r" (wait_field)
+      : "memory", "r2", "r3", "r12", "lr"
+      );
+  }
+}
+
+int yield_no_wait(void) {
+  if (task_cur != task_last) {
```
hudson-ayers, line 105: why not put all this logic into a `yield_internals(int wait)` and then have `yield()` and `yield_no_wait()` just be wrappers that pass the appropriate values for the wait field?

libtock/tock.c
```diff
@@ -277,19 +322,43 @@ void yield(void) {
...
+    register uint32_t a0  asm ("a0") = 1; // yield-wait
     asm volatile (
-      "li    a4, 0\n"
+      "li    a5, 0\n"
       "ecall\n"
       :
-      :
-      : "memory", "a0", "a1", "a2", "a3", "a4", "a5", "a6", "a7",
+      : "r" (a0)
+      : "memory", "a1", "a2", "a3", "a4", "a6", "a7",
```
hudson-ayers, line 331: doesn't this still need to mark a5 as clobbered? Especially now that this snippet directly overwrites it even before the `ecall`?

libtock/tock.c
```diff
@@ -277,19 +322,43 @@ void yield(void) {
...
   } else {
+    register uint32_t a0  asm ("a0") = 1; // yield-wait
     asm volatile (
-      "li    a4, 0\n"
+      "li    a5, 0\n"
       "ecall\n"
       :
-      :
-      : "memory", "a0", "a1", "a2", "a3", "a4", "a5", "a6", "a7",
+      : "r" (a0)
```
hudson-ayers, line 330: the ARM implementation of yield passes both r0 and r1 as inputs, but the risc-v one (this one) does not. I am not sure if it matters, since yield should not use the r1 argument, but it would be nice to use the same approach for ARM and risc-v

libtock/tock.c
```diff
@@ -277,19 +322,43 @@ void yield(void) {
...
+  } else {
+    uint8_t result = 0;
+    register uint32_t a0  asm ("a0") = 0; // yield-no-wait
+    register uint8_t* a1  asm ("a1") = &result;
+    asm volatile (
+      "li    a5, 0\n"
+      "ecall\n"
+      :
+      : "r" (a0), "r" (a1)
+      : "memory", "a2", "a3", "a4", "a6", "a7",
```
hudson-ayers, line 354: same comment about a5 here

hudson-ayers (commented):
This mostly looks good, but I think a5 should be marked as a clobber still in the risc-v implementation.

libtock/tock.c
phil-levis: I've generally been trying to do minimum-change PRs, rather than go down the path of refactoring.  But sure, I can refactor here.

libtock/tock.c
phil-levis: Yes! Nice catch. I have no idea what went wrong here.

libtock/tock.c
phil-levis: Yes! Nice catch. I have no idea why that happened.

libtock/tock.c
phil-levis: Agreed.

hudson-ayers (approved):
You'll have to remove a0 and a1 from the yield clobber lists on risc-v for travis to pass, but otherwise looks good to me

phil-levis:
Yup, cleaning it up now.

PR #98, opened 2020-07-30
bump up versions for newlib, libgcc

JOE1994:
This PR makes the following changes
* bump up newlib: **2.5.0.20170421 => 3.0.0.20180831**
  Other versions of newlib newer than 3.0.0.20180831 causes the
  `lua_hello` example to fail, so I chose this version of newlib.
* bump up libgcc: **6.2.0 => 9.3.0**

I locally succeeded in rebuilding `newlib` &`libgcc` & examples using `gcc 7.5.0`.

Thank you for reviewing this PR :cake:

PR #165, opened 2021-01-25
Bump up versions for newlib, libgcc (gcc 10.2)

ppannuto:
This is a replacement for #98, where the only substantial differences are that it also picks up rebuilds of all the ancillary libraries as well and updates to the now-latest gcc as some time has passed.

**Note:** This still needs platform testing.


hudson-ayers:
This looks great. Does "platform testing" in this case mean it needs to be tested on every chip Tock supports?

alistair23:
Only the ARM ones, for RISC-V we use the toolchain libc

ppannuto:
This got lost as a draft PR -- on second thought, I think it makes more sense to fold this into the other newlib changes in #237, then take advantage of the 2.0 testing for all the updates here.

PR #166, opened 2021-01-27
A handlful of fixes and cleanups

alistair23:
This PR contains a few small cleanups to:
 - Add some missing includes and remove ones that aren't required
 - Rename `asm` to `__asm__` to be more flexible with `-ansi` and `-std` arguments (see https://gcc.gnu.org/onlinedocs/gcc-7.2.0/gcc/Extended-Asm.html)

### DIFF
--- a/libtock/tock.h
+++ b/libtock/tock.h
@@ -2,7 +2,7 @@
 
 #include <stdbool.h>
 #include <stdint.h>
-#include <unistd.h>
+#include <stddef.h>
 
 #ifdef __cplusplus
 extern "C" {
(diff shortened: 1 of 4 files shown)


libtock/tock.h
```diff
@@ -2,7 +2,7 @@
 
 #include <stdbool.h>
 #include <stdint.h>
-#include <unistd.h>
+#include <stddef.h>
```
ppannuto, line 5: I believe this change is throwing the build error:

```
WRITE(2)                    BSD System Calls Manual                   WRITE(2)

NAME
     pwrite, write, pwritev, writev -- write output

LIBRARY
     Standard C Library (libc, -lc)

SYNOPSIS
     #include <unistd.h>
```

ppannuto (changes requested):
Otherwise all looks fine

ppannuto (approved):
bors r+

PR #169, opened 2021-01-27
I2C master library and test app.

phil-levis:
This adds userspace support for I2C Master in libtock-c and adds a test application that works with the I2C Master/Slave test application.


ppannuto (approved):
bors r+

Took me a second to understand why we needed this on top of the existing master_slave stuff -- at some point an organization / documentation layer for libtock would probably be beneficial, but this isn't the time or place for that.

Looks good.

PR #170, opened 2021-01-27
libtock: Support for userspace read allow and read only state

alistair23:
This is a companion PR with the Tock PR here: https://github.com/tock/tock/pull/2381

This adds support for getting information from the Tock kernel using read only state.

This also includes adding support for a `quick_yield()` that doesn't require a syscall if there are no pending tasks.

### DIFF
--- a/libtock/read_only_state.h
+++ b/libtock/read_only_state.h
@@ -40,6 +40,16 @@ uint32_t read_only_state_get_pending_tasks(void* base);
 // to get the current time returned from the kernel.
 uint64_t read_only_state_get_ticks(void* base);
 
+// Use ROS to check if there are any pending tasks. If there are
+// we yield, if not then we return immediately without any syscalls.
+//
+// If there are no pending tasks returns 0.
+// If there are pending tasks returns the number of pending tasks before
+// yield is called. That is is returning 1 there will be no more tasks
+// pending (as long as no new tasks occurred)
+// On error returns a negative value.
+int read_only_state_quick_yield(void* base);
+
 #ifdef __cplusplus
 }
 #endif
--- a/libtock/read_only_state.h
+++ b/libtock/read_only_state.h
@@ -1,0 +1,45 @@
+#pragma once
+
+#include <stdlib.h>
+#include <string.h>
+
+#include "tock.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#define READ_ONLY_STATEDRIVER_NUM 0x00009
+
+// We currently support ROS version 1
+// Version 1:
+//   |-------------------------|
+//   |       Count (u32)       |
+//   |-------------------------|
+//   |   Pending Tasks (u32)   |
+//   |-------------------------|
+//   |                         |
+//   |     Time Ticks (u64)    |
+//   |-------------------------|
+#define READ_ONLY_STATEBUFFER_LEN (4 * 4 + 4 * 4 + 8 * 4)
+
+// Get the latest version of the read only state supported by the kernel.
+int read_only_state_get_version(void);
+
+// Share a buffer with the kernel to use for read only state
+//
+// `base` the buffer to use
+// `len` should be READ_ONLY_STATEBUFFER_LEN
+int read_only_state_allocate_region(void* base, int len);
+
+// Use the read only state buffer provided by `base`
+// to get the number of pending tasks.
+uint32_t read_only_state_get_pending_tasks(void* base);
+
+// Use the read only state buffer provided by `base`
+// to get the current time returned from the kernel.
+uint64_t read_only_state_get_ticks(void* base);
+
+#ifdef __cplusplus
+}
+#endif
--- a/libtock/tock.c
+++ b/libtock/tock.c
@@ -103,7 +103,7 @@ void yield_for(bool *cond) {
 }
 
 // Returns 1 if a task is processed, 0 otherwise
-static int __yield_check_tasks(void) {
+int yield_check_tasks(void) {
   if (task_cur != task_last) {
     tock_task_t task = task_queue[task_cur];
     task_cur = (task_cur + 1) % TASK_QUEUE_SIZE;
@@ -118,7 +118,7 @@ static int __yield_check_tasks(void) {
 
 
 void yield(void) {
-  if (__yield_check_tasks()) {
+  if (yield_check_tasks()) {
     return;
   } else {
     // Note: A process stops yielding when there is a callback ready to run,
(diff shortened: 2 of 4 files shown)


alistair23:
This is ready to merge

libtock/tock.c
```diff
@@ -102,6 +103,24 @@ void yield_for(bool *cond) {
...
+    uint32_t tasks = ros_get_pending_tasks(base);
+
+    if (tasks > 0) {
+      // Waiting tasks, call yield
+      yield();
+    }
+
+    return tasks;
+  }
+}
```
hudson-ayers, line 122: It seems like this should be implemented in `ros.c`, so this file does not have a dependency on `ros.h`, and so that it is clear that this operation is only possible on boards with the `ros` capsule installed.

libtock/ros.c
```diff
@@ -0,0 +1,50 @@
...
+}
+
+uint64_t ros_get_ticks(void* base) {
+  uint32_t* ptr = base;
+
+  // Start with the high bytes set to 0
+  uint32_t high, low;
+
+  do {
+    // Set the high bytes the value in memory
```
ppannuto, line 41:
```suggestion
    // Get the high bytes the value in memory
```

libtock/ros.h
```diff
@@ -0,0 +1,55 @@
...
+//   |-------------------------|
+//   |   Pending Tasks (u32)   |
+//   |-------------------------|
+//   |                         |
+//   |     Time Ticks (u64)    |
+//   |-------------------------|
+#define ROS_BUFFER_LEN (4 * 4 + 4 * 4 + 8 * 4)
+
+// Get the latest version of the read only state supported by the kernel.
+int ros_get_version(void);
```
ppannuto, line 27: Should this maybe be some form of `bool read_only_state_is_supported_version(void)` ?

My thinking here is that library users don't really care what the version is, what they care about is whether the library is going to work correctly. This seems like a more future-proof interface to expose to users?

libtock/tock.c
```diff
@@ -102,6 +103,24 @@ void yield_for(bool *cond) {
...
+
```
ppannuto, line 123: I think there's some weird intermingling between the `tock` and `read_only_state` module here.

Can this possibly be decoupled a bit, i.e. maybe a `tock_execute_next_task()` method which executes the top half `if (task_cur != task_last)` case; that can be part of the `tock` module generally, then the `read_only_state_quick_yield(base)` can call that method and do the second half?

libtock/tock.c
```diff
@@ -317,6 +336,36 @@ allow_rw_return_t allow_readwrite(uint32_t driver, uint32_t allow, void* ptr, si
...
+    );
+  if (rtype == TOCK_SYSCALL_SUCCESS_U32_U32) {
+    allow_userspace_r_return_t rv = {true, (void*)rv1, (size_t)rv2, 0};
+    return rv;
+  } else if (rtype == TOCK_SYSCALL_FAILURE_U32_U32) {
+    allow_userspace_r_return_t rv = {false, (void*)rv2, (size_t)rv3, (statuscode_t)rv1};
+    return rv;
+  } else {
+    // Invalid return type
+    exit(1);
```
ppannuto, line 365:
```suggestion
    exit(-1);
```

.. This file has a mix of `1` and `-1` for this case. I think all of these should be `-1`.., will file a separate PR for the existing ones.

libtock/tock.c
```diff
@@ -488,6 +537,35 @@ allow_rw_return_t allow_readwrite(uint32_t driver, uint32_t allow,
...
+    : "memory");
+  if (rtype == TOCK_SYSCALL_SUCCESS_U32_U32) {
+    allow_userspace_r_return_t rv = {true, (void*)rv1, (size_t)rv2, 0};
+    return rv;
+  } else if (rtype == TOCK_SYSCALL_FAILURE_U32_U32) {
+    allow_userspace_r_return_t rv = {false, (void*)rv2, (size_t)rv3, (statuscode_t)rv1};
+    return rv;
+  } else {
+    // Invalid return type
+    exit(1);
```
ppannuto, line 565:
```suggestion
    exit(-1);
```

ppannuto (changes requested):

libtock/ros.c
```diff
@@ -0,0 +1,50 @@
+#include "ros.h"
```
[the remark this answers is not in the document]
alistair23: I changed the file names, do you also think the function names should be changed? That will make them fairly long

libtock/ros.h
alistair23: I actually think it's the other way around. Library users do care about the version, for example they might want to do this

```c
if (ros_get_version() >= 2) {
    /* Version 2 supports this new function, so let's use that */
    ros_cool_function();
} else {
   /* Let's do it the old fashioned way */
   really_slow_function();
}
```

hudson-ayers (approved):

libtock/read_only_state.h
```diff
@@ -0,0 +1,55 @@
...
+
+// Use ROS to check if there are any pending tasks. If there are
+// we yield, if not then we return immediately without any syscalls.
+//
+// If there are no pending tasks returns 0.
+// If there are pending tasks returns the number of pending tasks before
+// yield is called. That is is returning 1 there will be no more tasks
+// pending (as long as no new tasks occurred)
+// On error returns a negative value.
+int quick_yield(void* base);
```
ppannuto, line 51:
```suggestion
int ros_quick_yield(void* base);
```

```suggestion
int quick_yield_using_ros(void* base);
```

I think this should maybe include `ros` in the function name somewhere to give the context hint at usage points that it relies on `ros` being set up (and that's what the arguments is about).

libtock/ros.h
ppannuto: Yeah, that's true as well. My concern is that people can write code that doesn't check the version and then things break out from underneath them, but I suppose that's really a user-error issue, and we can't save them from everything.

libtock/ros.c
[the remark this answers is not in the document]
ppannuto: I think so? I know C is the land of `creat` and other gems, but I feel like self-documenting code is valuable. This does push towards long lines, `read_only_state_get_version()` is 28 characters, but that's still plenty of room for some `if [] > 2 {` and indentation. I'm willing to be pushed back on here, but I feel like our default is to be verbose/explicit, and we've been better off for it.

ppannuto (approved):
bors d+

I pushed a commit that also expanded out the `ROS` in the constants. I think this is good to go but didn't want to just merge with an extra commit of mine on top in case you had any issues.

alistair23:
Ahh... Any idea of why it's failing?

alistair23:
Maybe it's a download issue. Try again

bors r+

PR #173, opened 2021-02-10
Update BLE advertising and scanning for the Tock 2.0 interface

alevy:
This pull request makes minimal changes to update the libtock Bluetooth Low Energy driver to the Tock 2.0 interface as implemented in tock/tock#2411

I successfully tested both advertising and scanning on an Arduino Nano BLE 33 sense with the `ble_advertising` and `ble_scanning` apps.

Piggybacking on this is a small change to the Nix shell expression to use tockloader 1.6.0, which supports programming the Arduino Nano.

### DIFF
--- a/libtock/ble.c
+++ b/libtock/ble.c
@@ -10,15 +10,25 @@
 #include <string.h>
 
 int ble_start_advertising(int pdu_type, uint8_t* advd, int len, uint16_t interval) {
-  int err = allow(BLE_DRIVER_NUMBER, 0, advd, len);
-  if (err < TOCK_SUCCESS)
-    return err;
+  allow_ro_return_t err = allow_readonly(BLE_DRIVER_NUMBER, BLE_CFG_ADV_BUF_ALLOWRO, advd, len);
+  if (!err.success)
+    return tock_error_to_rcode(err.error);
 
-  return command(BLE_DRIVER_NUMBER, BLE_ADV_START_CMD, pdu_type, interval);
+  syscall_return_t res = command2(BLE_DRIVER_NUMBER, BLE_ADV_START_CMD, pdu_type, interval);
+  if (res.type == TOCK_SYSCALL_SUCCESS) {
+    return TOCK_SUCCESS;
+  } else {
+    return tock_error_to_rcode(res.data[0]);
+  }
 }
 
 int ble_stop_advertising(void) {
-  return command(BLE_DRIVER_NUMBER, BLE_ADV_STOP_CMD, 1, 0);
+  syscall_return_t res = command2(BLE_DRIVER_NUMBER, BLE_ADV_STOP_CMD, 1, 0);
+  if (res.type == TOCK_SYSCALL_SUCCESS) {
+    return TOCK_SUCCESS;
+  } else {
+    return tock_error_to_rcode(res.data[0]);
+  }
 }
 
 int ble_start_passive_scan(uint8_t *data, uint8_t max_len,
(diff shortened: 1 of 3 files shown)


libtock/ble.c
```diff
@@ -26,16 +26,15 @@ int ble_start_passive_scan(uint8_t *data, uint8_t max_len,
...
-    err =
-      allow(BLE_DRIVER_NUMBER, BLE_CFG_SCAN_BUF_ALLOW, (void *)data, max_len);
-    if (err < TOCK_SUCCESS)
-      return err;
+    allow_rw_return_t allow_err =
+      allow_readwrite(BLE_DRIVER_NUMBER, BLE_CFG_SCAN_BUF_ALLOWRW, (void *)data, max_len);
+    if (!allow_err.success)
+      return allow_err.error;
 
     return command(BLE_DRIVER_NUMBER, BLE_SCAN_CMD, 1, 0);
```
hudson-ayers, line 39:
```suggestion
    syscall_return_t res = command2(BLE_DRIVER_NUMBER, BLE_SCAN_CMD, 1, 0);
      if (res.type == TOCK_SYSCALL_SUCCESS) {
        return TOCK_SUCCESS;
      } else {
        return tock_error_to_rcode(res.data[0]);
      }
```

basically this should be done for all calls to command (though those that return values are of course a little different, see ADC for an example)

hudson-ayers (changes requested):
This needs to use subscribe2 and command2 instead of the tock 1.0 versions of those system calls

libtock/ble.c
alevy: Done

libtock/ble.c
```diff
@@ -26,16 +26,15 @@ int ble_start_passive_scan(uint8_t *data, uint8_t max_len,
...
     return TOCK_FAIL;
   } else {
-    int err;
 
-    err = subscribe(BLE_DRIVER_NUMBER, BLE_SCAN_SUB, callback, NULL);
-    if (err < TOCK_SUCCESS)
-      return err;
+    int sub_err = subscribe(BLE_DRIVER_NUMBER, BLE_SCAN_SUB, callback, NULL);
+    if (sub_err < TOCK_SUCCESS)
+      return sub_err;
```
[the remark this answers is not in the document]
alevy: Done

libtock/ble.c
```diff
@@ -10,41 +10,65 @@
...
-  if (err < TOCK_SUCCESS)
-    return err;
+  allow_ro_return_t err = allow_readonly(BLE_DRIVER_NUMBER, BLE_CFG_ADV_BUF_ALLOWRO, advd, len);
+  if (!err.success)
+    return tock_error_to_rcode(err.error);
 
-  return command(BLE_DRIVER_NUMBER, BLE_ADV_START_CMD, pdu_type, interval);
+  syscall_return_t res = command2(BLE_DRIVER_NUMBER, BLE_ADV_START_CMD, pdu_type, interval);
+  if (res.type == TOCK_SYSCALL_SUCCESS_U32) {
+    return res.data[0];
```
hudson-ayers, line 19:
```suggestion
  if (res.type == TOCK_SYSCALL_SUCCESS) {
    return TOCK_SUCCESS;
```

`TOCK_SYSCALL_SUCCESS_U32` will only be present when the capsule returns `CommandValue::success_u32()`, and looking at the updated capsule it does not seem to do that anywhere

hudson-ayers: This applies to all the command calls

alevy: Of course! Done

hudson-ayers (approved):

PR #174, opened 2021-02-10
Updating NRF serialization to 2.0 syscall API.

phil-levis:
This updates the userspace side of the nrf serialization library. It is not working -- the Nordic library fails on `simple_ble_init`.


bradjc:
There should be changes to libnrfserialization to match the changed API. I'm not sure how this would compile.

phil-levis:
> 
> 
> There should be changes to libnrfserialization to match the changed API. I'm not sure how this would compile.

Yup, pushed, it's working now.

libtock/nrf51_serialization.h
```diff
@@ -19,12 +19,21 @@ int nrf51_serialization_subscribe (subscribe_cb cb);
 
 // Pass a buffer for the driver to write received UART bytes to.
 __attribute__ ((warn_unused_result))
-int nrf51_serialization_setup_rx_buffer (char* rx, int rx_len);
+int nrf51_serialization_setup_receive_buffer (char* rx, int rx_len);
+
+// Pass a buffer for the driver to write received UART bytes to.
+__attribute__ ((warn_unused_result))
+int nrf51_serialization_setup_write_buffer (char* rx, int rx_len);
```
hudson-ayers, line 26: This no longer seems to be implemented in the c file

hudson-ayers (commented):
The code changes look good, though I have one comment. Also, it seems slightly surprising that this adds a new tar.gz file

libtock/nrf51_serialization.h
phil-levis: Fixed.

phil-levis:
> The code changes look good, though I have one comment. Also, it seems slightly surprising that this adds a new tar.gz file

Agreed -- this should be autofetched when you build libnrfserialization. I think it snuck in with the archive updates.

hudson-ayers:
It looks like I spoke too soon, travis fails without that file pushed to git. Sorry about that.

phil-levis:
Yeah -- trying to figure why it showed up as an addition: maybe it was just a replace? The file has been in there for the past 3 years.

hudson-ayers:
It seems that for binary files if you replace github just shows +100% rather than the typical +85 / -30 that you get for text files. Pretty confusing interface

hudson-ayers (approved):

ppannuto (approved):

PR #175, opened 2021-02-12
Update nrf-serialization

bradjc:
This works with my one patch to the kernel PR. I tested on hail.

### DIFF
--- a/examples/ble-uart/Makefile
+++ b/examples/ble-uart/Makefile
@@ -7,8 +7,6 @@ TOCK_USERLAND_BASE_DIR = ../..
 C_SRCS := $(wildcard *.c)
 
 APP_HEAP_SIZE=8096
-
-#CXX_FLAGS += -DAPP_HEAP_SIZE=8096
 # External libraries used
 EXTERN_LIBS += $(TOCK_USERLAND_BASE_DIR)/libnrfserialization
 
--- a/examples/ble-uart/Makefile
+++ b/examples/ble-uart/Makefile
@@ -6,6 +6,9 @@ TOCK_USERLAND_BASE_DIR = ../..
 # Which files to compile.
 C_SRCS := $(wildcard *.c)
 
+APP_HEAP_SIZE=8096
+
+#CXX_FLAGS += -DAPP_HEAP_SIZE=8096
 # External libraries used
 EXTERN_LIBS += $(TOCK_USERLAND_BASE_DIR)/libnrfserialization
 
--- a/libtock/nrf51_serialization.c
+++ b/libtock/nrf51_serialization.c
@@ -13,14 +13,24 @@ int nrf51_serialization_reset (void) {
 int nrf51_serialization_subscribe (subscribe_cb cb) {
   // get some callback love
   subscribe_return_t sval = subscribe2(DRIVER_NUM_NRF_SERIALIZATION, 0, cb, NULL);
-  if (sval.success) {
-    return TOCK_SUCCESS;
-  } else {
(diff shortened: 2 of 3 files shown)


libtock/nrf51_serialization.c
```diff
@@ -2,27 +2,56 @@
...
   // get some callback love
-  return subscribe(DRIVER_NUM_NRF_SERIALIZATION, 0, cb, NULL);
+  subscribe_return_t sval = subscribe2(DRIVER_NUM_NRF_SERIALIZATION, 0, cb, NULL);
+  if (!sval.success) {
+    return tock_error_to_rcode(sval.error);
+  }
+
+  // Instruct the capsule to start receiving. Doesn't
+  syscall_return_t cval = command2(DRIVER_NUM_NRF_SERIALIZATION, 2, 0, 0);
+  if (cval.type == TOCK_SYSCALL_SUCCESS_U32) {
```
hudson-ayers, line 22: sort of odd that this command returns a `success_u32` if userspace doesn't even need the value

libtock/nrf51_serialization.c
```diff
@@ -2,27 +2,56 @@
...
+  }
 
   // Do the write!!!!!
-  ret = command(DRIVER_NUM_NRF_SERIALIZATION, 1, 0, 0);
-  return ret;
+  syscall_return_t sval = command2(DRIVER_NUM_NRF_SERIALIZATION, 1, 0, 0);
+  if (sval.type == TOCK_SYSCALL_SUCCESS) {
+    return TOCK_SUCCESS;
+  } else {
+    return tock_error_to_rcode(sval.data[0]);
```
hudson-ayers, line 55:
```suggestion
  syscall_return_t cval = command2(DRIVER_NUM_NRF_SERIALIZATION, 1, 0, 0);
  if (cval.type == TOCK_SYSCALL_SUCCESS) {
    return TOCK_SUCCESS;
  } else {
    return tock_error_to_rcode(cval.data[0]);
```

hudson-ayers (approved):
This seems to me like a reasonable compromise between effort required to port to 2.0 and still making sure the capsule is written in a way that is compliant with tock 2.0 standards.

I agree with brad that it does not seem worth updating the interface between the `nrf_serialization` binary blob and this file just to make the function names slightly better map to the underlying system calls they issue, given how challenging this code can be to debug.

examples/ble-uart/Makefile
```diff
@@ -6,6 +6,9 @@ TOCK_USERLAND_BASE_DIR = ../..
 # Which files to compile.
 C_SRCS := $(wildcard *.c)
 
+APP_HEAP_SIZE=8096
+
+#CXX_FLAGS += -DAPP_HEAP_SIZE=8096
```
ppannuto, line 11:
```suggestion
```

libtock/nrf51_serialization.c
phil-levis: The userspace library API is pretty coupled to the semantics of the nRF stack, which is typical embedded code (makes lots of assumptions).

phil-levis (commented):
#174 is now working.

libtock/nrf51_serialization.c
phil-levis: It is a bit weird that a call to subscribe also makes a call to read; this isn't the behavior that I would expect. I'd rather not have this weird wart of non-standard behavior. #174 fixes `serialization_tock.c` to correctly invoke a read operation.

phil-levis:
#174 subsumes this.

PR #176, opened 2021-02-17
Port IPC to 2.0 driver

alevy:
The only change in the IPC interface is the return type of `ipc_share`,
which is now an `allow_return_t` instead of an int. The implementation
of `ipc_discover` now performs a command in addition to the
allow_readonly to perform the discover and an additional allow to get
ownership of the buffer back.

### DIFF
--- a/libtock/ipc.c
+++ b/libtock/ipc.c
@@ -1,12 +1,16 @@
 #include "ipc.h"
+#include "tock.h"
 
 int ipc_discover(const char* pkg_name) {
   int len = strlen(pkg_name);
-  char* pkg_name_buf = (char*)malloc(len * sizeof(char));
-  memcpy(pkg_name_buf, pkg_name, len);
-  int res = allow(IPC_DRIVER_NUM, 0, pkg_name_buf, len);
-  free(pkg_name_buf);
-  return res;
+  allow_ro_return_t prev = allow_readonly(IPC_DRIVER_NUM, 0, pkg_name, len);
+  syscall_return_t res   = command2(IPC_DRIVER_NUM, 1, 0, 0);
+  prev = allow_readonly(IPC_DRIVER_NUM, 0, prev.ptr, prev.size);
+  if (res.type != TOCK_SYSCALL_SUCCESS_U32) {
+    return tock_error_to_rcode(res.data[0]);
+  } else {
+    return res.data[0];
+  }
 }
 
 int ipc_register_svc(subscribe_cb callback, void *ud) {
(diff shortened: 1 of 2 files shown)


libtock/ipc.c
```diff
@@ -1,12 +1,16 @@
...
 int ipc_discover(const char* pkg_name) {
   int len = strlen(pkg_name);
-  char* pkg_name_buf = (char*)malloc(len * sizeof(char));
-  memcpy(pkg_name_buf, pkg_name, len);
-  int res = allow(IPC_DRIVER_NUM, 0, pkg_name_buf, len);
-  free(pkg_name_buf);
-  return res;
+  allow_ro_return_t prev = allow_readonly(IPC_DRIVER_NUM, 0, pkg_name, len);
+  syscall_return_t res   = command2(IPC_DRIVER_NUM, 1, 0, 0);
+  prev = allow_readonly(IPC_DRIVER_NUM, 0, prev.ptr, prev.size);
```
hudson-ayers, line 8: I don't see any reason to "unallow" here -- the process retains read access to the buffer after a readonly share, and the kernel cannot modify it anyway

hudson-ayers (approved):

libtock/ipc.c
alevy: That's true. It seems safest since the process might change it-tt's true that the kernel doesn't _in practice_ rely on the package name after discovery has happened, but this seems least ambiguous and also restores whatever was there before. It does add another system call, of course.

You can imagine fairly convoluted uses of this interfaces that would still work.

Suppose, for example, you have multiple coroutines, one of them sets the package name with allow then yields (expecting to perform discovery some later time). Then a coroutine runs that uses this library function for discover. At the end of this call, because we've re-allowed with whatever was there before, the state hasn't changed for the first co-routine, and it is fine for it to now perform discovery with a simple command.

But, it was also on a whim. If we're not doing something like this as a pattern, probably makes sense to avoid here too.

hudson-ayers: That's a good point, I hadn't considered that. I haven't really seen this as a pattern in other places, but am happy either way

libtock/ipc.c
```diff
@@ -21,17 +25,19 @@ int ipc_register_client_cb(int svc_id, subscribe_cb callback, void *ud) {
...
+  if (res.type != TOCK_SYSCALL_SUCCESS) {
+    return tock_error_to_rcode(res.data[0]);
+  } else {
+    return TOCK_SUCCESS;
+  }
 }
 
 int ipc_notify_client(int pid) {
-  return command(IPC_DRIVER_NUM, pid, 1, 0);
+  return command(IPC_DRIVER_NUM, 3, pid, 0);
```
phil-levis, line 37: Should this be a command2?

PR #177, opened 2021-02-20
Tock 2.0 switchover

phil-levis:
This PR 
1. removes the Tock 1.0 system calls from tock.c,
2. changes `subscribe_cb` to `subscribe_upcall`
3. updates `driver_exists` to use `command2`

It **does not** yet change the names (e.g., `command2` to `command`). This because many userspace libraries have not been ported to the new system calls:

- [x] tsl2561 @ppannuto #182
- [x] temperature @phil-levis #178 
- [x] ~aes~ 
- [x] ~st7735~ #185 
- [x] pca9544a @alevy 
- [x] lsm303dlhc #191 @alexandruradovici 
- [x] usb @alistair23 #189 
- [x] ~console~ (minor, fixed in this PR) @phil-levis #177 
- [x] proximity @alevy  #181
- [x] ltc294x @alevy #180 
- [x] ~tmp006~ #184 
- [x] l3gd20 #192 @alexandruradovici 
- [x] lps35hb #187 
- [x] humidity @phil-levis #178 
- [x] max17205 #183 
- [x] ipc @alevy #176, #177

This causes an undefined symbol when compiling, so libtock-c is not usable/linkable. But it does compile. As these are fixed, tested, and pushed to `tock-2.0-dev`, I will keep this branch up to date and unblock it when it is ready.

PR #178, opened 2021-02-20
Update temperature and humidity to use 2.0 API

phil-levis:
In porting over the capsules, somehow the libtock-c parts were missed for temperature and humidity. This updates them to the 2.0 syscall API. Part of #177

### DIFF
--- a/libtock/humidity.c
+++ b/libtock/humidity.c
@@ -9,16 +9,16 @@ struct data {
 static struct data result = { .fired = false };
 
 // Internal callback for faking synchronous reads
-static void upcall(int humidity,
-                   __attribute__ ((unused)) int unused,
-                   __attribute__ ((unused)) int unused1,
-                   void* ud) {
+static void humidity_upcall(int humidity,
+                            __attribute__ ((unused)) int unused,
+                            __attribute__ ((unused)) int unused1,
+                            void* ud) {
   struct data* data = (struct data*) ud;
   data->humidity = humidity;
   data->fired    = true;
 }
 
-int humidity_set_upcall(subscribe_cb callback, void* callback_args) {
+int humidity_set_callback(subscribe_cb callback, void* callback_args) {
   subscribe_return_t sval = subscribe2(DRIVER_NUM_HUMIDITY, 0, callback, callback_args);
   if (sval.success) {
     return TOCK_SUCCESS;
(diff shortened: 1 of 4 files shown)


hudson-ayers (approved):

libtock/humidity.c
```diff
@@ -9,28 +9,40 @@ struct data {
 static struct data result = { .fired = false };
 
 // Internal callback for faking synchronous reads
-static void cb(int humidity,
-               __attribute__ ((unused)) int unused,
-               __attribute__ ((unused)) int unused1,
-               void* ud) {
+static void upcall(int humidity,
```
ppannuto, line 12: Should the comment above this also change to `upcall`?

ppannuto (dismissed):
The logic looks fine, but this seems to inconsistently change things from `callback` to `upcall` -- I don't really have strong feelings about which word we use, but currently "callback" is the term used throughout most of libtock-c, if we're going to change names, I think we should do it consistently rather than have a mix of terms (inviting the question: what is the difference between a callback and an upcall?)

libtock/humidity.c
phil-levis: The plan is to change everything to upcall, to distinguish it from general callbacks within the kernel. Otherwise it starts to be confusing, whether you're talking about a `Callback` or a callback.

A callback is just a function that's called back on the caller; an upcall is when the kernel invokes a function on a userspace process.

bradjc:
I think it makes sense to call what happens when a capsule calls `.schedule()` an upcall. That differentiates it from a `.set_client()`-based callback.

I don't think that means we should automatically rename libtock-c, however. The API and terminology presented to userspace  applications using libtock-c does not need to be based on the terminology of the kernel. Callbacks in a main.c file look like any other callback in any other language, and renaming them to upcalls is I think unnecessarily adding complexity. Whether they are the direct result of a kernel upcall or come from more software layers in libtock-c (alarms are a good example of this) is not really a concern of an app developer, to them it just looks like a callback.

Changing `subscribe()` to take an upcall function pointer makes sense, but I don't think we should change the libtock-c APIs to upcall at the same time.

phil-levis:
I think it's fine if the libtock-c function names say "callback" rather than "upcall". But what about the typedef? The clean way to do this is for a libtock-c library to filter the upcall into a callback. While the upcall provides 4 arguments, for example, it might be the callback only has 2. So the type you pass to a subscribe is a `subscribe_upcall` but the type you pass to humidity_set_callback is a function pointer with a single argument, the humidity value.

PR #188, opened 2021-02-22
no aes driver in kernel any more

bradjc:


hudson-ayers (approved):
Looks good, but it seems believable someone will want to bring much of this back one day, should we put a reference to this PR somewhere discoverable?

bradjc:
I suppose? Where would that go?

hudson-ayers:
Created: https://github.com/tock/tock/issues/2453

PR #189, opened 2021-02-22
2.0 usb

bradjc:


libtock/usb.c
```diff
@@ -1,15 +1,36 @@
...
-  return command(DRIVER_NUM_USB, 0, 0, 0) >= 0;
+  syscall_return_t com = command2(DRIVER_NUM_USB, 0, 0, 0);
+  if (com.type == TOCK_SYSCALL_SUCCESS) {
+    return TOCK_SUCCESS;
+  } else if (com.type > TOCK_SYSCALL_SUCCESS) {
+    // Returned an incorrect success code
+    return TOCK_FAIL;
+  } else {
+    return tock_error_to_rcode(com.data[0]);
+  }
```
hudson-ayers, line 12: before this returned 1 if the driver was present and 0 otherwise, now it returns 0 if the driver is present

hudson-ayers (changes requested):

libtock/usb.c
bradjc: Fixed.

hudson-ayers (approved):

PR #191, opened 2021-02-24
Port Lsm303dlhc to Tock 2.0

hudson-ayers:
This PR ports the lsm303dlhc to tock 2.0. all command calls return success() and were grouped into helper functions already, so this was quite straightforward.

### DIFF
--- a/libtock/lsm303dlhc.c
+++ b/libtock/lsm303dlhc.c
@@ -27,6 +27,10 @@ static int lsm303dlhc_subscribe (subscribe_cb cb, void *userdata) {
   return TOCK_SUCCESS;
 }
 
+// Helper function for command system calls that do not return an extra value
+// in the case of success. If a command is added to the driver that returns
+// success_u32, for example, this function can not be used for that system
+// call.
 static int lsm303dlhc_command_noval (uint32_t command_num, uint32_t data1, uint32_t data2) {
   syscall_return_t com = command2(DRIVER_NUM_LSM303DLHC, command_num, data1, data2);
   if (com.type == TOCK_SYSCALL_SUCCESS) {
--- a/libtock/lsm303dlhc.c
+++ b/libtock/lsm303dlhc.c
@@ -20,11 +20,22 @@ static void command_callback_yield (int data1, int data2, int data3, void* ud) {
 
 
 static int lsm303dlhc_subscribe (subscribe_cb cb, void *userdata) {
-  return subscribe(DRIVER_NUM_LSM303DLHC, 0, cb, userdata);
+  subscribe_return_t subval = subscribe2(DRIVER_NUM_LSM303DLHC, 0, cb, userdata);
+  if (subval.success == 0) {
+    return tock_error_to_rcode(subval.error);
+  }
+  return TOCK_SUCCESS;
 }
 
-static int lsm303dlhc_command (uint32_t command_num, uint32_t data1, uint32_t data2) {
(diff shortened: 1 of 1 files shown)


libtock/lsm303dlhc.c
```diff
@@ -20,11 +20,22 @@ static void command_callback_yield (int data1, int data2, int data3, void* ud) {
...
-static int lsm303dlhc_command (uint32_t command_num, uint32_t data1, uint32_t data2) {
-  return command(DRIVER_NUM_LSM303DLHC, command_num, data1, data2);
+static int lsm303dlhc_command_noval (uint32_t command_num, uint32_t data1, uint32_t data2) {
+  syscall_return_t com = command2(DRIVER_NUM_LSM303DLHC, command_num, data1, data2);
+  if (com.type == TOCK_SYSCALL_SUCCESS) {
+    return TOCK_SUCCESS;
+  } else if (com.type == TOCK_SYSCALL_FAILURE) {
+    return tock_error_to_rcode(com.data[0]);
+  } else {
+    return TOCK_EBADRVAL;
```
bradjc, line 37: Uh, does this mean I messed up all of my PRs? I copied from console, I thought that would be a good example to go by, and return `TOCK_FAIL`.

libtock/lsm303dlhc.c
```diff
@@ -20,11 +20,22 @@ static void command_callback_yield (int data1, int data2, int data3, void* ud) {
...
+  subscribe_return_t subval = subscribe2(DRIVER_NUM_LSM303DLHC, 0, cb, userdata);
+  if (subval.success == 0) {
+    return tock_error_to_rcode(subval.error);
+  }
+  return TOCK_SUCCESS;
 }
 
-static int lsm303dlhc_command (uint32_t command_num, uint32_t data1, uint32_t data2) {
-  return command(DRIVER_NUM_LSM303DLHC, command_num, data1, data2);
+static int lsm303dlhc_command_noval (uint32_t command_num, uint32_t data1, uint32_t data2) {
```
bradjc, line 30: What does "noval" mean? Shoudn't there be a "val" then?

hudson-ayers: the lsm303dlhc driver never returns a `success_u32()`, so no need. noval in this case means "the kernel return a `CommandResult:success()` as opposed to any other success variant which contains a value"

hudson-ayers: open to naming suggestions though

libtock/lsm303dlhc.c
hudson-ayers: Unfortunately I think `TOCK_EBADRVAL` was added partway through the porting process, so the vast majority of transitioned drivers do this the wrong way. Some even directly call `exit()`, I noticed. This would be good to address, but probably does not need to be blocking

libtock/lsm303dlhc.c
bradjc: A comment would be good. I found the specificity somewhat confusing, since there is no other command function in the file. As a reader I wonder, does "noval" matter? Why not specify other properties of the function in the name?

hudson-ayers: done

libtock/lsm303dlhc.c
phil-levis: Yeah, let's fix this between alpha and RC1.

phil-levis (approved):
Looks good.

PR #194, opened 2021-03-12
libtock: Use __asm__ instead of asm

alistair23:
The GCC documentation states:
"When writing code that can be compiled with -ansi and the various
-std options, use __asm__ instead of asm"
Switch to use __asm__ to be more flexible.


ppannuto (approved):
bors r+

PR #195, opened 2021-03-20
Tock 2.0 Alpha1 

phil-levis:
This updates tock-2.0-dev to master and irons out a few remaining wrinkles I found, particularly with the imix test app. It also adds an additional memory test, which walks to the end of the heap to find the end of the readable region. Finally, it updates the MPU test that checks the end of RAM to note that it should fail, since there are grants there.

This PR was tested by running the imix test application, `mpu_flash_end`, `mpu_heap_end`, and `mpu_ram_end`.

Note that in this state libtock-c does not uniformly follow one aspect of the Tock 2.0 calling conventions, particularly returning EBADVAL if the return type of a system call does not match what is expected. Some modules do, others don't. The plan is to clean this up before when we do a release.


bradjc (approved):

hudson-ayers (approved):
I did not review all 171 changed files, but I did look through the few commits at the beginning before all changes were submitted via PRs that were reviewed separately. LGTM

ppannuto (approved):
Like Hudson, I think I've reviewed most of these along the way

PR #199, opened 2021-04-09
Add bare bones example app

bradjc:
This example shows how to make the smallest app:

- It requests no memory.
- It does not use crt0.c.
- It just immediately calls yield.

### DIFF
--- a/examples/tests/bare_bones/main.c
+++ b/examples/tests/bare_bones/main.c
@@ -18,7 +18,8 @@ void _start(void* app_start __attribute__((unused)),
 
 #elif defined(__riscv)
   __asm__ volatile (
-      "li    a0, 0\n"
+      "li    a0, 1\n" // yield-wait
+      "li    a4, 0\n" // yield system call class
       "ecall\n"
       :
       :
--- a/examples/tests/bare_bones/main.c
+++ b/examples/tests/bare_bones/main.c
@@ -8,10 +8,22 @@ void _start(void* app_start __attribute__((unused)),
             void* memory_len __attribute__((unused)),
             void* app_heap_break __attribute__((unused))) {
   // Yield.
+#if defined(__thumb__)
   __asm__ volatile (
       "svc 0       \n"
       :
       :
       : "memory", "r0", "r1", "r2", "r3", "r12", "lr"
       );
+
+#elif defined(__riscv)
+  __asm__ volatile (
+      "li    a0, 0\n"
+      "ecall\n"
+      :
+      :
+      : "memory", "a0", "a1", "a2", "a3", "a4", "a5", "a6", "a7",
+      "t0", "t1", "t2", "t3", "t4", "t5", "t6", "ra"
+      );
+#endif
 }
--- a/examples/tests/bare_bones/main.c
+++ b/examples/tests/bare_bones/main.c
@@ -11,22 +11,22 @@ void _start(void* app_start __attribute__((unused)),
   // Yield.
 #if defined(__thumb__)
   __asm__ volatile (
-      "movs r0, #1 \n"
-      "svc 0       \n"
-      :
-      :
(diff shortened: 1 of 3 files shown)


examples/tests/bare_bones/main.c
```diff
@@ -0,0 +1,17 @@
...
+            void* mem_start __attribute__((unused)),
+            void* memory_len __attribute__((unused)),
+            void* app_heap_break __attribute__((unused))) {
+  // Yield.
+  __asm__ volatile (
+      "svc 0       \n"
+      :
+      :
+      : "memory", "r0", "r1", "r2", "r3", "r12", "lr"
+      );
```
alistair23, line 16: This will fail to build for RISC-V

alistair23 (changes requested):
Do you have an idea of what this would be useful for?

lschuermann:
@alistair23 I think this is pretty awesome to have a template and starting point for developing an app which does not use `libtock-c`'s `crt0` bootstrapping, but shows you the way the kernel will perform the initial call into the app. I can imagine this can be rather useful if one were to port runtimes / interpreters / JITs to Tock, such as MicroPython (which might not want to use the regular `libtock-c` app skeleton).

bradjc:
Good starting point if you want to skip crt0.c, and might be useful to test any assumptions the kernel is making about processes.

examples/tests/bare_bones/main.c
```diff
@@ -0,0 +1,29 @@
...
+
+#elif defined(__riscv)
+  __asm__ volatile (
+      "li    a0, 0\n"
```
lschuermann, line 21: This doesn't look right. AFAIK the system call type is passed in `a4` on RISC-V, with `a0` to `a3` being the arguments (where for yield we're probably going to want to load a well-defined value into `a0` anyways).
```suggestion
      "li    a0, 1\n" // yield-wait
      "li    a4, 0\n" // yield system call class
```

lschuermann (changes requested):

examples/tests/bare_bones/main.c
```diff
@@ -0,0 +1,29 @@
...
+void _start(void* app_start, void* mem_start, void* memory_len, void* app_heap_break);
+
+void _start(void* app_start __attribute__((unused)),
+            void* mem_start __attribute__((unused)),
+            void* memory_len __attribute__((unused)),
+            void* app_heap_break __attribute__((unused))) {
+  // Yield.
+#if defined(__thumb__)
+  __asm__ volatile (
+      "svc 0       \n"
```
lschuermann, line 13: We might want to deterministically execute a yield-wait call here as well. (https://github.com/tock/tock/blob/a66504cd9bc339d85e857bec0afd9797e8d81297/kernel/src/syscall.rs#L42-L48)

I don't really know ARM ASM so don't want to make a suggestion. :)

ppannuto: Adding a `movs r1, #1` before the `svc` should be portable across all cortex-m

examples/tests/bare_bones/main.c
bradjc: Ohhhh, ok I'm not crazy. a0 was the correct register when I wrote this app.

examples/tests/bare_bones/main.c
bradjc, line 13:
```suggestion
      "movs r0, #1 \n"
      "svc 0       \n"
```

alistair23 (approved):

lschuermann (approved):

ppannuto (approved):

bradjc:
Can we merge this?

hudson-ayers:
I had seen the CI didn't pass and ignored it as a result

bors r+

PR #200, opened 2021-04-09
Add crash immediate with syscall test app

bradjc:
This is a very small test app for testing what happens if an app crashes on its first instruction.

### DIFF
--- a/examples/tests/crash_immediately_syscall/main.c
+++ b/examples/tests/crash_immediately_syscall/main.c
@@ -1,0 +1,17 @@
+// Do nothing. yield() immediately.
+
+// Define the `_start_` signature.
+void _start(void* app_start, void* mem_start, void* memory_len, void* app_heap_break);
+
+void _start(void* app_start __attribute__((unused)),
+            void* mem_start __attribute__((unused)),
+            void* memory_len __attribute__((unused)),
+            void* app_heap_break __attribute__((unused))) {
+  // invalid syscall
+  __asm__ volatile (
+      "svc 0xbd       \n"
+      :
+      :
+      : "memory", "r0", "r1", "r2", "r3", "r12", "lr"
+      );
+}
--- a/examples/tests/crash_immediately_syscall/main.c
+++ b/examples/tests/crash_immediately_syscall/main.c
@@ -1,4 +1,4 @@
-// Do nothing. yield() immediately.
+// Crash immediately by executing a system call with an invalid system call class id
 
 // Define the `_start_` signature.
 void _start(void* app_start, void* mem_start, void* memory_len, void* app_heap_break);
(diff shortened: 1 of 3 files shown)


examples/tests/crash_immediately_syscall/main.c
```diff
@@ -0,0 +1,17 @@
...
+            void* mem_start __attribute__((unused)),
+            void* memory_len __attribute__((unused)),
+            void* app_heap_break __attribute__((unused))) {
+  // invalid syscall
+  __asm__ volatile (
+      "svc 0xbd       \n"
+      :
+      :
+      : "memory", "r0", "r1", "r2", "r3", "r12", "lr"
+      );
```
alistair23, line 16: This will fail to build fro RISC-V

alistair23 (changes requested):

hudson-ayers (commented):
Looks good, but agree with alistair that adding risc-v support is probably important

ppannuto:
Pushed a commit to mark this as a naked function (otherwise gcc will emit a prologue instruction that we don't really want for this). Also pushed what I think is the right thing for riscv, but something is broken on this machine with that toolchain that is beyond tonight's patience to debug.

hudson-ayers:
amended your commit with a format run so we can just see if travis can build the app for risc-v

ppannuto:
Oops. Thanks. The riscv toolchain is also fixed now on my machine (just needed to rebuild for reasons unknown, which became an overnight chore for ye olde laptop), and I can confirm it builds there happily as well.

hudson-ayers (approved):

alistair23 (approved):

examples/tests/crash_immediately_syscall/main.c
```diff
@@ -0,0 +1,29 @@
...
+#if defined(__thumb__)
+  __asm__ volatile (
+    "svc 0xbd       \n"
+    :
+    :
+    : "memory", "r0", "r1", "r2", "r3", "r12", "lr"
+    );
+#elif defined(__riscv)
+  __asm__ volatile (
+    "li    a4, 0xbd\n"
```
bradjc, line 21:
```suggestion
    "li    a0, 0xbd\n"
```
?

lschuermann: `a4` is the syscall argument register on RISC-V now: https://github.com/tock/tock/blob/master/doc/reference/trd104-syscalls.md#31-registers

examples/tests/crash_immediately_syscall/main.c
```diff
@@ -0,0 +1,29 @@
+// Do nothing. yield() immediately.
```
lschuermann, line 1: I think this comment is slightly misleading. :)

PR #203, opened 2021-04-19
remove volatile vars from crt0

### LINKED ISSUE #90: Remove unused `volatile` in crt0
I don't really want to trample on #88, nor do I want to add a logic change to arm stuff to that PR, but once it's through, we should fix https://github.com/tock/libtock-c/pull/88#discussion_r439862359

bradjc:
Fixes #90


ppannuto (approved):
bors r+

PR #206, opened 2021-04-20
transition CI to github actions

hudson-ayers:
Travis-CI seems to have stopped working. It is probably fixable, but given that Tock and libtock-rs already use github actions, this PR transitions our CI to github actions instead.

I *think* this covers everything but I have not tested the bors setup. Also, this now uses a third party Github Action to install the arm toolchain (https://github.com/fiam/arm-none-eabi-gcc). I could not get the existing install technique to work -- tar kept throwing errors (maybe we used too much disk space?) and this was easier. If someone else wants to dig into that more please feel free.


hudson-ayers:
closing this PR in favor of an identical one started on an in-repo branch. The CI wouldn't start until I had done so (and only shows up here because the commit is identical to the one on the in-repo branch)

PR #207, opened 2021-04-20
transition CI to github actions

hudson-ayers:
Travis-CI seems to have stopped working. It is probably fixable, but given that Tock and libtock-rs already use github actions, this PR transitions our CI to github actions instead.

I *think* this covers everything but I have not tested the bors setup. Also, this now uses a third party Github Action to install the arm toolchain (https://github.com/fiam/arm-none-eabi-gcc). I could not get the existing install technique to work -- tar kept throwing errors (maybe we used too much disk space?) and this was easier. If someone else wants to dig into that more please feel free.

This no longer caches the toolchains in the way that our travis-CI setup did, but the build typically completes in 3-4 minutes anyway so I don't think it is worth the effort.


alistair23 (approved):

bradjc:
Can we split the checks up like in Tock? As in have formatting be a separate check, say.

hudson-ayers:
> Can we split the checks up like in Tock? As in have formatting be a separate check, say.

Done

alistair23 (approved):

.github/workflows/ci.yml
```diff
@@ -0,0 +1,55 @@
...
+    runs-on: ${{ matrix.os }}
+
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: false # LVGL makefile manually installs the submodule
+      - uses: fiam/arm-none-eabi-gcc@v1
+        with:
+          release: '9-2019-q4' # The arm-none-eabi-gcc release to use.
```
ppannuto, line 27: FWIW, my machine is currently on `10-2020-q4-major` -- since we have some more aggressive warning checks, probably most useful for CI to be on the newest gcc version

ppannuto (approved):
LGTM, though I'd see if we can't bump the gcc version.

Thanks for doing the legwork here.

.github/workflows/ci.yml
hudson-ayers: 9-2019-q4 is what the CI ran on for Travis, but yeah I am happy to bump it here as well.

hudson-ayers: Though I will point out that 9-2019-q4 is the newest version that you can currently get through `apt` on ubuntu-20.04 (which is how we recommend ubuntu users install the toolchain)

bradjc (approved):

PR #211, opened 2021-05-11
driver_exists function needs to accept any success case

bradjc:
Currently it only accepts success with no args. This breaks for a driver like ADC.


lschuermann (approved):
I'm wary to rely on the fact that all success variants are >= 128, but it seems technically valid and that's the C way of doing things I suppose. 😄

PR #212, opened 2021-05-18
HMAC: Initial support for HMAC and HMAC test

alistair23:


hudson-ayers (approved):

alistair23:
This should be ready to merge

hudson-ayers:
I assumed this was blocked on https://github.com/tock/tock/pull/2586 -- does this work with the existing hmac driver?

alistair23:
Yep! There is no change to the HMAC driver so this will work. Although you will hit the double grant enter panic, but the actual interface is working.

PR #213, opened 2021-06-07
SHA: Initial support for SHA operations

alistair23:


examples/tests/sha/README.md
```diff
@@ -0,0 +1,4 @@
+SHA
+===
+
+This test performs a SHA operation on a string and prints the output
```
ppannuto, line 4: Can you add an example of what correct output looks like when this runs?

ppannuto (approved):
bors d+

Pending on kernel update.

Looks good; example output would be nice to add, but that's a quick fix and doesn't need to gate on me.

Issue #217, opened 2021-07-06
Build fails on arm-none-eabi-gcc 11.1.0


cyberbemon:
Followed the instructions from the readme. But my build fails. My gcc version is
```
arm-none-eabi-gcc (Arch Repository) 11.1.0
Copyright (C) 2021 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

I get the following error during the build process. Full log attached. 

```
In function 'text_screen_allow',
    inlined from 'text_screen_init' at ../../../../libtock/text_screen.c:50:11:
../../../../libtock/text_screen.c:37:28: error: '<unknown>' may be used uninitialized [-Werror=maybe-uninitialized]
   37 |   allow_ro_return_t aval = allow_readonly(DRIVER_NUM_TEXT_SCREEN, 0, ptr, size);
      |                            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

```
[log.txt](https://github.com/tock/libtock-c/files/6768987/log.txt)

ppannuto:
Huh -- looks like gcc has added a more advanced check in the newest version (n.b. OS X / brew still on 10.2.1, so hard to repro here). Technically, the compiler is correct, we _are_ passing an uninitialized buffer to the kernel here. It's fine to do so, since this is just setting up a buffer for a screen to draw from later.

The two fixes are:
 1. Change `malloc` to `calloc` so this memory is zero'd. Small performance hit, but maybe a nicer interface for library users
 2. Wrap in `#pragma GCC diagnostic push        #pragma GCC diagnostic ignored "-Wmaybe-uninitialized"     #pragma GCC diagnostic pop` as an explicit marker that this is intentionally passed uninit'd.

@alexandruradovici  I'm fine with either approach, as the primary user of the screen drivers, do you have a preference?

PR #218, opened 2021-07-06
add missing const qualifiers

ppannuto:
With new allow_readonly syscall, interfaces should reflect immutable buffer.

This might make #217 happy, but I doubt it. The interface actually _does_ pass uninitialized memory to the initial allow, which is fine, as we haven't asked the screen to draw the buffer yet. Still, the little const qualifiers should definitely be added; the initialization is separate concern.


alistair23 (approved):

hudson-ayers (approved):
bors r+

ppannuto:
bors retry

PR #219, opened 2021-07-09
Updating CRC library for new system call capsule

phil-levis:
The system call capsule removes the two SAM4L-specific algorithms from CRC and changes the compute command from 2 to 1. This updates libtock-c for the new API. It also adds a test case for a very large buffer.


alistair23 (approved):

hudson-ayers:
I believe this is blocked on Phil's remaining changes to CRC in the kernel being submitted as a fresh PR

lschuermann:
https://github.com/tock/tock/pull/2662 is in, removing the `blocked` label.

hudson-ayers (approved):

PR #223, opened 2021-07-24
Fix i2c_master

Gerharddc:
TOCK_I2C_MASTER_CB should be 0 not 1 according to https://github.com/tock/tock/blob/9f159e0c06e27b2357e3b20be6c90b3396bbd7b9/capsules/src/i2c_master.rs#L200 . I have also confirmed that it only works with this change.


alistair23 (approved):

ppannuto (approved):
bors r+

PR #231, opened 2021-08-09
Add rv32i target (LiteX sim) and improve nix-shell RISC-V toolchain integration

lschuermann:
### Pull Request Overview

This pull request adds support for `rv32i` targets in the Makefile infrastructure.

It further improves (rather: introduces) RISC-V toolchain support in the `shell.nix` Nix expression by providing an appropriate toolchain with newlib support. Because this toolchain is exposed with the tuple-prefix `riscv32-none-elf` and the tuples of the RISC-V embedded toolchains tend to vary between distributions, and because 32-bit-configured toolchains are sufficient, this further makes the toolchain prefixes configurable through an environment variable.

It also updates the `elf2tab` version provided in the Nix expression.

### Testing Strategy

This pull request was tested by compiling some ARM Cortex-M, RISC-V 32-bit I, IMC and IMAC apps in a Nix environment.

### TODO or Help Wanted

N/A


Configuration.mk
```diff
@@ -103,12 +103,13 @@ ELF2TAB_ARGS += -n $(PACKAGE_NAME)
...
-TOOLCHAIN_cortex-m7 := arm-none-eabi
-TOOLCHAIN_rv32imac := riscv64-unknown-elf
-TOOLCHAIN_rv32imc := riscv64-unknown-elf
+TOOLCHAIN_cortex-m0 ?= arm-none-eabi
+TOOLCHAIN_cortex-m3 ?= arm-none-eabi
+TOOLCHAIN_cortex-m4 ?= arm-none-eabi
+TOOLCHAIN_cortex-m7 ?= arm-none-eabi
+TOOLCHAIN_rv32imac ?= riscv64-unknown-elf
+TOOLCHAIN_rv32imc ?= riscv64-unknown-elf
+TOOLCHAIN_rv32i ?= riscv64-unknown-elf
```
ppannuto, line 112: Is this a required change?

The `:=` operator performs immediate binding, while the `=` and `?=` operators do lazy binding. We've run into issues with the nested macros in makefiles and lazy binding before, so I'm a bit leery about changing this if we don't need to.

lschuermann: Mhm. Would it help to do `?=` and then rebind with `:=` afterwards? I'm not super familiar with Makefile specifics. :smile: 

It's required in the sense that different platforms expose the RISC-V 32bit embedded compilers under prefixes (target tuples) given on their default configuration. [To cite five-embeddev.com:](https://five-embeddev.com/toolchain/2019/06/26/gcc-targets/) "Depending on how the compiler was configured at build time the name may be riscv-none-embed-gcc, or riscv32-unknown-elf-gcc or riscv64-unknown-elf-gcc, however they can all compile for any base ISA and extensions and address size." I'm fairly certain that in terms of address size that's not necessarily true in every case, but in general they are correct.

I think that different distributions generally provide different target tuples. Nix just happens to provide either a `riscv64-unknown-linux-gnu` or `riscv32-none-elf` target, neither of which match the expected `riscv64-unknown-elf`. I could try and look into exposing the generated compilers under a different target tuple but this will prevent these builds from ever being cached by the Nix people and just suits Nix as a single platform. Eventually people will have some weird setup and run into the same issues...

ppannuto: Ugh, yeah, what a mess.

Historically our approach has been to do platform detection and then bind the known name for that platform (or offering / automatically installing toolchain if missing, since we know how to do it now that we know what platform we're on).

TBH, not sure if `?=` followed by `:=` works the way we would want, but I think that platform detection + installation is more user-friendly anyway

lschuermann: Okay, so I should change it to something like _if is nix shell then "riscv32-none-elf" else "riscv64-unknown-elf"_? I guess we still need a fallback, we can't enumerate every possible platform...

Alternatively we can just iterate through the set of all known-working tuple combinations (i.e. "riscv{32,64}-{unknown,none}-{elf,linux-gnu}") and use the first one we find. However, I don't know how to do that elegantly in a Makefile and we'd still need an override -- some options are baked into some compilers and not overriden by us, for instance in Nix I build a custom newlib and tell the compiler to use that as it's libc.

ppannuto: Yeah, `command` is the canonical way to do this in bash IIUC, i.e. from the `Configuration.mk`:
```make
# Check if elf2tab exists, if not, install it using cargo.
ELF2TAB ?= elf2tab
ELF2TAB_EXISTS := $(shell $(SHELL) -c "command -v $(ELF2TAB)")
ifndef ELF2TAB_EXISTS
  $(shell cargo install elf2tab)
endif
```

I think it's actually pretty reasonable to enumerate every platform, the list of commonly used things is small (apt, yum, brew, nix is our list to-date I believe). Then hard error out and ask someone on a funkier platform to add the steps for their platform.

Configuration.mk
```diff
@@ -154,6 +161,7 @@ override CPPFLAGS_rv32imac += \
 override LINK_LIBS_rv32 += \
       -lc -lgcc -lm -lstdc++ -lsupc++
 
+override LINK_LIBS_rv32i  += $(LINK_LIBS_rv32)
```
bradjc, line 164:
```suggestion
override LINK_LIBS_rv32i   += $(LINK_LIBS_rv32)
```

bradjc (approved):

bradjc:
Why does this not change `TOCK_TARGETS`?

lschuermann:
Assuming I didn't screw up copying these `TOCK_TARGET` definitions from my shell history, this should be ready.

@ppannuto I didn't add a check for the precise distribution (as I don't have a portable way to do that anyways, not everything follows LSB!  :smile:) so instead with this PR we're starting to just try known-good tuple names. I tested building with and without different toolchains and everything seems to work just as expected. Although with Makefiles you'll never know for sure whether it breaks somewhere ;)

Also, the shipped RISC-V newlibs are a blessing, even though with Nix it's only a single command, my laptop doesn't sweat for hours building toolchains now, we can just use the prebuilt ones on most distributions now.

lschuermann:
Argh, the changes in `libnrfserialization/headers.tar.gz` are exclusively ownership and timestamp information. Blocked this PR, we should modify the tar command to avoid storing this information or dummy information such that we don't re-vendor this archive if nothing really changed.

bradjc:
I added some makefile cleanups.

lschuermann:
Alright, I will push the `rv32i` lvgl build now.

lschuermann:
Doing a final test of the changes now.

lschuermann:
> having trouble with newlib's `rv32i` build

I cannot reproduce this issue on a Ubuntu 20.04 installation. Thus I'd vouch that this is now ready, and I'm going to take an in-depth look of where the linking issue lies and create a follow-up PR. This already contains some significant improvements to the Nix build system and allows other systems to build for the `rv32i` Simulator target.

bradjc (approved):

PR #232, opened 2021-08-09
examples/tests/mpu_walk_region: handle missing button driver

lschuermann:
### Pull Request Overview

This pull request makes the `mpu_walk_region` test check for errors on a call to `button_count`, to handle boards which don't have a button driver installed (such as the LiteX simulator).

### Testing Strategy

This pull request was tested by running in the LiteX simulator. To deliberately overrun the memory region the user must set the variable manually. I expect users who are running on boards without buttons _and_ run MPU tests to be more advanced and thus be able to do this quite easily.

### TODO or Help Wanted

N/A


ppannuto (approved):
bors r+

Reasonable balance here IMO

PR #233, opened 2021-08-09
fix udp tests that relied on 15.4 driver being present

hudson-ayers:
This ensures that the udp_rx test does not attempt to block on the 15.4 driver since that driver is no longer present on Imix due to code size limitations.


lschuermann (approved):

ppannuto (approved):
bors r+

PR #234, opened 2021-08-09
Increase some app allocated sizes

alistair23:
In order for the OpenTitan release tests to pass let's increase some of the default app sizes.

### DIFF
--- a/examples/tests/malloc_test01/Makefile
+++ b/examples/tests/malloc_test01/Makefile
@@ -6,6 +6,8 @@ TOCK_USERLAND_BASE_DIR = ../../..
 # Which files to compile.
 C_SRCS := $(wildcard *.c)
 
+STACK_SIZE       = 4096
+
 # Include userland master makefile. Contains rules and flags for actually
 # building the application.
 include $(TOCK_USERLAND_BASE_DIR)/AppMakefile.mk
--- a/examples/tests/stack_size_test01/Makefile
+++ b/examples/tests/stack_size_test01/Makefile
@@ -6,7 +6,9 @@ TOCK_USERLAND_BASE_DIR = ../../..
 # Which files to compile.
 C_SRCS := $(wildcard *.c)
 
-STACK_SIZE := 900
+STACK_SIZE       := 2048
+APP_HEAP_SIZE    := 4096
+KERNEL_HEAP_SIZE := 2048
 
 # Include userland master makefile. Contains rules and flags for actually
 # building the application.
--- a/examples/tests/stack_size_test02/Makefile
+++ b/examples/tests/stack_size_test02/Makefile
@@ -7,6 +7,9 @@ TOCK_USERLAND_BASE_DIR = ../../..
 C_SRCS := $(wildcard *.c)
 
 STACK_SIZE := 9001
+# Required to work on RISC-V boards
+APP_HEAP_SIZE    = 4096
+KERNEL_HEAP_SIZE = 2048
 
 # Include userland master makefile. Contains rules and flags for actually
 # building the application.
--- a/examples/tests/mpu_walk_region/Makefile
+++ b/examples/tests/mpu_walk_region/Makefile
@@ -6,7 +6,7 @@ TOCK_USERLAND_BASE_DIR = ../../..
 # Which files to compile.
 C_SRCS := $(wildcard *.c)
 
-STACK_SIZE := 1024
+STACK_SIZE := 2048
 
 # Include userland master makefile. Contains rules and flags for actually
 # building the application.
--- a/examples/tests/printf_long/Makefile
+++ b/examples/tests/printf_long/Makefile
@@ -6,6 +6,10 @@ TOCK_USERLAND_BASE_DIR = ../../..
 # Which files to compile.
 C_SRCS := $(wildcard *.c)
 
+# Required to work on RISC-V boards
+APP_HEAP_SIZE    = 2048
+KERNEL_HEAP_SIZE = 2048
+
 # Include userland master makefile. Contains rules and flags for actually
 # building the application.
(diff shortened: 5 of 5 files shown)


examples/tests/stack_size_test01/Makefile
```diff
@@ -6,7 +6,9 @@ TOCK_USERLAND_BASE_DIR = ../../..
 # Which files to compile.
 C_SRCS := $(wildcard *.c)
 
-STACK_SIZE := 900
+STACK_SIZE       := 2048
```
hudson-ayers, line 9: The point of this test is to try setting a non-default stack size, but this changes it to set to what is already the default stack size.

examples/tests/malloc_test01/Makefile
```diff
@@ -6,6 +6,8 @@ TOCK_USERLAND_BASE_DIR = ../../..
 # Which files to compile.
 C_SRCS := $(wildcard *.c)
 
+STACK_SIZE       = 4096
```
hudson-ayers, line 9: I am shocked this requires a 4kB stack on RISC-V -- all it does is malloc a 5, 15, and 20 byte buffer and iterate through each before printing a short string. 4kB is the stack size for the entire kernel on some boards.

I wonder if something else is going on here and changing the stack size is somehow masking the actual problem

examples/tests/stack_size_test02/Makefile
```diff
@@ -7,6 +7,9 @@ TOCK_USERLAND_BASE_DIR = ../../..
 C_SRCS := $(wildcard *.c)
 
 STACK_SIZE := 9001
+# Required to work on RISC-V boards
+APP_HEAP_SIZE    = 4096
+KERNEL_HEAP_SIZE = 2048
```
hudson-ayers, line 12: So without these changes you see a different failure instead of a stack error? I am confused because this app does almost nothing at all aside from attempting to set a too-large stack size -- why would the kernel need a 2kB heap for this? Is that just for the console driver?

hudson-ayers (changes requested):

examples/tests/malloc_test01/Makefile
bradjc: Well...it also prints "success" if it gets that far. And I'm pretty sure that is what the issue is.

bradjc:
The issue here is that the `printf()` stack in the risc-v toolchain is very different from the one in the arm-none toolchain. With risc-v, the printf stack allocates a _lot_ of memory before trying to print. Consider the following strace on riscv:

```
[0] function_call @0x40430088(0x40430060, 0x80004000, 0x1ef4, 0x80004000)
[0] memop(0, 0x80005340) = Success
[0] memop(10, 0x80004800) = Success
[0] memop(11, 0x80005340) = Success
[0] memop(1, 0x418) = SuccessU32(2147504960)
[0] memop(1, 0x8a8) = Failure(NOMEM)
[0] memop(1, 0x1000) = Failure(NOMEM)
[0] memop(1, 0x1000) = Failure(NOMEM)
[0] yield. which: 1
```

and on hail:

```
[0] function_call @0x30061(0x30038, 0x20006000, 0x2000, 0x20006020)
[0] memop(0, 0x20006a38) = Success
[0] memop(10, 0x20006800) = Success
[0] memop(11, 0x20006a38) = Success
[0] memop(1, 0x0) = SuccessU32(536898104)
[0] memop(1, 0x1b4) = SuccessU32(536898104)
[0] memop(1, 0x408) = SuccessU32(536898540)
[0] memop(1, 0x18) = SuccessU32(536899572)
[0] memop(1, 0x68) = SuccessU32(536899596)
[0] read-only allow(0x1, 1, @0x20007010, 0x5d) = AllowReadOnlySuccess(0x0, 0)
[0] remove_pending_upcalls[0x1:1] = 0 upcall(s) removed
[0] subscribe(0x1, 1, @0x30f0d, 0x0) = SubscribeSuccess(0x0, 0)
[0] cmd(0x1, 1, 0x5d, 0x0) = Success
[0] yield. which: 1
```

The risc-v implementation tries to allocate at least twice the memory.

hudson-ayers:
I see. It seems like maybe we should just be increasing the minimum values of these variables for RISC-V only, instead of making all apps pay the price.

bradjc:
> I see. It seems like maybe we should just be increasing the minimum values of these variables for RISC-V only, instead of making all apps pay the price.

Yes, but, um, we didn't think of this possibility when designing the interface for elf2tab. Requested memory is provided as a global flag for all ELF files...

bradjc:
With small newlib on arty e21:

```
Initialization complete. Entering main loop.
[0] function_call @0x40430088(0x40430060, 0x80004000, 0x1580, 0x80004000)
[0] memop(0, 0x800049cc) = Success
[0] memop(10, 0x80004800) = Success
[0] memop(11, 0x800049cc) = Success
[0] memop(1, 0x0) = SuccessU32(2147502540)
[0] memop(1, 0x1b4) = SuccessU32(2147502540)
[0] memop(1, 0x408) = SuccessU32(2147502976)
[0] memop(1, 0x18) = SuccessU32(2147504008)
[0] memop(1, 0x68) = SuccessU32(2147504032)
[0] read-only allow(0x1, 1, @0x80004fa8, 0x5d) = AllowReadOnlySuccess(0x0, 0)
[0] remove_pending_upcalls[0x1:1] = 0 upcall(s) removed
[0] subscribe(0x1, 1, @0x4043113c, 0x0) = SubscribeSuccess(0x0, 0)
[0] cmd(0x1, 1, 0x5d, 0x0) = Success
Hi welcome to Tock. This test makes sure that a greater than 64 byte message can be printed.
[0] schedule[0x1:1] @0x4043113c(0x5d, 0x0, 0x0, 0x0) = Ok(())
[0] yield. which: 1
```
Same as hail!

bradjc:
Fixed in #237.

PR #236, opened 2021-08-10
Build all apps

hudson-ayers:
The build_all script (which is run by CI) only searched 4 directories deep so some apps were not being built. This PR modifies that to 6 directories deep and fixes breakages which had gone uncaught. Depends on https://github.com/tock/libtock-c/pull/235

### DIFF
--- a/libtock/udp.c
+++ b/libtock/udp.c
@@ -77,7 +77,7 @@ ssize_t udp_send_to(void *buf, size_t len,
                     sock_addr_t *dst_addr) {
   // Set dest addr
   // NOTE: bind() must be called previously for this to work
-  // If bind() has not been called, command(COMMAND_SEND) will return ??
+  // If bind() has not been called, command(COMMAND_SEND) will return RESERVE
   int bytes = sizeof(sock_addr_t);
   memcpy(BUF_TX_CFG + bytes, dst_addr, bytes);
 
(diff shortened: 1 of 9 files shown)


alistair23 (approved):

bradjc (approved):

libtock/udp.c
```diff
@@ -90,12 +90,12 @@ ssize_t udp_send_to(void *buf, size_t len,
   if (!sval.success) return tock_status_to_returncode(sval.status);
 
   syscall_return_t ret = command(UDP_DRIVER, COMMAND_SEND, 0, 0);
-  if (ret.type == TOCK_SYSCALL_SUCCESS) {
-    return RETURNCODE_SUCCESS;
+  if (ret.type == TOCK_SYSCALL_FAILURE) {
+    return tock_status_to_returncode(ret.data[0]);
```
brghena, line 94: Is this right? Before the code would only `yield_for()` if the command failed. Now the code only `yield_for()` if the command succeeds. This new version does seem like normal behavior, I just wanted to check.

hudson-ayers: Yeah the old version did not work, which I caught in release testing. It looks like when porting to 2.0 I made a mistake and changed the logic. I didn't just sneak that change into this PR, it was part of #235 which this depended on.

brghena: Cool. I was just checking for things that seem weird. As long as you've thought it through and tested it, then it's good.

PR #237, opened 2021-08-11
Compile small newlib for RISC-V

bradjc:
After:

```
❯ cd ../tests/console_recv_short
❯ make RISCV=1
  LD        build/rv32imac/rv32imac.0x20040040.0x80002800.elf
  LD        build/rv32imac/rv32imac.0x40430060.0x80004000.elf
  LD        build/rv32imac/rv32imac.0x40440060.0x80007000.elf
  LD        build/rv32imc/rv32imc.0x20030080.0x10005000.elf
Application size report for target cortex-m0:
   text	   data	    bss	    dec	    hex	filename
   6136	    292	   2500	   8928	   22e0	build/cortex-m0/cortex-m0.elf
Application size report for target cortex-m3:
   text	   data	    bss	    dec	    hex	filename
   6116	    292	   2500	   8908	   22cc	build/cortex-m3/cortex-m3.elf
Application size report for target cortex-m4:
   text	   data	    bss	    dec	    hex	filename
   6116	    292	   2500	   8908	   22cc	build/cortex-m4/cortex-m4.elf
Application size report for target cortex-m7:
   text	   data	    bss	    dec	    hex	filename
   6116	    292	   2500	   8908	   22cc	build/cortex-m7/cortex-m7.elf
Application size report for target rv32imac.0x20040040.0x80002800:
   text	   data	    bss	    dec	    hex	filename
   6816	    148	   2500	   9464	   24f8	build/rv32imac/rv32imac.0x20040040.0x80002800.elf
Application size report for target rv32imac.0x40430060.0x80004000:
   text	   data	    bss	    dec	    hex	filename
   6816	    148	   2500	   9464	   24f8	build/rv32imac/rv32imac.0x40430060.0x80004000.elf
Application size report for target rv32imac.0x40440060.0x80007000:
   text	   data	    bss	    dec	    hex	filename
   6816	    148	   2500	   9464	   24f8	build/rv32imac/rv32imac.0x40440060.0x80007000.elf
Application size report for target rv32imc.0x20030080.0x10005000:
   text	   data	    bss	    dec	    hex	filename
   9476	    148	   2500	  12124	   2f5c	build/rv32imc/rv32imc.0x20030080.0x10005000.elf
```


Configuration.mk
```diff
@@ -152,11 +152,25 @@ override CPPFLAGS_rv32imac += \
       -Wl,--no-relax   # Prevent use of global_pointer for riscv
 
 override LINK_LIBS_rv32 += \
-      -lc -lgcc -lm -lstdc++ -lsupc++
+      -lgcc -lstdc++ -lsupc++
 
 override LINK_LIBS_rv32imc  += $(LINK_LIBS_rv32)
 override LINK_LIBS_rv32imac += $(LINK_LIBS_rv32)
 
+override LEGACY_LIBS_rv32 += \
```
lschuermann, line 160:
```suggestion
override LEGACY_LIBS_rv32i += \
```

Configuration.mk
```diff
@@ -152,11 +152,25 @@ override CPPFLAGS_rv32imac += \
...
+      $(TOCK_USERLAND_BASE_DIR)/newlib/rv32/rv32i/libc.a\
+      $(TOCK_USERLAND_BASE_DIR)/newlib/rv32/rv32i/libm.a
+
+override LEGACY_LIBS_rv32im += \
+      $(TOCK_USERLAND_BASE_DIR)/newlib/rv32/rv32im/libc.a\
+      $(TOCK_USERLAND_BASE_DIR)/newlib/rv32/rv32im/libm.a
+
+override LEGACY_LIBS_rv32imc += $(LEGACY_LIBS_rv32im)
```
lschuermann, line 168: This isn't great. This means that IMC targets which do support compressed instructions won't use them, bloating the binary and making it slower. Judging from the output that does seem to be the case:

```
Application size report for target rv32imac.0x40440060.0x80007000:
   text	   data	    bss	    dec	    hex	filename
   6816	    148	   2500	   9464	   24f8	build/rv32imac/rv32imac.0x40440060.0x80007000.elf
Application size report for target rv32imc.0x20030080.0x10005000:
   text	   data	    bss	    dec	    hex	filename
   9476	    148	   2500	  12124	   2f5c	build/rv32imc/rv32imc.0x20030080.0x10005000.elf
```

lschuermann:
This is great. While I'm generally somewhat opposed to vendoring binaries I agree the RISC-V ecosystem (esp. embedded) currently isn't in a state where we can reasonably expect users to figure this out themselves and build their own `newlib`. This is also a great building block for #231 which will get easier because of this.

newlib/build-riscv.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+
+NEWLIB_SRC_DIR=$1
+
+$NEWLIB_SRC_DIR/configure --target=riscv64-unknown-elf \
```
hudson-ayers, line 5: why is this riscv64?

bradjc: Because gcc is not user friendly? I have no idea. These are the library versions that are generated:

```
drwxr-xr-x   3 bradjc  staff    96B Aug 11 10:34 rv32i
drwxr-xr-x   3 bradjc  staff    96B Aug 11 10:35 rv32iac
drwxr-xr-x   3 bradjc  staff    96B Aug 11 10:34 rv32im
drwxr-xr-x   3 bradjc  staff    96B Aug 11 10:35 rv32imac
drwxr-xr-x   3 bradjc  staff    96B Aug 11 10:35 rv32imafc
drwxr-xr-x   3 bradjc  staff    96B Aug 11 10:36 rv64imac
```

Configuration.mk
bradjc: If someone can get an `imc` version to build, we could easily update to it.

newlib/build-riscv.sh
lschuermann: I have an explanation in #231. Basically the tuple indicates only the default compiler configuration.

ppannuto:
I think this is great -- we already vendor newlib for arm for the same reason anyway. Speaking of which, #165 accidentally got lost I think. I'm going to pull those changes into here since I think it would be weird to have arm and riscv on different newlib verisons. Also bumping libc++ while I'm at it.

Computer is churning away, should have commits in ~hour

ppannuto (approved):

bradjc:
@alistair23

alistair23 (commented):
I don't like this.

Overall the code looks fine, but I don't like the concept.

I don't think we should be vendoring in a libc. libc should come from a users toolchain. The vendored version here quickly becomes out of date (as the ARM version is until this PR). As well as that we introduce all sorts of weird license issues. newlib has a strange license, so if someone wants to avoid that (by using [picolibc](https://github.com/picolibc/picolibc) for example) they can't as we have vendored them in.

For the size issue that is something we should bring up with upstream distros instead of just hacking around it, by building our own libc.

bradjc:
> I don't think we should be vendoring in a libc. libc should come from a users toolchain. The vendored version here quickly becomes out of date (as the ARM version is until this PR). As well as that we introduce all sorts of weird license issues. newlib has a strange license, so if someone wants to avoid that (by using [picolibc](https://github.com/picolibc/picolibc) for example) they can't as we have vendored them in.

Changing the libc being used is as simple as setting an environment variable (or at least we could make it that easy). We aren't customizing newlib in any way that means libtock-c is incompatible with any generic libc implementation.


> For the size issue that is something we should bring up with upstream distros instead of just hacking around it, by building our own libc.

Presumably you are using a custom libc? But even then you are having memory issues with printf, which suggests this is a hard problem to get right. Also, _we can't even get the Mac OS riscv toolchain to ship a prebuilt 32 bit compiler_. Fixing this on ever platform for every user seems, frankly, impossible. I don't see what choice we have other than to provide our own prebuilt libc versions and let users override the variables if needed.

alistair23:
> We aren't customizing newlib in any way that means libtock-c is incompatible with any generic libc implementation.

We are changing the build flags though. So if someone uses their own newlib it might not work

> Presumably you are using a custom libc?

Nope, just the one shipped with my distro

I agree the current situation is not great. But I'm not convinced just locking in our own libc is the solution. Keeping it up to date is a pain and something we don't really do. By including this we are now in charge of maintaining it and backporting patches. 

There was a [newlib CVE](https://nvd.nist.gov/vuln/detail/CVE-2021-3420) recently, did we patch this to fix it? For a secure OS it isn't great that we have critical issues in our libraries that I don't think has still been fixed.

By leaving this up to the user to provide (via their toolchain or build it themselves) we don't have to worry about maintaining it.

Also it's strange that those flags aren't enabled by default anywhere else, that makes me think maybe we are missing something and we might not want them all set.

bradjc:
If the correct PIC support ever happens we have to do this anyway. That's why we vendor the cortex-m version.

To use system libc all we need to get the build system to do is remove `$(LEGACY_LIBS_xxx)` and add back `-lm -lc`.

We want the kernel to be secure, we don't make any expectations about userland. But even so, we update rustc every two months (ish), we could do the same here, it's just I don't think anyone is that concerned about it. Also, we don't have "sudo" apps, so our exposure to system-level issues is low.

bradjc:
Not a perfect solution, but at least a stop gap for now.

bors r+

PR #238, opened 2021-08-12
make: more space for hifive1b header

bradjc:
Since we are now adding more TLVs we need to allocate more space before the start of the actual binary for hifive1b.


ppannuto (approved):
bors r+

PR #246, opened 2021-08-20
Use new exit syscall on process finish

bradjc:
I've been meaning to write this PR for a long time, but now that we have the exit syscall I think we should actually use it. If we had exit from the beginning I don't think we would end a process by having it sit in a yield loop, we would just call exit.

Example pconsole:

```
Initialization complete. Entering main loop.
Hello World!
Hi welcome to Tock. This test makes sure that a greater than 64 byte message can be printed.
And a short message.
list
 PID    Name                Quanta  Syscalls  Dropped Callbacks  Restarts    State  Grants
  0	printf_long              0        17                  0         0  Terminated    0/14
  1	blink                    0        39                  0         0  Yielded    1/14
  2	c_hello                  0         7                  0         0  Terminated    0/14
```

From the discussion on the PR, an oddity that comes from our current approach is that this c app:

```
int main (void) {
  return 1;
}
```

Will not exit with code 1, instead it will just yield.

### DIFF
--- a/examples/c_hello/main.c
+++ b/examples/c_hello/main.c
@@ -17,5 +17,5 @@ static void nop(
 
 int main(void) {
   putnstr_async(hello, strlen(hello), nop, NULL);
-  return 0;
+  yield();
 }
--- a/examples/c_hello/main.c
+++ b/examples/c_hello/main.c
@@ -17,5 +17,9 @@ static void nop(
 
 int main(void) {
   putnstr_async(hello, strlen(hello), nop, NULL);
+  // Because we used the async method (as opposed to something synchronous,
+  // such as printf), we must explicitly wait for the asynchronous write to complete.
   yield();
+  // Now we are done.
+  return 0;
 }
--- a/examples/courses/2018-11-SenSys/sensys_udp_rx/main.c
+++ b/examples/courses/2018-11-SenSys/sensys_udp_rx/main.c
@@ -101,7 +101,6 @@ int main(void) {
       printf("Failed to bind to socket %d\n", result);
       break;
   }
-
   // Note: Prior to 2.0, libtock-c kept apps alive in an implicit busy loop.
   // Since 2.0, apps must now explicitly include their own event loop to
   // avoid termination.
--- a/examples/courses/2018-11-SenSys/sensys_udp_rx/main.c
+++ b/examples/courses/2018-11-SenSys/sensys_udp_rx/main.c
@@ -101,6 +101,8 @@ int main(void) {
       printf("Failed to bind to socket %d\n", result);
       break;
   }
-  /* Tock keeps the app alive waiting for callbacks after
-   * returning from main, so no need to busy wait */
+
+  while (1) {
+    yield();
+  }
 }
--- a/examples/courses/2018-11-SenSys/sensys_udp_rx/main.c
+++ b/examples/courses/2018-11-SenSys/sensys_udp_rx/main.c
@@ -102,6 +102,9 @@ int main(void) {
       break;
   }
 
+  // Note: Prior to 2.0, libtock-c kept apps alive in an implicit busy loop.
+  // Since 2.0, apps must now explicitly include their own event loop to
+  // avoid termination.
   while (1) {
     yield();
   }
--- a/libtock/crt0.c
+++ b/libtock/crt0.c
@@ -328,8 +328,7 @@ void _c_start_pic(uint32_t app_start, uint32_t mem_start) {
     }
   }
 
-  main();
-  exit(0);
+  exit(main());
 }
 
 // C startup routine for apps compiled with fixed addresses (i.e. no PIC).
@@ -363,6 +362,5 @@ void _c_start_nopic(uint32_t app_start, uint32_t mem_start) {
   char* bss_start = (char*)(myhdr->bss_start + mem_start);
   memset(bss_start, 0, myhdr->bss_size);
 
-  main();
-  exit(0);
+  exit(main());
 }
--- a/libtock/crt0.c
+++ b/libtock/crt0.c
@@ -1,4 +1,5 @@
 #include "tock.h"
+#include <stdlib.h>
 #include <string.h>
 
 #if defined(STACK_SIZE)
@@ -328,9 +329,7 @@ void _c_start_pic(uint32_t app_start, uint32_t mem_start) {
   }
 
   main();
-  while (1) {
-    yield();
-  }
+  exit(0);
 }
 
 // C startup routine for apps compiled with fixed addresses (i.e. no PIC).
@@ -365,7 +364,5 @@ void _c_start_nopic(uint32_t app_start, uint32_t mem_start) {
   memset(bss_start, 0, myhdr->bss_size);
 
   main();
-  while (1) {
-    yield();
-  }
+  exit(0);
 }
--- a/libtock/crt0.c
+++ b/libtock/crt0.c
@@ -13,7 +13,7 @@
 #error Fixed STACK_SIZE.
 #endif
 
-extern int main(void);
+extern int main(int argc, char *argv[]);
 
 // Allow _start to go undeclared
 #pragma GCC diagnostic ignored "-Wmissing-declarations"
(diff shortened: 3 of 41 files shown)


brghena (changes requested):
This is a big change to make. There are apps that rely on the existing behavior, returning from main and expecting everything to execute in callbacks from then on out.

For example:
 * [`examples/ble_advertising/main.c`](https://github.com/tock/libtock-c/blob/280b23000a6ad7bfa4151b6e2e53beccedc1cc71/examples/ble-uart/main.c) (actually, `examples/ble*` do this)
 * [`examples/rot13_client/main.c`](https://github.com/tock/libtock-c/blob/8c358c5658a0170e1e94115ab9bb98913f97c1b0/examples/rot13_client/main.c) (and rot13_service)
 * [`examples/sensors/main.c`](https://github.com/tock/libtock-c/blob/4bfb80e9e80f94c3df5f8920cc6a93ca5948d165/examples/sensors/main.c)
 
Before 2.0 is the time to discuss a big breaking change like this, but generally, I'm a little on the side of disagreeing with it. I think the expectation for Tock apps should be that their callbacks continue to be run after returning. Mapping `exit()` to `tock_exit()` is reasonable, but I'm not sure that `exit()` should be the default behavior.

At the minimum, if you want this change you've got to fix all the apps it will break.

alexandruradovici:
I agree with @brghena, I was counting on this behavior. I think we can use the exit if and only if we can be sure that there is no registered callback.

lschuermann:
> I think we can use the exit if and only if we can be sure that there is no registered callback.

Independent of whether this is technically feasible, I somewhat strongly feel we shouldn't introduce behavior like this. Implicitly `exit`ing or not depending on previous application behavior would be _really_ confusing. Imagine an application wanting to exit on process finish and relying on this behavior. Then an error anywhere in the system -- including kernel capsules -- could cause a callback to not be unsubscribed and thus the application not exiting.

We should either introduce this now, or in a subsequent breaking release of `libtock-c`. Applications relying on this behavior can presumably always include a simple `while (1) { yield(); }` at the end of their main, at least as a quick compatibility fix.

alexandruradovici:
> > I think we can use the exit if and only if we can be sure that there is no registered callback.
> 
> Independent of whether this is technically feasible, I somewhat strongly feel we shouldn't introduce behavior like this. Implicitly `exit`ing or not depending on previous application behavior would be _really_ confusing. Imagine an application wanting to exit on process finish and relying on this behavior. Then an error anywhere in the system -- including kernel capsules -- could cause a callback to not be unsubscribed and thus the application not exiting.
> 
This is how nodejs works, it only exists if there is no other possible task that can be scheduled. In think this was a nice feature of Tock. In most of the cases applications will not want to exit, but schedule callbacks and wait for them. Having this dine automatically when main finishes I think is a plus. Another example is services using the ipc driver. 

On the other hand, applications that want to exit can use the `exit` system call.

> We should either introduce this now, or in a subsequent breaking release of `libtock-c`. Applications relying on the previous behavior can presumably always include a simple `while (1) { yield(); }` at the end of their main, at least as a quick compatibility fix.

Wouldn't this result in most of the applications ending like this?

libtock/crt0.c
```diff
@@ -326,9 +328,7 @@ void _c_start_pic(uint32_t app_start, uint32_t mem_start) {
   }
 
   main();
```
alistair23, line 330: What is `return 1` is used in `main()`?

I'm fully onboard with removing this `while (1)` loop and having apps do that if they want to, but this should be `exit(main())`

alistair23 (changes requested):

lschuermann:
I think one very important distinction in the way the NodeJS event loop works compared to a a Tock process is that NodeJS seems to have rather sophisticated mechanisms to check whether an event is pending (as outlined [here](https://nodejs.org/en/docs/guides/event-loop-timers-and-nexttick/)). In contrast to that, a subscription in Tock only allows to make conclusions about whether events could arrive at all, not if, or when, or how many.

Thus imagine the following case: an application subscribes a callback, errors somewhere, then does not unsubscribe -- even though there is no event scheduled which could possibly result in the callback scheduled. Now, as a result of runtime behaviour, the application would never exit, thus never possibly be scheduled to restart or free up resources for other processes.

I'm not saying that this programming paradigm is not valid, just that it should be an explicit decision. An app exiting once it reaches the end of its main is very explicit, and a `while (1) { yield(); }` hints developers at the fact that they (a) won't exit automatically and (b) can employ other techniques for forcing an exit such as a global app timeout etc.

bradjc:
I think libtock-js would reasonably wait until there is no more work to be done, but I think libtock-c should exit when main returns.

alexandruradovici:
@lschuermann I agree with you point of view.

I am still in favour of keeping the current behaviour and asking developers to explicitly use the exit system call if they want to exit. The idea behind this is that microcontroller apps usually do not exit and I think there will be more apps that use the while (1) than apps that will manually exit. I might be wrong 😃.

brghena:
My basis for disagreeing is along the same lines as Alexandru. The default behavior for the majority of embedded apps is to never exit, so that makes sense to me as the default when returning from `main`.

That being said, Tock has many other aspects which follow standard OS behavior rather than embedded-specific. And the standard C premise is that when `main` returns, your program is done. So I can definitely see the rationale for what Brad is proposing.

bradjc:
What would you expect this program to do?

```c
int main (void) {
  return 1;
}
```

I think it would be surprising to not get the return code in the kernel.

bradjc:
I am in favor of finding all of the apps that rely on a yield forever loop and adding the loop in at the end of main. That makes it explicit that they intend to run forever.

alexandruradovici:
> What would you expect this program to do?
> 
> 
> 
> ```c
> 
> int main (void) {
> 
>   return 1;
> 
> }
> 
> ```
> 
> 
> 
> I think it would be surprising to not get the return code in the kernel.

I have to admit that you have a point. I was under the impression that main was declared with a void return type as its return value is never used.

examples/c_hello/main.c
```diff
@@ -17,5 +17,5 @@ static void nop(
 
 int main(void) {
   putnstr_async(hello, strlen(hello), nop, NULL);
-  return 0;
+  yield();
```
ppannuto, line 20:
```suggestion
  // Because we used the async method (as opposed to something synchronous,
  // such as printf), we must explicitly wait for the asynchronous write to complete. 
  yield();
  // Now we are done.
  return 0;
```

I think this is a bit of a weird example, and being that it's the `c_hello` example---i.e., one of the simplest---some extra explanation is warranted here.

examples/courses/2018-11-SenSys/sensys_udp_rx/main.c
```diff
@@ -101,6 +101,7 @@ int main(void) {
       break;
   }
 
-  /* Tock keeps the app alive waiting for callbacks after
-   * returning from main, so no need to busy wait */
+  while (1) {
```
ppannuto, line 104:
```suggestion
  // Note: Prior to 2.0, libtock-c kept apps alive in an implicit busy loop.
  // Since 2.0, apps must now explicitly include their own event loop to
  // avoid termination.
  while (1) {
```

Give some indicator that something changed in case there's supporting text out somewhere that points to this.

ppannuto (commented):
Looks good, just two cases where I added some comments that I think help explain things a bit

ppannuto (dismissed):
I'm convinced this is a good idea and worth changing

brghena (dismissed):
I would like to discuss this briefly on the call this Friday to make sure everyone's aware and on board with the change.

These changes have resolved my initial concern though. And I do see the argument for maintaining typical C application behavior, rather than embedded-specific behavior.

alistair23 (approved):

bradjc:
Ok! I came back to this. I went through all of the examples and added `while(1) yield();` loops where needed.

lschuermann (dismissed):

alistair23 (approved):

alexandruradovici:
+1On Jan 25, 2024, at 05:18, Leon Schuermann ***@***.***> wrote:﻿
This is breaking the library API semantics in a pretty major way, so I propose we quickly talk about how this would interact with any release schedules on the call on Friday.

lschuermann:
@bradjc I can't see immediately why this is / was marked blocked. if it's no longer, feel free to remove. Marking last-call.

bradjc:
This passes CI finally!

lschuermann (approved):

bradjc:
Last call?

PR #247, opened 2021-08-20
ble-uart app does not need that much memory

bradjc:
Not sure why this was added in 156b67b2ae6925478e7e9d6ff69c7729ecd6817b


hudson-ayers (approved):
Tested on Imix, works fine!

PR #249, opened 2021-08-21
libtock: fix clobbering of RISC-V a4 register in syscalls

luismarques:
The a4 register was being clobbered but it wasn't marked as so. This bug was exposed when compiling apps with Clang.

### DIFF
--- a/libtock/tock.c
+++ b/libtock/tock.c
@@ -358,7 +358,7 @@ void yield(void) {
     register uint32_t a0  __asm__ ("a0")        = 1; // yield-wait
     register uint32_t wait_field __asm__ ("a1") = 0; // yield result ptr
     __asm__ volatile (
-      "li    a4, 0\n"
+      "li       a4, 0\n"
       "ecall\n"
       :
       : "r" (a0), "r" (wait_field)
@@ -377,7 +377,7 @@ int yield_no_wait(void) {
     register uint32_t a0  __asm__ ("a0") = 0; // yield-no-wait
     register uint8_t* a1  __asm__ ("a1") = &result;
     __asm__ volatile (
-      "li    a4, 0\n"
+      "li       a4, 0\n"
       "ecall\n"
       :
       : "r" (a0), "r" (a1)
@@ -392,11 +392,11 @@ int yield_no_wait(void) {
 void tock_restart(uint32_t completion_code) {
   register uint32_t a0  __asm__ ("a0") = 1; // exit-restart
   register uint32_t a1  __asm__ ("a1") = completion_code;
+  register uint32_t a4  __asm__ ("a4") = 6;
   __asm__ volatile (
-    "li    a4, 6\n"
     "ecall\n"
     :
-    : "r" (a0), "r" (a1)
+    : "r" (a0), "r" (a1), "r" (a4)
     : "memory");
   __builtin_unreachable();
 }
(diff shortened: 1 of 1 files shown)


lschuermann:
Great find, thanks for this contribution! Adding the `tock-2.0` label as a fix for this should definitely be included in the libtock-c release accompanying Tock 2.0.

In an effort to reduce the likelihood of this happening again, I'd be in favor of assigning the system call class ID ([reference](https://github.com/tock/tock/blob/master/doc/reference/trd104-syscalls.md#4-system-call-api)) stored in `a4` to a register-variable, similar to how the other registers are set. Less explicit assembly is always good. :smile: 

The compiler does optimize this down to a simple load immediate:
```asm
  ...
  register uint32_t a3  __asm__ ("a3") = arg2;
200302f8:	86ba                	mv	a3,a4
  // Place the Command system call class ID (2) in a4
  register uint32_t a4  __asm__ ("a4") = 2;
200302fa:	4709                	li	a4,2
  register int rtype __asm__ ("a0");
  register int rv1 __asm__ ("a1");
  register int rv2 __asm__ ("a2");
  register int rv3 __asm__ ("a3");
  __asm__ volatile (
200302fc:	00000073          	ecall
    "ecall\n"
    : "=r" (rtype), "=r" (rv1), "=r" (rv2), "=r" (rv3)
    : "r" (a0), "r" (a1), "r" (a2), "r" (a3), "r" (a4)
    : "memory");
  syscall_return_t rval = {rtype, {rv1, rv2, rv3}};
  ...
```

luismarques:
Changed to the suggested alternate approach.
For consistency, I also changed `yield_no_wait`, even though there `a4` was already marked as a clobber (and must remain so).

lschuermann:
I took the liberty to push one fix for `yield_no_wait`, which had an `a4` variable-register assigned but not used in the assembly. For `yield` and `yield_no_wait` we can't simply define a register as input _and_ clobber, it probably would've been required to define it as an input and output. Thus I've changed it back to use a simply load-immediate.

lschuermann (approved):

luismarques:
> I took the liberty to push one fix for `yield_no_wait`, which had an `a4` variable-register assigned but not used in the assembly. 

Thanks for spotting that it was missing from the input list.

> For `yield` and `yield_no_wait` we can't simply define a register as input _and_ clobber, it probably would've been required to define it as an input and output. Thus I've changed it back to use a simply load-immediate as `a4` was already marked as clobbered.

That one was correct as it was. A clobber indicates that a register is modified, even though it isn't listed in the outputs. I've changed it back, but kept your other changes (input list and whitespace changes). Let me know if this still looks good to you this way.

alistair23 (approved):

luismarques:
> That one was correct as it was. A clobber indicates that a register is modified, even though it isn't listed in the outputs. I've changed it back, but kept your other changes (input list and whitespace changes). Let me know if this still looks good to you this way.

I just noticed that when compiling with GCC it complains about that. Sorry, I hadn't realized there was a compilation issue, I thought you had spotted the issue from reading the PR code changes. In any case, I *think* it's a false positive in GCC's diagnostic, but I'll follow that up with the RISC-V GCC contributors. Register variables are a bit of an underspecified minefield. There was a proposal a couple of months ago that would improve on them a lot: <https://gcc.gnu.org/pipermail/gcc/2021-June/236269.html>. That's basically how LLVM already works internally.

I've force pushed your commit back. Thanks!

lschuermann:
@luismarques Haha, thanks, and no worries! I'm on vacation and only occasionally have access to my working environment. But I was almost certain it didn't work on `gcc` so I'm glad to have confirmation. :) 

I think this looks good now, and having `yield` use explicit load immediates is fine, given these are clobbered, caller-saved registers anyways.

lschuermann (approved):

hudson-ayers (approved):

bradjc (approved):

PR #250, opened 2021-08-21
Update UI libraries

alexandruradovici:
This pull request:
- update screen driver the new allow usage
- update touch driver
- add touch example
- update lvgl to version 8
- update lvgl example

This was tested using Tock 2.0

### DIFF
--- a/examples/lvgl/Makefile
+++ b/examples/lvgl/Makefile
@@ -3,17 +3,6 @@
 # Specify this directory relative to the current application.
 TOCK_USERLAND_BASE_DIR = ../..
 
-TOCK_TARGETS := cortex-m0\
-                cortex-m3\
-                cortex-m4\
-                cortex-m7\
-                rv32imac|rv32imac.0x20040060.0x80002800|0x20040060|0x80002800\
-                rv32imac|rv32imac.0x40430060.0x80004000|0x40430060|0x80004000\
-                rv32imac|rv32imac.0x40440060.0x80007000|0x40440060|0x80007000\
-                rv32imac|rv32imac.0x403B0060.0x3FCC0000|0x403B0060|0x3FCC0000\
-                rv32imc|rv32imc.0x20030080.0x10005000|0x20030080|0x10005000\
-                rv32imc|rv32imc.0x41000060.0x42008000|0x41000060|0x42008000
-
 # External libraries used
 EXTERN_LIBS += $(TOCK_USERLAND_BASE_DIR)/lvgl
 
@@ -21,8 +10,6 @@ override CFLAGS += -I$(TOCK_USERLAND_BASE_DIR)/lvgl
 
 # Which files to compile.
 C_SRCS += $(wildcard *.c)
-
-# include $(TOCK_USERLAND_BASE_DIR)/lvgl/lvgl.mk
 
 APP_HEAP_SIZE := 40000
 STACK_SIZE := 4096
--- a/examples/lvgl/Makefile
+++ b/examples/lvgl/Makefile
@@ -3,6 +3,17 @@
 # Specify this directory relative to the current application.
 TOCK_USERLAND_BASE_DIR = ../..
 
+TOCK_TARGETS := cortex-m0\
+                cortex-m3\
+                cortex-m4\
+                cortex-m7\
+                rv32imac|rv32imac.0x20040060.0x80002800|0x20040060|0x80002800\
+                rv32imac|rv32imac.0x40430060.0x80004000|0x40430060|0x80004000\
+                rv32imac|rv32imac.0x40440060.0x80007000|0x40440060|0x80007000\
+                rv32imac|rv32imac.0x403B0060.0x3FCC0000|0x403B0060|0x3FCC0000\
+                rv32imc|rv32imc.0x20030080.0x10005000|0x20030080|0x10005000\
+                rv32imc|rv32imc.0x41000060.0x42008000|0x41000060|0x42008000
+
 # External libraries used
 EXTERN_LIBS += $(TOCK_USERLAND_BASE_DIR)/lvgl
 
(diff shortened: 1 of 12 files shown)


alexandruradovici:
I'm not sure why this fails for RISCV, it seems to build on my computer.

lschuermann:
> I'm not sure why this fails for RISCV, it seems to build on my computer.

You presumably didn't build with the changes on latest master. These changes introduce a new `rv32i` target defintion. The CI tests the result of this being merged into master, which is where this error would occur. A simple `git rebase master`, along with a rebuild of `lvgl` should suffice.

hudson-ayers:
+1 to leons comments on the CI failure, but this otherwise looks good.

Note for other reviewers: the PR is much easier to review than it seems, most of the changes are in a configuration file that I don't think we need to worry about.

alexandruradovici:
> +1 to leons comments on the CI failure, but this otherwise looks good.
> 
> Note for other reviewers: the PR is much easier to review than it seems, most of the changes are in a configuration file that I don't think we need to worry about.

That is lvgl's configuration file, it changed a lot as lvgl has been updated with two releases.

alexandruradovici:
I have rebased it and rebuilt the riscv lvgl library. I used the [10.2.0 version](https://www.sifive.com/software) of the riscv compiler.

examples/touch/main.c
```diff
@@ -0,0 +1,72 @@
+#include <screen.h>
+#include <stdlib.h>
+#include <timer.h>
+#include <touch.h>
+
+static void tocuh_event (int status, int x, int y, void *ud __attribute__ ((unused))) {
```
bradjc, line 6:
```suggestion
static void touch_event (int status, int x, int y, void *ud __attribute__ ((unused))) {
```

examples/touch/Makefile
```diff
@@ -0,0 +1,13 @@
+# Makefile for user application
```
bradjc, line 1: Can you add a readme?

examples/lvgl/Makefile
```diff
@@ -3,6 +3,17 @@
...
+TOCK_TARGETS := cortex-m0\
+                cortex-m3\
+                cortex-m4\
+                cortex-m7\
+                rv32imac|rv32imac.0x20040060.0x80002800|0x20040060|0x80002800\
+                rv32imac|rv32imac.0x40430060.0x80004000|0x40430060|0x80004000\
+                rv32imac|rv32imac.0x40440060.0x80007000|0x40440060|0x80007000\
+                rv32imac|rv32imac.0x403B0060.0x3FCC0000|0x403B0060|0x3FCC0000\
+                rv32imc|rv32imc.0x20030080.0x10005000|0x20030080|0x10005000\
+                rv32imc|rv32imc.0x41000060.0x42008000|0x41000060|0x42008000
```
bradjc, line 15: Why do we need this here? This will very quickly become out of sync wit the main version in Configuration.mk.

alexandruradovici: I added this here as this example cannot be compiled for rv32i. Is there another way of excluding a target?

lschuermann: There really shouldn't be any reason why it wouldn't work for `rv32i`. I have seen some weird linking errors before, can you please post the error your experiencing?

alexandruradovici:
The data segment overflows. This needs around 24k of data.

examples/lvgl/Makefile
bradjc: The issue is the binary is too big (presumably because multiply instructions are being done manually?). I think we should update the linker script to allow for larger apps and remove this special case.

bradjc: I think a rebase and removing this should work.

alexandruradovici: Now I'm getting a different error related to then gcc library.

```
LD        build/rv32i/rv32i.0x00080060.0x40008000.elf
/Users/alexandru/Downloads/riscv64-unknown-elf-toolchain-10.2.0-2020.12.8-x86_64-apple-darwin/bin/../lib/gcc/riscv64-unknown-elf/10.2.0/../../../../riscv64-unknown-elf/bin/ld: cannot find -lstdc++
/Users/alexandru/Downloads/riscv64-unknown-elf-toolchain-10.2.0-2020.12.8-x86_64-apple-darwin/bin/../lib/gcc/riscv64-unknown-elf/10.2.0/../../../../riscv64-unknown-elf/bin/ld: cannot find -lsupc++
collect2: error: ld returned 1 exit status
make: *** [build/rv32i/rv32i.0x00080060.0x40008000.elf] Error 1
```

I am using the 10.2.0 version of gcc for risc-v.

hudson-ayers: That doesn't seem to be an issue in CI, however this error now surfaces: 
```
main.c:65:32: error: 'tocuh_event' undeclared (first use in this function); did you mean 'touch_event'?
   65 |     single_touch_set_callback (tocuh_event, NULL);
      |                                ^~~~~~~~~~~
      |                                touch_event
main.c:65:32: note: each undeclared identifier is reported only once for each function it appears in
```

alexandruradovici:
I was trying the `lvgl` example and it seems I forgot the build the `touch` example. It should work now.

hudson-ayers (approved):

examples/lvgl/Makefile
hudson-ayers: Are you still having an issue with gcc? If so I suspect it is because 2 weeks ago we updated c++ to use gcc 11.2, so your toolchain is out of date.

alexandruradovici: Yes, still fails on my side. For risc-v I am using gcc 10.2.0 from sifive. Do I need to get it from somewhere else? I suppose the gcc (eabi and x86) toolchain installed on my mac is not used for risc-v.

PR #261, opened 2021-09-02
Configuration.mk: Build system improvements

alistair23:
When running multiple apps on OpenTitan we need to build apps at a few
different addresses. This commit sets up the build system to build the
OpenTitan apps at all of the addresses required to run the automated app
tests included in the Tock kernel.

This greatly simplifies testing apps on OpenTitan.

The overhead of building an app for multiple different addresses is low,
so this doesn't extend the build time by very much.

While we are at it let's also split out the current Arty build addresses
to make things clear.

This PR also adds support for the Arch toolchain name.


hudson-ayers (approved):

ppannuto (approved):
bors r+

PR #268, opened 2021-09-21
Readme Updates

bradjc:
- brew install should work now
- no longer need newlib
- update usage instructions with tockloader changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,66 +71,11 @@ Prerequisites
    $ sudo apt install gcc-riscv64-unknown-elf
    ```
 
-   Unfortunately, Ubuntu does not currently (June 2020) provide a package for
-   RISC-V libc. We have created a .deb file you can use to install a suitable
-   libc based on newlib:
-   ```
-   $ wget http://cs.virginia.edu/~bjc8c/archive/newlib_3.3.0-1_amd64.deb
-   $ sudo dpkg -i newlib_3.3.0-1_amd64.deb
-   ```
-
-   If you would rather compile your own newlib-based libc, follow the steps below.
-   Section [newlib-nano][newlib-nano] describes some extra config options to
-   build a size optimised newlib.
-   ```
-   # Download newlib 3.3 from https://sourceware.org/newlib/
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -54,87 +54,28 @@ Prerequisites
...
-   $ unzip gcc-riscv64-unknown-elf-8.3.0-ubuntu.zip
-   # add gcc-riscv64-unknown-elf-8.3.0-ubuntu/bin to your `$PATH` variable.
-   ```
-
    **Arch**:
    ```
    $ sudo pacman -Syu riscv64-elf-gcc riscv32-elf-newlib
    ```
 
-   **newlib-nano**:
```
hudson-ayers, line 116: Might people not still want to use this for the size savings?

alistair23: libtock-c now uses it's own newlib so we don't need to manually build it

README.md
```diff
@@ -54,87 +54,28 @@ Prerequisites
...
-   Crosstool-NG.
-   ```
-   $ wget http://cs.virginia.edu/~bjc8c/archive/gcc-riscv64-unknown-elf-8.3.0-ubuntu.zip
-   $ unzip gcc-riscv64-unknown-elf-8.3.0-ubuntu.zip
-   # add gcc-riscv64-unknown-elf-8.3.0-ubuntu/bin to your `$PATH` variable.
-   ```
-
    **Arch**:
    ```
    $ sudo pacman -Syu riscv64-elf-gcc riscv32-elf-newlib
```
alistair23, line 76: It's worth removing the newlb dependency here

lschuermann: At least on Ubuntu 20.04, I haven't gotten compiling `libtock-c` to work with the standard `gcc-riscv64-unknown-elf` package at it couldn't find some libs. I can reproduce this error if desired. Thus in my CI workflows I'm still resorting to the prebuilt package by @bradjc. If it doesn't work out of the box, we should leave these instructions in.

bradjc: That prebuilt package is just newlib, right? Since we have built packages in the repo now is it still required? Can we look at errors and maybe figure out what is going wrong?

PR #269, opened 2021-09-27
libtock: Consistent error code for Invalid exit

ppannuto:
When reviewing #170, I noticed that there's a mix of `1` and `-1`
used to express the same error (I think)?

Looks like this came as part of the big batch change in 77e90c1f.

@phil-levis Did you have a reason for some of these being `1` and some `-1`?

### DIFF
--- a/libtock/tock.c
+++ b/libtock/tock.c
@@ -239,7 +239,7 @@ subscribe_return_t subscribe(uint32_t driver, uint32_t subscribe,
     subscribe_return_t rval = {false, (subscribe_upcall*)rv2, (void*)rv3, (statuscode_t)rv1};
     return rval;
   } else {
-    exit(-1);
+    exit(1);
   }
 }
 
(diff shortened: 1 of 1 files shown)


lschuermann:
In the kernel, the exit syscall parameters are represented as the following ([source](https://github.com/tock/tock/blob/5e1af28cacf5d8e120449fdf18f552e6a8655dfe/kernel/src/syscall.rs#L139-L145)):

```rust
    /// Structure representing an invocation of the Exit system call
    /// class. `which` is the exit identifier and `completion_code` is
    /// the completion code passed into the kernel.
    Exit {
        which: usize,
        completion_code: usize,
    },
```

Based on that (and also in line with us using strictly positive error codes across the system call boundary, see [this comment](https://github.com/tock/tock/pull/2511#issuecomment-860487932)) I'd argue it would be good to just follow the POSIX convention of `0` for generic success and higher, positive values for specific error conditions or other specialized exit values.

ppannuto:
Flipped to consistent `1`'s instead of `-1`'s

hudson-ayers (approved):

phil-levis (approved):

PR #270, opened 2021-11-30
ipc: register service callback with service name

jettr:
Use package name of service to register the service callback. This makes it more clear API and makes the kernel side easier.

It will allow us to drop any special case code in kernel for IPC and it will most likely allow us to drop the `NUM_UPCALLS` const generic as well (and we can just use `NUM_PROCS` for both upcall and allows)

See [RFC: Move allow_readwrite and allow_readonly from SyscallDriver into common Grant framework][0] for motivation

[0]:https://github.com/tock/tock/pull/2906

### DIFF
--- a/examples/rot13_service/main.c
+++ b/examples/rot13_service/main.c
@@ -23,7 +23,8 @@ static void rot13_callback(int pid, int len, int buf, __attribute__ ((unused)) v
 }
 
 int main(void) {
-  ipc_register_service_callback(rot13_callback, NULL);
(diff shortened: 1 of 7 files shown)


hudson-ayers (approved):

alistair23 (approved):

jettr:
FYI: this will need to land in tandem with the tock OS PR mentioned in the first post.

jettr:
Looks like it is time to land this change, so we can launch the CI again for the kernel change and let bors land the kernel change

PR #271, opened 2021-12-13
ipc: allow registering client callback with svc_id = 0

lschuermann:
### Pull Request Overview

With the changes of 321a5a01e4 ("ipc: register service callback with service name") and tock/tock#2906, it is possible for a service to be assigned the svc_id = 0. Thus we must allow clients to register callbacks with this svc_id.

### Testing Strategy

This pull request was tested by running the failing LiteX simulation CI workflow locally. This resolves one of the issues of the failing CI.

### TODO or Help Wanted

This pull request does not fix all of the issues of tock/tock#2906. Somehow these changes appear to affect either the load order or scheduling behavior of the rot13_{service,client} apps. Because IPC service discovery seems fundamentally racy, with currently no reliable way to tell whether notifying a service worked, it requires further changes to work reliably.

### DIFF
--- a/libtock/ipc.c
+++ b/libtock/ipc.c
@@ -1,7 +1,7 @@
 #include "ipc.h"
 #include "tock.h"
 
-int ipc_discover(const char* pkg_name, int* svc_id) {
+int ipc_discover(const char* pkg_name, size_t* svc_id) {
   int len = strlen(pkg_name);
 
   allow_ro_return_t prev = allow_readonly(IPC_DRIVER_NUM, 0, pkg_name, len);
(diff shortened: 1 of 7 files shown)


libtock/ipc.c
```diff
@@ -30,9 +30,6 @@ int ipc_register_service_callback(const char *pkg_name,
 }
 
 int ipc_register_client_callback(int svc_id, subscribe_upcall callback, void *ud) {
-  if (svc_id <= 0) {
```
alistair23, line 33: Won't this now allow negative values? Why not make `svc_id` a uint

lschuermann: Yes, a `uint32_t` is better suited for this purpose. We should change it everywhere though to be consistent.

lschuermann: Actually I've changed it to a `size_t`. While the ABI is currently only specified for 32-bit platforms, the kernel does represent the process IDs as `usize`s underneath, so it appears to make sense to use the native integer width in userspace as well.

alistair23 (approved):

jettr (approved):
Don't know how I missed this 0 check (but I guess that is what happens when you can't test directly though :( )

Thanks for the fix!

hudson-ayers (approved):
bors r+

Thanks for this fix!

PR #274, opened 2022-01-05
Alarm: Avoid silent failures and fixup command return function call.

thulithwilfred:
Signed-off-by: Wilfred Mallawa wilfred.mallawa@wdc.com

**Pull Request Overview**
We noticed a seperate bug in the alarm setting process. Where a call to `delay_ms()` would **silently fail** and hang the application. This PR aims to add a notice to when an alarm is not set, so as to not fail silently.

 Additionally, changing the syscall return value handler to the appropriate 'u32' variant instead of 'novalue'.

**Testing Strategy**
make RISCV=1

Application testing through qemu/opentitan.

**Documentation Updated**
N/A

### DIFF
--- a/libtock/alarm_timer.c
+++ b/libtock/alarm_timer.c
@@ -1,11 +1,8 @@
 #include "alarm.h"
 #include "internal/alarm.h"
 #include "timer.h"
-
-#include <console.h>
 #include <limits.h>
 #include <stdlib.h>
-
 
 // Returns 0 if a <= b < c, 1 otherwise
 static int within_range(uint32_t a, uint32_t b, uint32_t c) {
@@ -74,9 +71,7 @@ static void callback( __attribute__ ((unused)) int unused0,
     // has the alarm not expired yet? (distance from `now` has to be larger or
     // equal to distance from current clock value.
     if (alarm->dt > now - alarm->reference) {
-      if (alarm_internal_set(alarm->reference, alarm->dt) != RETURNCODE_SUCCESS)
-        putnstr("Err: failed to set Alarm\n", sizeof("Err: failed to set Alarm"));
-
+      alarm_internal_set(alarm->reference, alarm->dt);
       break;
     } else {
       root_pop();
@@ -89,7 +84,7 @@ static void callback( __attribute__ ((unused)) int unused0,
   }
 }
 
-void alarm_at(uint32_t reference, uint32_t dt, subscribe_upcall cb, void* ud, alarm_t* alarm) {
+int alarm_at(uint32_t reference, uint32_t dt, subscribe_upcall cb, void* ud, alarm_t* alarm) {
   alarm->reference = reference;
   alarm->dt        = dt;
   alarm->callback  = cb;
@@ -106,9 +101,9 @@ void alarm_at(uint32_t reference, uint32_t dt, subscribe_upcall cb, void* ud, al
   if (root_peek() == alarm) {
     alarm_internal_subscribe((subscribe_upcall*)callback, NULL);
 
-    if (alarm_internal_set(alarm->reference, alarm->dt) != RETURNCODE_SUCCESS)
-      putnstr("Err: failed to set Alarm\n", sizeof("Err: failed to set Alarm"));
+    return alarm_internal_set(alarm->reference, alarm->dt);
   }
+  return RETURNCODE_SUCCESS;
 }
 
 void alarm_cancel(alarm_t* alarm) {
@@ -122,8 +117,7 @@ void alarm_cancel(alarm_t* alarm) {
   if (root == alarm) {
     root = alarm->next;
     if (root != NULL) {
-      if (alarm_internal_set(root->reference, root->dt) != RETURNCODE_SUCCESS)
-        putnstr("Err: failed to set Alarm\n", sizeof("Err: failed to set Alarm"));
+      alarm_internal_set(root->reference, root->dt);
     }
   }
 
@@ -134,13 +128,13 @@ void alarm_cancel(alarm_t* alarm) {
 
 // Timer implementation
 
-void timer_in(uint32_t ms, subscribe_upcall cb, void* ud, tock_timer_t *timer) {
+int timer_in(uint32_t ms, subscribe_upcall cb, void* ud, tock_timer_t *timer) {
   uint32_t frequency;
   alarm_internal_frequency(&frequency);
   uint32_t interval = (ms / 1000) * frequency + (ms % 1000) * (frequency / 1000);
   uint32_t now;
   alarm_internal_read(&now);
-  alarm_at(now, interval, cb, ud, &timer->alarm);
+  return alarm_at(now, interval, cb, ud, &timer->alarm);
 }
 
 static void repeating_upcall( uint32_t now,
@@ -181,11 +175,17 @@ static void delay_upcall(__attribute__ ((unused)) int unused0,
   *((bool*)ud) = true;
 }
 
-void delay_ms(uint32_t ms) {
+int delay_ms(uint32_t ms) {
   bool cond = false;
   tock_timer_t timer;
-  timer_in(ms, delay_upcall, &cond, &timer);
+  int rc;
+
+  if ((rc = timer_in(ms, delay_upcall, &cond, &timer)) != RETURNCODE_SUCCESS) {
+    return rc;
+  }
+
   yield_for(&cond);
+  return rc;
 }
 
 static void yield_for_timeout_upcall(__attribute__ ((unused)) int unused0,
--- a/libtock/alarm_timer.c
+++ b/libtock/alarm_timer.c
@@ -1,8 +1,11 @@
 #include "alarm.h"
 #include "internal/alarm.h"
 #include "timer.h"
+
+#include <console.h>
 #include <limits.h>
 #include <stdlib.h>
+
 
 // Returns 0 if a <= b < c, 1 otherwise
 static int within_range(uint32_t a, uint32_t b, uint32_t c) {
@@ -71,7 +74,9 @@ static void callback( __attribute__ ((unused)) int unused0,
     // has the alarm not expired yet? (distance from `now` has to be larger or
     // equal to distance from current clock value.
     if (alarm->dt > now - alarm->reference) {
-      alarm_internal_set(alarm->reference, alarm->dt);
+      if (alarm_internal_set(alarm->reference, alarm->dt) != RETURNCODE_SUCCESS)
+        putnstr("Err: failed to set Alarm\n", sizeof("Err: failed to set Alarm"));
+
       break;
     } else {
       root_pop();
@@ -100,7 +105,9 @@ void alarm_at(uint32_t reference, uint32_t dt, subscribe_upcall cb, void* ud, al
 
   if (root_peek() == alarm) {
     alarm_internal_subscribe((subscribe_upcall*)callback, NULL);
-    alarm_internal_set(alarm->reference, alarm->dt);
+
+    if (alarm_internal_set(alarm->reference, alarm->dt) != RETURNCODE_SUCCESS)
+      putnstr("Err: failed to set Alarm\n", sizeof("Err: failed to set Alarm"));
   }
 }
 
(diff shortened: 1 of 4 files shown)


thulithwilfred:
@alistair23

alistair23 (approved):

hudson-ayers (commented):
Thanks for identifying this issue -- it is a problem that delay_ms() can silently fail. That said, I am not sure this is the right fix. First, I think that alarm_at() and timer_in() should just return an error rather than printing in the case of an error. There are plenty of use cases where failing to set an alarm might need to be handled by something other than a printed message. This would allow at least moving 2 of these prints up into delay_ms().

The possibility of a failure to set an alarm within the callback is more challenging -- I am not sure of a way to reasonably communicate that to a caller except for printing. On the other hand I do not love that this creates a dependency on the console driver in alarm_timer -- this makes it impossible to build even a simple blink app without paying the code size price of the console. For example, this change brings the .text size of a cortex-m blink app from 1460 bytes to 2452 bytes, and the .data size from 196 bytes to 228 bytes.

thulithwilfred:
>

thulithwilfred:
> Thanks for identifying this issue -- it is a problem that delay_ms() can silently hang apps. That said, I am not sure this is the right fix. First, I think that alarm_at() and timer_in() should just return an error rather than printing in the case of an error. There are plenty of use cases where failing to set an alarm might need to be handled by something other than a printed message. This would allow at least moving 2 of these prints up into delay_ms().
> 
> The possibility of a failure to set an alarm within the callback is more challenging -- I am not sure of a way to reasonably communicate that to a caller except for printing. On the other hand I do not love that this creates a dependency on the console driver in alarm_timer -- this makes it impossible to build even a simple blink app without paying the code size price of the console. For example, this change brings the .text size of a cortex-m blink app from 1460 bytes to 2452 bytes, and the .data size from 196 bytes to 228 bytes.

Thanks for all the great feedback. As you've suggested, I made changed for `alarm_at()` and `timer_in()` to propagate the error back to `delay_ms()`.

However, the dependency on the console in alarm_timer is still the issue, and I totally get the concern, it does significantly increase the footprint. In this case, can we justify this overhead vs silent hangs? Alternatively, could return error from `delay_ms()` and let the application handle it? Although typically you would expect it to just work.

libtock/alarm_timer.c
```diff
@@ -71,7 +74,9 @@ static void callback( __attribute__ ((unused)) int unused0,
     // has the alarm not expired yet? (distance from `now` has to be larger or
     // equal to distance from current clock value.
     if (alarm->dt > now - alarm->reference) {
-      alarm_internal_set(alarm->reference, alarm->dt);
+      if (alarm_internal_set(alarm->reference, alarm->dt) != RETURNCODE_SUCCESS)
+        putnstr("Err: failed to set Alarm\n", sizeof("Err: failed to set Alarm"));
+
```
brghena, line 79: Please add curly braces to your if statements. We try to avoid [goto fail](https://nakedsecurity.sophos.com/2014/02/24/anatomy-of-a-goto-fail-apples-ssl-bug-explained-plus-an-unofficial-patch/) by construction.

libtock/alarm_timer.c
```diff
@@ -100,8 +105,11 @@ void alarm_at(uint32_t reference, uint32_t dt, subscribe_upcall cb, void* ud, al
 
   if (root_peek() == alarm) {
     alarm_internal_subscribe((subscribe_upcall*)callback, NULL);
-    alarm_internal_set(alarm->reference, alarm->dt);
+
+    if (alarm_internal_set(alarm->reference, alarm->dt) != RETURNCODE_SUCCESS)
+      return RETURNCODE_FAIL;
```
brghena, line 110: Curly braces here too

libtock/alarm_timer.c
```diff
@@ -176,7 +184,10 @@ static void delay_upcall(__attribute__ ((unused)) int unused0,
 void delay_ms(uint32_t ms) {
   bool cond = false;
   tock_timer_t timer;
-  timer_in(ms, delay_upcall, &cond, &timer);
+
+  if (timer_in(ms, delay_upcall, &cond, &timer) != RETURNCODE_SUCCESS)
+    putnstr("Err: failed to set Alarm\n", sizeof("Err: failed to set Alarm"));
+
```
brghena, line 190: Curly braces here too.

libtock/alarm_timer.c
```diff
@@ -71,7 +74,9 @@ static void callback( __attribute__ ((unused)) int unused0,
     // has the alarm not expired yet? (distance from `now` has to be larger or
     // equal to distance from current clock value.
     if (alarm->dt > now - alarm->reference) {
-      alarm_internal_set(alarm->reference, alarm->dt);
+      if (alarm_internal_set(alarm->reference, alarm->dt) != RETURNCODE_SUCCESS)
+        putnstr("Err: failed to set Alarm\n", sizeof("Err: failed to set Alarm"));
```
brghena, line 78: Missed the "\n" in the `sizeof()`.

libtock/alarm_timer.c
```diff
@@ -176,7 +184,10 @@ static void delay_upcall(__attribute__ ((unused)) int unused0,
 void delay_ms(uint32_t ms) {
   bool cond = false;
   tock_timer_t timer;
-  timer_in(ms, delay_upcall, &cond, &timer);
+
+  if (timer_in(ms, delay_upcall, &cond, &timer) != RETURNCODE_SUCCESS)
+    putnstr("Err: failed to set Alarm\n", sizeof("Err: failed to set Alarm"));
```
brghena, line 189: Missed the `\n` in `sizeof()`.

brghena (changes requested):

brghena:
Since the problem only occurs in a rare error case, what if the alarm driver checks for the existence of the console driver (command 0) before attempting to print? Then if it's not installed, nothing prints, but we're not worse off than before.

hudson-ayers:
We talked about this on the call today, to summarize:

1. Looking at the alarm driver code in the kernel it seems that the only way for setting an alarm to return an error is if the call to grant::enter() fails, which should only happen if grant allocation fails, which should only happen on the first call into the alarm driver by a given application. Is this the failure that you observed?
2. If we are correct that we only expect setting an alarm to fail on the *first* call, then there is no reason to worry about failing to set an alarm from within the alarm callback -- if an alarm has fired, it should be impossible for a subsequent call to error. Given this, it seems that the right call is probably returning errors all the way up -- from alarm_at() into timer_in() into delay_ms(), even modifying the signature of delay_ms() to return errors. In the case of an error, delay_ms() would not yield, and instead return an error immediately. Most callers would ignore the error returned from delay_ms(), but at least those callers would not hang in the case of this failure. Apps that are especially worried about this particular failure case now *could* handle it if they so choose. Best of all, the timer driver does not become bloated with console code.

Can you confirm that a grant::enter() failure is what led to the actual errors you observed being returned? Is so, what do you think of this solution?

thulithwilfred:
> We talked about this on the call today, to summarize:
> 
>     1. Looking at the alarm driver code in the kernel it seems that the only way for setting an alarm to return an error is if the call to grant::enter() fails, which should only happen if grant allocation fails, which should only happen on the first call into the alarm driver by a given application. Is this the failure that you observed?
> 
>     2. If we are correct that we only expect setting an alarm to fail on the _first_ call, then there is no reason to worry about failing to set an alarm from within the alarm callback -- if an alarm has fired, it should be impossible for a subsequent call to error. Given this, it seems that the right call is probably returning errors all the way up -- from alarm_at() into timer_in() into delay_ms(), even modifying the signature of delay_ms() to return errors. In the case of an error, delay_ms() would not yield, and instead return an error immediately. Most callers would ignore the error returned from delay_ms(), but at least those callers would not hang in the case of this failure. Apps that are especially worried about this particular failure case now _could_ handle it if they so choose. Best of all, the timer driver does not become bloated with console code.
> 
> 
> Can you confirm that a grant::enter() failure is what led to the actual errors you observed being returned? Is so, what do you think of this solution?

When investigating this bug, what we first found was that the incorrect function was being used to handle the syscall return value (`tock_command_return_novalue_to_returncode`), I believe this always returned error (despite the alarm being set), so I had initially thought the alarm was not being set. Fixing this to the correct call, `tock_command_return_u32_to_returncode` it would return success. So it wasn't actually failing at a grant::enter(). The alarm was actually being set. 

I think it make sense to get rid of the prints, considering the overhead and all -- that sounds like a good solution to have the all the errors return all the way up, and make delay_ms() not yield in the case of error and immediately return an error code.

hudson-ayers:
> When investigating this bug, what we first found was that the incorrect function was being used to handle the syscall return value (`tock_command_return_novalue_to_returncode`), I believe this always returned error (despite the alarm being set), so I had initially thought the alarm was not being set. Fixing this to the correct call, `tock_command_return_u32_to_returncode` it would return success. So it wasn't actually failing at a grant::enter(). The alarm was actually being set.

That makes sense -- glad to hear we weren't actually seeing grant failures on the upstream board.

> I think it makes sense to get rid of the prints, considering the overhead and all -- that sounds like a good solution to have the all the errors return all the way up, and make delay_ms() not yield in the case of error and immediately return an error code.

Great, feel free to implement that here if you like! Or if you don't have time, feel free to pare this down to just the syscall return value fix, and then we can fixup the alarm error propagation in a followup PR.

thulithwilfred:
Got rid of the console driver and implemented `delay_ms()` to immediately return an error in the case an alarm is not set. 

Let me know if this is missing anything.

libtock/alarm_timer.c
```diff
@@ -4,6 +4,7 @@
 #include <limits.h>
 #include <stdlib.h>
 
+
```
alistair23, line 7: You don't need this newline

libtock/alarm_timer.c
```diff
@@ -100,8 +101,12 @@ void alarm_at(uint32_t reference, uint32_t dt, subscribe_upcall cb, void* ud, al
 
   if (root_peek() == alarm) {
     alarm_internal_subscribe((subscribe_upcall*)callback, NULL);
-    alarm_internal_set(alarm->reference, alarm->dt);
+
+    if (alarm_internal_set(alarm->reference, alarm->dt) != RETURNCODE_SUCCESS) {
+      return RETURNCODE_FAIL;
```
alistair23, line 106: Can't you just directly `return alarm_internal_set(...`?

thulithwilfred: Can do, just wondering in the case the syscall fails, this can return a status-code from the kernel (still < 0), from `tock_status_to_returncode()` to `tock_command_return_u32_to_returncode()` back to  `alarm_internal_set()` and out.  Not sure if there's a preference to how this should be handled?

hudson-ayers (approved):

alistair23 (approved):

libtock/alarm_timer.c
```diff
@@ -173,11 +175,16 @@ static void delay_upcall(__attribute__ ((unused)) int unused0,
...
 }
 
-void delay_ms(uint32_t ms) {
+int delay_ms(uint32_t ms) {
   bool cond = false;
   tock_timer_t timer;
-  timer_in(ms, delay_upcall, &cond, &timer);
+
+  if (timer_in(ms, delay_upcall, &cond, &timer) != RETURNCODE_SUCCESS) {
+    return RETURNCODE_FAIL;
```
brghena, line 183: You should probably return the real result from `timer_in()` rather than `RETURNCODE_FAIL` here. That would make this code more future-proof. It's not really an error as-is though.

brghena (approved):
This looks fine. I posted one possible fix that would improve it a little, but there's nothing wrong, as I can tell.

libtock/alarm_timer.c
thulithwilfred: Thanks, updated to return the actual return value of `timer_in()`.

brghena (approved):

hudson-ayers:
The CI failure should be fixed by https://github.com/tock/libtock-c/pull/275

thulithwilfred:
> The CI failure should be fixed by #275

Cheers!

brghena:
Merged. Thanks for the fix!!

PR #278, opened 2022-04-29
README: Added dependencies for Fedora

thulithwilfred:
**Pull Request Overview**
Adding Fedora deps and instructions required to cross compile
libtock-c apps for arm/risc V.

**Testing Strategy**
make RISCV=1 and make on Fedora 35

**Documentation Updated**
README updated

Signed-off-by: Wilfred Mallawa <thulith.mallawa@uqconnect.edu.au>

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,12 @@ Prerequisites
    $ sudo pacman -Syu arm-none-eabi-gcc
    ```
 
-1. Optional: libtock-c also includes support for building for RISC-V targets.
+   **Fedora**:
+   ```
+   $ sudo dnf install arm-none-eabi-newlib arm-none-eabi-gcc-cs
+   ```
+
+2. Optional: libtock-c also includes support for building for ***RISC-V targets***.
    These are not included by default since obtaining the toolchain can be
    difficult (as of June 2020). You will need a `riscv64-unknown-elf` GCC
    toolchain that supports rv32 targets as well (i.e. is compiled with multilib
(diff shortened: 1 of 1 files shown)


hudson-ayers (approved):
I have not tested this, but always helpful to have build instructions for additional distros, thank you for the contribution!

brghena (commented):
You probably want to split some of that into the next enumerated item below. This one is installation for `arm`, while the next is for `RISC-v`.

thulithwilfred:
> I have not tested this, but always helpful to have build instructions for additional distros, thank you for the contribution!

thulithwilfred:
> I have not tested this, but always helpful to have build instructions for additional distros, thank you for the contribution!

No worries!

thulithwilfred:
> You probably want to split some of that into the next enumerated item below. This one is installation for `arm`, while the next is for `RISC-v`.

Updated, let me know if that's not what you meant.

README.md
```diff
@@ -52,7 +52,35 @@ Prerequisites
...
+   ```
+   **Note: add /opt/riscv/bin to your PATH**, then,
+   ```
+   $ ./configure --prefix=/opt/riscv --enable-multilib
+   ```
+   `--enable-multilib` ensures that "the multilib compiler will have the prefix riscv64-unknown-elf- or riscv64-unknown-linux-gnu- but will be able to target both 32-bit and 64-bit systems."
+   ```
+   $ sudo make linux         [might need elevated privileges writing to `/opt/`]
+   ```
+    After the the source has been compiled and copied to `/opt/riscv` and `/opt/riscv/bin`has appended to the PATH, the toolchain is ready to be used.
```
brghena, line 81: Sorry! I was unclear. I meant to move all of this stuff into the next enumerated item below. You can take a look at the rendered view here: https://github.com/tock/libtock-c/blob/0614a609de1e9e19f4fa163aef5f1a553f656c7d/README.md which makes it more clear.

This item (3. in the rendered view) is instructions for all OSes for ARM.

The one below that starts "Optional: libtock-c also includes support for building for RISC-V targets." (4. in the rendered view) has instructions for all the OSes for RISC-V. So the Fedora instructions for RISC-V should be split from the ARM instructions and moved down there.

thulithwilfred: Gotcha, fixed. Thanks for that!

brghena (approved):

brghena:
Thanks so much!

PR #280, opened 2022-06-26
Adds support for App ID in build process

phil-levis:
This PR does not need to merged. It includes changes to the Makefiles such that libtock-c binaries use program headers and include App ID credentials. It allows you to test kernel support for App IDs.


ppannuto:
Closing here in favor of including in documentation in the book in the future ( https://github.com/tock/book/issues/56 ).

PR #281, opened 2022-07-02
OTA app project

BravoHoseok:
### Pull Request Overview
[2022-07-02] First prototype

[2022-07-23] Change Log
1) Uart version ota app is implemented.
2) At current implementation, ota app provides only loading a new app (not erase)
3) Applied feedbacks I received, and I made a couple of improvements in terms of functionality.
4) Added 'ota_uart.py' tools. With this tool, a new application can be updated via UART at runtime. Please note that the size of the new application has to be smaller or equal than the size of ota app because of MPU alignment rule. 
5) 'ota app' keep track the dynamically changing start address of flash and sram memory after loading applications by tockloader and ota app. By doing so, the update procedure by 'ota app' doesn't interfere with the memory region that is occupied by kernel and other apps.
6) I left nonvolatile_storage_driver at main.rs. To load an application with 'ota app', it is a necessary component. Also, the addresses of kernel and app flash memory are independent on the size of 'ota app'. It means that the addresses are the fixed specification of microbit_v2 platform. We don't need to change the addresses according to the size of 'ota app'

[2022-08-02] Change Log
1) Added a new feature that finds a start address of flash satisfying MPU rules. Now we can load 3 applications by OTA app. For simplicity, I didn't consider MPU subregion rules.
 2) Since tockloader adds 512 bytes of 01 padding from the end of an app, tockloader should not be used together with OTA app after loading an app with OTA app. I also deleted writing that padding bytes from 'ota_uart.py' tool
3) Whether there is enough flash region satisfying MPU rules and index to save a new app is transferred into `process_load_utilities'
4) After finding a start address based on MPU rules, we check whether or not the new region for new app invades other regions already occupied by other apps as fail-safety.
5) Please refer to [2022-08-02] section of 'OTA_app_system_documnet.md'

[2022-08-11] Change Log
1) Added a new function to insert padding apps. So, the original process_load_advanced function can successfully load the loaded apps from OTA app
2) I also check CRC32 consistency of padding apps
3) Added a few commands

[2022-08-14] Change Log
1) Solved two issues caused by phantom apps after `tockloader erase-apps`
2) Fixed bug in check_overlap_region function
3) Organize code

[2022-08-15] Change Log
1) Added validation check of TBF base header
- The header length isn't greater than the entire app
- The header length is at least as large as the v2 required header (which is 16 bytes)
- Check Base Header Checksum consistency
- Check consistency between the requested app size and the actual app size in TBF header

2) Added a security feature.
- Attack Scenario: A malicious ota app is installed via OTA app, and it deletes  (0xff) all of the flash region  by using `nonvolatile_storage_driver`. 
- Result: Although the malicious ota app manipulate the regions unoccupied by the existing apps, it doesn't have to invade the other regions occupied by the existing apps.

### Testing Strategy
[2022-07-02] First prototype. It is not necessary to test.

[2022-07-23] Change Log
1) For demo, please refer to the guide section of 'OTA_app_system_documnet.md'.
2) Test cases in terms of functionalities.
  - ota app + 3 apps (by ota app) -> tockloader erase-apps -> load ota app again -> load an apps (by ota app) : Success
  - ota app + 2 apps (by ota app) -> push reset button -> load 1 app: Success
  - A big size app + ota app (positioned at index 1) -> load an app (by ota app) : Success
  - ota app + 1 app (by tockloader) -> load a new app (by ota app): Success
  - ota app -> load an app with crc fail -> erase the loaded app: Success
  - ota all -> load a big size app which doesn't follow MPU alighnemt rule -> erase the loaded app and do not load the entry point of the app: Success
  - Checked the dynamically changing flash and sram start address by printing out that values

[2022-08-02] Change Log
1) I'm testing about 800 combination of app bundles (256k - 512 byte) loaded by OTA app. As soon as I finish the test, I will load the result [2022-08-04 Done and Pass]

[2022-08-11] Change Log
1) After loading apps, check the app bundles with `tockloader list --verbose`. If the app bundles are loaded successfully, tockloader can read the loaded apps successfully. [2022-08-11 Done and Pass]
2) Please refer to `Alignment_Test.xlsx` in /doc/OTA_app

[2022-08-14] Change Log
1) Tested two issues caused by phantom apps after `tockloader erase-apps`

[2022-08-15] Change Log
1) Tested the above attack scenario
2) Result: No corruption
3) Tested TBF base header validation check

### TODO or Help Wanted
[2022-07-02] First prototype.
[2022-07-23]
1) Need to modify make file and elf2tab to add permission header to ota_app.tab for security [2022-08-02 Solved]
2) Need to come up with an idea satisfying the MPU alignment rule, when loading application by ota app [2022-08-02 Solved]
3) Try to erase and update function as future work

[2022-08-02]
1) After loading apps with OTA app, when I push the reset button, the original 'load_process_advanced' cannot load the sparsely located apps. I need to find a way to parse the apps. [2022-08-11 Solved]

[2022-08-11]
1) Is it worthy to improve finding a start address based on MPU rules including subregion?
(Currently, I didn't consider subregion rules) => Not priority for now.

2) If we load a new app, but another app that has same name as the new app, I think, it is reasonable to erase the old app and flash the new app again. However, we can't assure that the new app uses the same size of sram and flash memory as the old app. How to deal with this situation? Although they have same flash size, they can use different size of sram.

[2022-08-14]
1) Add security feature to prevent a malicious OTA app from manipulating the flash region occupied by the existing apps

### Documentation Updated

- [x] Updated the relevant files in `/docs/OTA_app`, Please refer to 'OTA_app_system_documnet.md' at docs directory. I summarized the OTA app design concept.

### Formatting

- [] Ran `make prepush`.


ppannuto:
Closed due to inactivity without prejudice.

PR #283, opened 2022-07-07
readme: July 2022 update and formatting

bradjc:
This fixes some formatting issues, and removes the note about brew on mac compiling the riscv compiler as the multilib version is now standard.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,14 +78,16 @@ Prerequisites
    $ brew tap riscv/riscv && brew update && brew install riscv-gnu-toolchain
    ```
 
-   **Ubuntu (19.10 or later)**:
+   **Ubuntu (21.10 or later)**:
    ```
-   $ sudo apt install gcc-riscv64-unknown-elf
+   $ sudo apt install gcc-riscv64-unknown-elf picolibc-riscv64-unknown-elf
    ```
 
-   Unfortunately, Ubuntu does not currently (June 2022) provide a package for
-   RISC-V libc. We have created a .deb file you can use to install a suitable
-   libc based on newlib:
+   **Ubuntu (21.04 or earlier)**:
+
+   Unfortunately, older Ubuntu does not provide a package for RISC-V libc. We
+   have created a .deb file you can use to install a suitable libc based on
+   newlib:
    ```
    $ wget http://cs.virginia.edu/~bjc8c/archive/newlib_3.3.0-1_amd64.deb
    $ sudo dpkg -i newlib_3.3.0-1_amd64.deb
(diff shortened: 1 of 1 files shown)


hudson-ayers (approved):

README.md
```diff
@@ -69,40 +69,37 @@ Prerequisites
...
-   build a bleeding-edge version of GCC, and there is a small chance it will not
-   work. However, we have successfully obtained a toolchain this way.
 
    **Ubuntu (19.10 or later)**:
    ```
    $ sudo apt install gcc-riscv64-unknown-elf
    ```
 
-   Unfortunately, Ubuntu does not currently (June 2020) provide a package for
+   Unfortunately, Ubuntu does not currently (June 2022) provide a package for
```
alistair23, line 80: I don't think this is true. Picolib seems to be packaged for Ubuntu now

alistair23 (changes requested):

README.md
```diff
@@ -57,9 +57,9 @@ Prerequisites
...
    ```
 
-2. Optional: libtock-c also includes support for building for ***RISC-V targets***.
-   These are not included by default since obtaining the toolchain can be
-   difficult (as of June 2020). You will need a `riscv64-unknown-elf` GCC
+2. Optional: libtock-c also includes support for building for ***RISC-V
+   targets***. These are not included by default since obtaining the toolchain
+   can be difficult (as of July 2022). You will need a `riscv64-unknown-elf` GCC
    toolchain that supports rv32 targets as well (i.e. is compiled with multilib
    support).
```
lschuermann, line 64: I think that's not generally true. All RISC-V compilers are able to compile for `rv32i/ilp32` platforms, it's just the compiler intrinsics shipped as part of `libgcc` which can cause issues (which realistically I found to be an issue only for targets without hardware multiplication support, i.e. `rv32i`, see NixOS/nixpkgs#178140). Also we don't require a `riscv64-` toolchain, `riscv32-` and other tuples can work just fine.

So I agree with it being hard to acquire the correct toolchain, but perhaps we'd want to rephrase saying that we need a RISC-V toolchain able to target the respective chip's architecture?

README.md
bradjc: https://packages.ubuntu.com/jammy/picolibc-riscv64-unknown-elf ?

alistair23: Yeah!

bradjc: Added picolib.

hudson-ayers (approved):
Great to see a little RISC-V toolchain progress

hudson-ayers:
It looks like just installing picolibc is not sufficient for the libtock-c Makefile to actually use it, unfortunately -- just get missing header errors

PR #285, opened 2022-07-13
Configuration.mk: change litex_sim target definition to rv32imc

### LINKED ISSUE #282: RISC-V Build Requires libstdc++/libsupc++
Trying to build for RISC-V but I keep getting this error at link time:

```
/usr/lib/riscv64-unknown-elf/bin/ld: cannot find -lstdc++
/usr/lib/riscv64-unknown-elf/bin/ld: cannot find -lsupc++
```

In other words, my build environment does not contain copies of libstdc++ or libsupc++ for RISC-V.

I followed the instructions to install newlib using the prebuilt .deb package. Is there perhaps another similar step I can follow to get a copy of these C++ libs?

Also interesting: I removed the flags from [this line](https://github.com/tock/libtock-c/blob/14b0118654ddc09d55b704c825c5d685261b38b1/Configuration.mk#L250-L251) in Configuration.mk and my build was successful. Perhaps these libs aren't needed for a simple C application?

@bradjc it looks like you're heavily involved with RISC-V support. Is this perhaps something you have context on?

lschuermann:
This is in response to issue tock/libtock-c#282, which exposed that `rv32i` support is not prevalent among prebuilt RISC-V embedded toolchains. Users of such toolchains are faced with linker errors when trying to build a `rv32i` target, which is included by default in all RISC-V builds. As of tock/tock#2860 the `litex_sim` board is actually using a VexRiscv CPU with multiplication and compressed instruction support and hence this target definition can be changed to target `-march=rv32imc` instead. Yet, it does not solve the underlying toolchain issue. Refer to [1] for more information.

Signed-off-by: Leon Schuermann <leon@is.currently.online>

[1]: https://github.com/tock/libtock-c/issues/282

Closes #282.

### DIFF
--- a/Configuration.mk
+++ b/Configuration.mk
@@ -69,7 +69,7 @@ ARTY_E21_TOCK_TARGETS := rv32imac|rv32imac.0x40430060.0x80004000|0x40430060|0x80
 #  rv32imac|rv32imac.0x20040060.0x80002800 # RISC-V for HiFive1b
 #  rv32imac|rv32imac.0x403B0060.0x3FCC0000 # RISC-V for ESP32-C3
 #  rv32imc|rv32imc.0x41000060.0x42008000   # RISC-V for LiteX Arty-A7
-#  rv32i|rv32i.0x00080060.0x40008000       # RISC-V for LiteX Simulator
+#  rv32imc|rv32imc.0x00080060.0x40008000   # RISC-V for LiteX Simulator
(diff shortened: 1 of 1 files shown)


alistair23 (approved):

bradjc:
Also update the comment

ppannuto (approved):
bors r+

PR #288, opened 2022-07-22
analog_comparator: dont marked used callback arguments as unused

hudson-ayers:
These callback arguments were incorrectly marked as unused.


ppannuto (approved):
bors r+

PR #291, opened 2022-08-26
Fix number of bits used for voltage

alexandruradovici:
The PR fixes the number of bits used to determine the voltage. The trait specifies that all values should be u16, the example was using only 12 bits.


bradjc (approved):

brghena (changes requested):
This isn't a correct change, unfortunately.

We return ADC value in a 16-bit object, per the trait. However, the formula here is based on the resolution and reference voltage of the ADC.

You're correct that the formula assumes 12-bit resolution and 3.3 v reference, which is wrong. Instead, we should be querying the ADC for those values and then applying the formula. This is really a fix that needs to apply in `libtock` for the ADC driver to add the commands to [get resolution bits](https://github.com/tock/tock/blob/master/capsules/src/adc.rs#L1257) and to [get voltage reference millivolts](https://github.com/tock/tock/blob/master/capsules/src/adc.rs#L1266). They won't change, so an application like this can get them once in `main()` at setup time and then just use that information later when interpreting samples.

ppannuto:
@brghena should we just add a convenience method `adc::get_milivolts` to `libtock` that does this correctly?

alexandruradovici:
Let me look into it and try to fix it.

alexandruradovici:
> You're correct that the formula assumes 12-bit resolution and 3.3 v reference, which is wrong. Instead, we should be querying the ADC for those values and then applying the formula. 

Agree for the voltage, but why should we query the number of bits if we always have 16 bits (as per trait)?

alexandruradovici:
I added the reference voltage and resolution functions and used them in the example.

ppannuto (approved):

examples/tests/adc/main.c
```diff
@@ -21,7 +24,7 @@ static void test_single_samples(uint8_t channel) {
   } else {
     // 12 bit, reference = VCC/2, gain = 0.5
     // millivolts = ((sample * 2) / (2^12 - 1)) * (3.3 V / 2) * 1000
-    int millivolts = (sample * 3300) / 4095;
+    int millivolts = ((sample >> (16 - resolution)) * reference_voltage) / ((1 << resolution) - 1);
```
brghena, line 27: The equation should just be `sample * reference_voltage / ((1 << resolution) -1)` right? https://learn.sparkfun.com/tutorials/analog-to-digital-conversion/relating-adc-value-to-voltage

I'm not sure where the `>> (16 - resolution)` comes from.

examples/tests/adc/main.c
```diff
@@ -40,7 +43,7 @@ static void test_sampling_buffer(uint8_t channel, int index) {
     printf("\t[ ");
     for (uint32_t i = 0; i < length; i++) {
       // convert to millivolts
-      printf("%u ", (buf[i] * 3300) / 4095);
+      printf("%u ", ((buf[i] >> (16 - resolution)) * reference_voltage) / ((1 << resolution) - 1));
```
brghena, line 46: Same here.

brghena (changes requested):

brghena:
@alexandruradovici Unless I'm misunderstanding something, the trait specifies that the resolution of the samples fits within 16 bits. It doesn't specify that the resolution _is_ 16 bits. So if a chip has an 8-bit ADC, it will still have 8 bits of sample resolution, even if that data is placed in a `u16`.

examples/tests/adc/main.c
brghena: Okay, so given that all samples essentially are 16-bit resolution (given that they are left justified) removing the `resolution` part of this altogether seems like the solution. Sorry for the confusion on my part!

alexandruradovici: I did not use the resolution initially, I'll revert it.

alexandruradovici: I reverted it.

examples/tests/adc/main.c
```diff
@@ -21,7 +23,7 @@ static void test_single_samples(uint8_t channel) {
   } else {
     // 12 bit, reference = VCC/2, gain = 0.5
     // millivolts = ((sample * 2) / (2^12 - 1)) * (3.3 V / 2) * 1000
```
brghena, line 25:
```suggestion
    // All ADC results are left-aligned, resulting in a 16-bit resolution
    // millivolts = sample * reference_voltage_mv / resolution
```

brghena (changes requested):
Sorry, one more comment to fix-up. I can't push to your remote repo or I would do it myself.

brghena (approved):

PR #292, opened 2022-08-26
Fixes the screen API to use the improved driver

alexandruradovici:
The PR fixes the screen API to use the new screen driver that uses synchronous system call for getting the resolution and pixel mode.

### DIFF
--- a/libtock/screen.c
+++ b/libtock/screen.c
@@ -34,75 +34,53 @@ static int screen_allow (const void* ptr, size_t size) {
 }
 
 int screen_get_supported_resolutions (int* resolutions) {
-  ScreenReturn fbr;
-  fbr.data1 = 0;
-  fbr.done  = false;
-
-  int ret = screen_subscribe(screen_callback, &fbr);
-  if (ret == RETURNCODE_SUCCESS) {
-    syscall_return_t com = command(DRIVER_NUM_SCREEN, 11, 0, 0);
-    ret = tock_command_return_novalue_to_returncode(com);
-    if (ret == RETURNCODE_SUCCESS) {
-      yield_for (&fbr.done);
-      *resolutions = fbr.data1;
-      ret = fbr.error;
-    }
+  syscall_return_t com = command(DRIVER_NUM_SCREEN, 11, 0, 0);
+  if (com.type == TOCK_SYSCALL_SUCCESS_U32) {
+    *resolutions = com.data[0];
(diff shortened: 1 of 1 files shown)


alexandruradovici:
The formatter does not seem to work on my M1 mac, it would help if anyone could format this.

ppannuto:
I can't push to that branch, but this is what's needed:

```
diff --git a/libtock/screen.c b/libtock/screen.c
index adbb202..4c2b70d 100644
--- a/libtock/screen.c
+++ b/libtock/screen.c
@@ -38,8 +38,7 @@ int screen_get_supported_resolutions (int* resolutions) {
   if (com.type == TOCK_SYSCALL_SUCCESS_U32) {
     *resolutions = com.data[0];
     return RETURNCODE_SUCCESS;
-  } else if (com.type == TOCK_SYSCALL_FAILURE)
-  {
+  } else if (com.type == TOCK_SYSCALL_FAILURE) {
     return tock_status_to_returncode(com.data[0]);
   } else {
     return RETURNCODE_EINVAL;
@@ -49,7 +48,7 @@ int screen_get_supported_resolutions (int* resolutions) {
 int screen_get_supported_resolution (size_t index, size_t *width, size_t *height) {
   syscall_return_t com = command(DRIVER_NUM_SCREEN, 12, index, 0);
   if (com.type == TOCK_SYSCALL_SUCCESS_U32_U32) {
-    *width = com.data[0];
+    *width  = com.data[0];
     *height = com.data[1];
     return RETURNCODE_SUCCESS;
   } else if (com.type == TOCK_SYSCALL_FAILURE) {
@@ -161,7 +160,7 @@ uint8_t * screen_buffer (void)
 int screen_get_resolution (size_t *width, size_t *height) {
   syscall_return_t com = command(DRIVER_NUM_SCREEN, 23, 0, 0);
   if (com.type == TOCK_SYSCALL_SUCCESS_U32_U32) {
-    *width = com.data[0];
+    *width  = com.data[0];
     *height = com.data[1];
     return RETURNCODE_SUCCESS;
   } else if (com.type == TOCK_SYSCALL_FAILURE) {
```

alexandruradovici:
Uncrustify 0.75.1 seems to work.

bradjc (approved):

bradjc:
So what you are saying is that having a tool which appended a format commit on this PR would have helped :)

PR #293, opened 2022-08-26
bump uncrustify version

ppannuto:
This doesn't look to have any effect on our formatting, but might
hopefully fix M1 mac issues?

@alexandruradovici Does the newest version run on M1 mac?

(uncrustify is also available in homebrew, and the script will pick up your local copy if it's available).

### DIFF
--- a/examples/ble-uart/ble_nus.c
+++ b/examples/ble-uart/ble_nus.c
@@ -159,10 +159,10 @@ static uint32_t tx_char_add(ble_nus_t * p_nus, const ble_nus_init_t * p_nus_init
   char_md.char_props.write         = 1;
   char_md.char_props.write_wo_resp = 1;
   char_md.p_char_user_desc         = NULL;
-  char_md.p_char_pf                = NULL;
-  char_md.p_user_desc_md           = NULL;
-  char_md.p_cccd_md                = NULL;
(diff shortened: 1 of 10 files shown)


hudson-ayers (approved):
I don't have my m1 mac at school, but confirmed that this works fine on Ubuntu 22.04

ppannuto:
Spoke too soon; the format-check cache doesn't check uncrustify version, understandably. There are a handful of small formatting changes, but I don't think the churn is too bad to merit not updating.

hudson-ayers (approved):

brghena (approved):

alexandruradovici:
> This doesn't look to have any effect on our formatting, but might hopefully fix M1 mac issues?
> 
> @alexandruradovici Does the newest version run on M1 mac?
>
It seems to work, but when I use `make format` it modifies `buzzer.c` and `ieee802154.c`.
> (uncrustify is also available in homebrew, and the script will pick up your local copy if it's available).

ppannuto:
Additional changes beyond those in fc401d3? That would suggest some non-determinism in the tool, which would be... sad.

Let's land this after the release and then see if there are still issues on a clean tree?

bradjc (approved):

bradjc:
Can we merge this?

PR #294, opened 2022-08-29
fix SuccessU32U64 syscall return variant name to match TRD104

hudson-ayers:
Matches updates in https://github.com/tock/tock/pull/3175


ppannuto (approved):
bors r+

Mercifully, nothing is actually using this yet, so an easy fix :)

PR #298, opened 2022-09-02
Update Uncrustify Options 2022

bradjc:
I don't think our current uncrustify options are opinionated enough, as the same code can be formatted multiple ways (particularly with spaces).

I added a bunch of uncrustify [flags](https://raw.githubusercontent.com/uncrustify/uncrustify/master/documentation/htdocs/ben.cfg.txt) I think we should use. Thoughts?


ppannuto:
I'm pro the tool being opinionated; mildly surprised / annoyed it doesn't just enable and set defaults for everything, but oh well.

The only style thing I don't like in that diff is the weird space-after-if-but-not-after-else:
```c
if (xxx) {
  ..
} else{
 // y no space??
}
```

This should also probably be on top of #293.

alexandruradovici:
I think this is a bug, running the tool several times yield different spacing there.

brghena:
For that `else` issue, the two you're looking for are `sp_else_brace` (force) and `sp_brace_else` (force). (Found here: http://uncrustify.sourceforge.net/config.txt)

hudson-ayers (approved):

brghena (approved):

bradjc:
Updated.

hudson-ayers (approved):

ppannuto (approved):

hudson-ayers (approved):
bors r+

Talked to Pat, sounds like this was a format caching issue that only applies because we updated uncrustify

PR #299, opened 2022-09-02
i2c_master: fix len in syscalls

### LINKED ISSUE #260: Mismatch in i2c_master syscall
There seems to be a mismatch between the i2c_master_write_read code in libtock-c and the capsule code in Tock OS.
The len is prepended to the address by shifting left 16 bits in libtock-c.
The write_len is recovered in Tock by shifting right 8 bits. 

```
tock/libtock-c/libtock/i2c_master.c

int i2c_master_write_read(uint8_t address, uint8_t len) {
uint32_t a = (((uint32_t) len) << 16) | address;
```

```
tock/tock/capsules/src/i2c_master.rs

Cmd::WriteRead => {
                    let addr = arg1 as u8;
                    let write_len = arg1 >> 8; // can extend to 24 bit write length
                    let read_len = arg2; // can extend to 32 bit read length
```

bradjc:
To match https://github.com/tock/tock/blob/master/capsules/src/i2c_master.rs#L130

Fixes #260.

### DIFF
--- a/libtock/i2c_master.c
+++ b/libtock/i2c_master.c
@@ -16,20 +16,19 @@ int i2c_master_set_buffer(uint8_t* buffer, uint32_t len) {
   return tock_allow_rw_return_to_returncode(aval);
 }
 
-int i2c_master_write(uint8_t address, uint8_t len) {
-  uint32_t a = (((uint32_t) len) << 16) | address;
-  syscall_return_t cval = command(DRIVER_NUM_I2CMASTER, 1, a, len);
+int i2c_master_write(uint8_t address, uint16_t len) {
+  syscall_return_t cval = command(DRIVER_NUM_I2CMASTER, 1, address, len);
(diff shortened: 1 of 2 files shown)


hudson-ayers (approved):
Looks good, sorta strange how we only shift the `write_len`, and only for the `write_read` call and not others.

lschuermann:
> Looks good, sorta strange how we only shift the `write_len`, and only for the `write_read` call and not others.

Indeed, but I suppose that this is the best we can do. Encoding both lengths in the same register will limit that to 16 bit, which might not be sufficient for some usecases (although 64k is a lot to read/write in a single operation). And shifting lengths in other calls would incur unnecessary overhead.

lschuermann (approved):

ppannuto (approved):
@AnthonyQ619, is the is same I2C issue you encountered a little while back?

PR #300, opened 2022-09-08
Configuration.mk: automatically generate TOCK_ARCH_FAMILIES

lschuermann:
This automatically generates `TOCK_ARCH_FAMILIES` from the `TOCK_TARGETS`. It allows users to only specify the list of target binaries generated, including fixed flash and RAM addresses for RISC-V targets. The Makefile infrastructure will automatically determine which toolchain steps to use throughout the build process (e.g. in the size calculation phase).


lschuermann:
> This is great. Would you mind also deleting the explicit set in lua hello?

Sure!

bradjc (approved):

PR #308, opened 2022-10-22
created a Micro:bit v2 test app, similar to the Hail test app

qpalzmz112:
This app tests the supported libtock-c libraries on the Microbit. It has limited usefulness without using tockloader listen, but some tests use the LED matrix to display results.

### DIFF
--- a/examples/tests/microbit_v2/Makefile
+++ b/examples/tests/microbit_v2/Makefile
@@ -13,4 +13,3 @@ EXTERN_LIBS += $(TOCK_USERLAND_BASE_DIR)/simple-ble
 # Include userland master makefile. Contains rules and flags for actually
 # building the application
 include $(TOCK_USERLAND_BASE_DIR)/AppMakefile.mk
-include $(TOCK_USERLAND_BASE_DIR)/simple-ble/Makefile
--- a/examples/tests/microbit_v2/Makefile
+++ b/examples/tests/microbit_v2/Makefile
@@ -3,8 +3,10 @@
 # Specify this directory relative to the current application
 TOCK_USERLAND_BASE_DIR = ../../..
 
+include $(TOCK_USERLAND_BASE_DIR)/simple-ble/Makefile
 # Which files to compile
 C_SRCS := $(wildcard *.c)
 # Include userland master makefile. Contains rules and flags for actually
 # building the application
 include $(TOCK_USERLAND_BASE_DIR)/AppMakefile.mk
+EXTERN_LIBS += $(TOCK_USERLAND_BASE_DIR)/simple-ble
--- a/examples/tests/microbit_v2/Makefile
+++ b/examples/tests/microbit_v2/Makefile
@@ -5,11 +5,6 @@ TOCK_USERLAND_BASE_DIR = ../../..
 
 # Which files to compile
 C_SRCS := $(wildcard *.c)
-
-#APP_HEAP_SIZE := 2048
-
-# External libraries used
-EXTERN_LIBS += $(TOCK_USERLAND_BASE_DIR)/simple-ble
 # Include userland master makefile. Contains rules and flags for actually
 # building the application
 include $(TOCK_USERLAND_BASE_DIR)/AppMakefile.mk
--- a/examples/tests/microbit_v2/Makefile
+++ b/examples/tests/microbit_v2/Makefile
@@ -1,4 +1,4 @@
-# Makefile for user application - copied from bradjc's Hail test app
+# Makefile for user application
 
 # Specify this directory relative to the current application
 TOCK_USERLAND_BASE_DIR = ../../..
--- a/examples/tests/microbit_v2/Makefile
+++ b/examples/tests/microbit_v2/Makefile
@@ -1,0 +1,17 @@
+# Makefile for user application - copied from bradjc's Hail test app
+
+# Specify this directory relative to the current application
+TOCK_USERLAND_BASE_DIR = ../../..
+
+# Which files to compile
+C_SRCS := $(wildcard *.c)
+
+APP_HEAP_SIZE := 2048
+
+# External libraries used
+EXTERN_LIBS += $(TOCK_USERLAND_BASE_DIR)/libnrfserialization
+EXTERN_LIBS += $(TOCK_USERLAND_BASE_DIR)/simple-ble
+# Include userland master makefile. Contains rules and flags for actually
+# building the application
+include $(TOCK_USERLAND_BASE_DIR)/AppMakefile.mk
+include $(TOCK_USERLAND_BASE_DIR)/simple-ble/Makefile
--- a/examples/tests/microbit_v2/Makefile
+++ b/examples/tests/microbit_v2/Makefile
@@ -6,10 +6,9 @@ TOCK_USERLAND_BASE_DIR = ../../..
 # Which files to compile
 C_SRCS := $(wildcard *.c)
 
-APP_HEAP_SIZE := 2048
+#APP_HEAP_SIZE := 2048
 
 # External libraries used
-EXTERN_LIBS += $(TOCK_USERLAND_BASE_DIR)/libnrfserialization
 EXTERN_LIBS += $(TOCK_USERLAND_BASE_DIR)/simple-ble
 # Include userland master makefile. Contains rules and flags for actually
 # building the application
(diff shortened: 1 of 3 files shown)


brghena:
We talked about this and qpalzmz112 knows the next steps to fix the format and build CI issues. Ping me when they're fixed and I'll do a full review of the code.

qpalzmz112:
@brghena I made the changes we talked about!

examples/tests/microbit_v2/Makefile
```diff
@@ -0,0 +1,16 @@
+# Makefile for user application - copied from bradjc's Hail test app
```
brghena, line 1:
```suggestion
# Makefile for user application
```

examples/tests/microbit_v2/Makefile
```diff
@@ -0,0 +1,16 @@
...
+C_SRCS := $(wildcard *.c)
+
+#APP_HEAP_SIZE := 2048
+
+# External libraries used
+EXTERN_LIBS += $(TOCK_USERLAND_BASE_DIR)/simple-ble
+# Include userland master makefile. Contains rules and flags for actually
+# building the application
+include $(TOCK_USERLAND_BASE_DIR)/AppMakefile.mk
+include $(TOCK_USERLAND_BASE_DIR)/simple-ble/Makefile
```
brghena, line 16:
```suggestion
```

examples/tests/microbit_v2/Makefile
```diff
@@ -0,0 +1,16 @@
...
+# Specify this directory relative to the current application
+TOCK_USERLAND_BASE_DIR = ../../..
+
+# Which files to compile
+C_SRCS := $(wildcard *.c)
+
+#APP_HEAP_SIZE := 2048
+
+# External libraries used
+EXTERN_LIBS += $(TOCK_USERLAND_BASE_DIR)/simple-ble
```
brghena, line 12:
```suggestion

```

brghena: I don't believe we're using `simple-ble` anymore, are we?

examples/tests/microbit_v2/README.md
```diff
@@ -0,0 +1,59 @@
+Micro:bit v2 Test App
```
brghena, line 1: This entire README is great! Very helpful.

examples/tests/microbit_v2/main.c
```diff
@@ -0,0 +1,159 @@
...
+#include <crc.h>
+#include <gap.h>
+#include <gpio.h>
+#include <led.h>
+#include <ninedof.h>
+#include <nrf51_serialization.h>
+#include <sound_pressure.h>
+#include <temperature.h>
+#include <timer.h>
+#include <tock.h>
```
brghena, line 20: Double-check through here on which of these includes we actually use. For instance "nrf51_serialization.h" is definitely not used.

examples/tests/microbit_v2/main.c
```diff
@@ -0,0 +1,159 @@
...
+                            __attribute__ ((unused)) int   arg2,
+                            __attribute__ ((unused)) void *ud) {
+  if (!val) { // button was released, not pressed
+    return;
+  }
+  int note0 = NOTE_C4;
+  int note1 = NOTE_E4;
+  int note2 = NOTE_G4;
+  if (btn_num == 0) {
+    check_err(tone(note0 * 3, 500, tone_callback), "tone");
```
brghena, line 54: Why the "* 3" here?

examples/tests/microbit_v2/main.c
```diff
@@ -0,0 +1,159 @@
...
+#include <led.h>
+#include <ninedof.h>
+#include <nrf51_serialization.h>
+#include <sound_pressure.h>
+#include <temperature.h>
+#include <timer.h>
+#include <tock.h>
+
+//          temperature works, but can't print floats
+//          buttons work -- in callback, including a call to tone_sync prevents code from resuming execution after callback
```
brghena, line 23:
```suggestion
```

examples/tests/microbit_v2/Makefile
brghena: Or maybe we are for "gap.h"?

examples/tests/microbit_v2/README.md
qpalzmz112: Thanks!

examples/tests/microbit_v2/Makefile
qpalzmz112: Yep. I just double-checked, and the simple-ble lines are needed for gap.h

examples/tests/microbit_v2/main.c
qpalzmz112: You're right, I missed some includes when I was removing the unneeded ones.

examples/tests/microbit_v2/main.c
brghena: For the `* 3`, I guess we can ignore it for now.

For the 3 buttons at once, I added a variable for tracking whether we are in a playing state or not. The problem is that queuing up more than a single note is not possible, as there is only memory for one queued note. Now notes will never be queued at all, which solves the issue.

brghena (approved):
I pushed a small cleanup. I'm now happy with this app!

examples/tests/microbit_v2/main.c
```diff
@@ -0,0 +1,163 @@
...
+  check_err(gpio_read(8, &p8), "gpio_read");
+  int p9 = 0;
+  check_err(gpio_read(9, &p9), "gpio_read");
+  int p16 = 0;
+  check_err(gpio_read(16, &p16), "gpio_read");
+
+  // print results
+  printf("[Micro:bit Sensor Reading]\n");
+  // Temp is given in hundredths of a degree C, in format XX00
+  printf("  Temperature: reading: %d.0 degrees C\n", temp / 100);
```
brghena, line 109: When you run the app, do you get a temperature that varies or changes when you put a finger on the microcontroller?

Mine always seems to read 24.0 degrees, which is weird. I can't see how that would be the fault of this application.

brghena:
@alexandruradovici Any opinions on what you want to see in a Microbit test app? This looks pretty great to me.

examples/tests/microbit_v2/main.c
brghena: Looks like I just wasn't patient enough yesterday. Temperature works fine!

hudson-ayers (approved):

examples/tests/microbit_v2/main.c
```diff
@@ -0,0 +1,163 @@
...
+}
+
+static void button_callback(int                            btn_num,
+                            int                            button_pressed,
+                            __attribute__ ((unused)) int   arg2,
+                            __attribute__ ((unused)) void *ud) {
+
+  // Do not run if this was a button release or if tone is not finished
+  if (!button_pressed || !tone_complete) {
+    return;
```
alexandruradovici, line 49:
```suggestion
    if (!tone_complete) printf("Please wait for the tone to stop\n");
    return;

```

examples/tests/microbit_v2/main.c
```diff
@@ -0,0 +1,163 @@
...
+  tone_complete = false;
+  if (btn_num == 0) {
+    check_err(tone(NOTE_C4 * 3, 500, tone_callback), "tone");
+    int leds[10] = {2, 6, 8, 11, 12, 13, 16, 18, 21, 23};
+    light_leds(leds, 10);
+  } else if (btn_num == 1) {
+    check_err(tone(NOTE_E4 * 3, 500, tone_callback), "tone");
+    int leds[11] = {1, 2, 6, 8, 11, 12, 13, 16, 18, 21, 22};
+    light_leds(leds, 11);
+  } else if (btn_num == 2) {
```
alexandruradovici, line 62: Is button 3 the logo? This might not work as expected on every board as the touch logo is not really a button.

alexandruradovici (approved):
Looks really good to me, I have just some small suggestions.

examples/tests/microbit_v2/main.c
brghena: Until just now, I had assumed that was implemented through some kind of capacitive touch indirection layer. But it doesn't seem like it is: https://github.com/tock/tock/blob/master/boards/microbit_v2/src/main.rs#L261

Amusingly it did work on the board I tested this with. Although I agree it probably shouldn't be relied on.

brghena (approved):

PR #310, opened 2022-11-22
remove alarm_read() from alarm.h

bradjc:
As far as I can tell the `alarm_read()` function does not exist.


hudson-ayers (approved):
Yeah, the function doesn't exist so we should remove the definition. But ideally we would actually implement this, since it can certainly be useful and there isn't a great alternative right now, as I understand it.

alistair23 (approved):

bradjc:
Can we merge this?

PR #316, opened 2022-12-15
libtock: Support 64-bit time values

alistair23:
This adds support for the new 64-bit alarm capsule introduced in https://github.com/tock/tock/pull/3355

We don't support setting alarms with a 64-bit reference, as the Tock syscall architecture doesn't allow it.


lschuermann:
Closing in line with #3343. Again, we're happy to support this with a separate dedicated capsule, in which case feel free to re-open!

PR #318, opened 2023-01-08
shell.nix: use riscv32-none-elf toolchain w/ newlib & ssp enabled

lschuermann:
This is required for `_FORTIFY_SOURCE` to work. In the default configuration, the newlib libc shipped with the compiler and used for header includes during compilation does not have SSP enabled. Thus use the flag `--enable-libssp` to enable SSP support.

However, GCC also by default builds its own libc, which needs to be disabled. Further, during compilation of GCC, the
`gcc_cv_libc_provides_ssp` environment variable must be set, as otherwise the newlib SSP support conflicts with GCC's SSP support (itself requiring glibc).

While this does work, it compiles GCC when entering libtock-c's nix-shell for the first time. This seems to be a worthwhile tradeoff, considering that this makes the libtock-c build infrastructure work without further adjustments.

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -60,7 +60,26 @@ in
       gcc-arm-embedded
       python3Full
       pythonPackages.tockloader
-    ] ++ (lib.optional (!disableRiscvToolchain) pkgsCross.riscv32-embedded.buildPackages.gcc);
+    ] ++ (lib.optional (!disableRiscvToolchain) (
+      pkgsCross.riscv32-embedded.buildPackages.gcc.override (oldCc: {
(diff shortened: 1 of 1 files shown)


hudson-ayers (approved):
My policy has been "let leon/amit specify shell.nix however they like" so I am sticking to that policy with this approval :)

ppannuto (approved):
bors r+

This is also a touch deeper in the gcc arcana than I know offhand (this comment also in response to @lschuermann's message on slack). In this context, a working solution seems like a good solution to me.

PR #319, opened 2023-01-30
changed notes to ones that the Microbit can play and removed *3

qpalzmz112:
More specifically, I raised the notes in music up two octaves so that each note played is at least 489 hz, and raised the notes in the Microbit test app by one octave for the same reason. Also, I added a comment in the Microbit test app mentioning that the Microbit can't play frequencies below 489 hz.


lschuermann (approved):

hudson-ayers (approved):
bors r+

PR #321, opened 2023-02-13
led: separate "exists" and led count functionality

thedcap:
Added the "led_check" function as command 0. The led_count was moved to (now calls)
command 4.

I have also submitted a PR for the led capsule in tock regarding issue #3375.


hudson-ayers:
Closing per https://github.com/tock/tock/pull/3397#issuecomment-1440655938

PR #324, opened 2023-05-19
Add support for USB HID keyboard

bradjc:
Libtock and example app to use USB keyboard with Tock.

Tested after the kernel side was merged.

### DIFF
--- a/examples/tests/keyboard_hid/main.c
+++ b/examples/tests/keyboard_hid/main.c
@@ -1,0 +1,30 @@
+#include <string.h>
+
+#include <button.h>
+#include <usb_keyboard_hid.h>
+
+
+uint8_t keyboard_buffer[64];
+
+// Callback for button presses.
+//   btn_num: The index of the button associated with the callback
+//   val: 1 if pressed, 0 if depressed
+static void button_callback(__attribute__ ((unused)) int   btn_num,
+                            __attribute__ ((unused)) int   val,
+                            __attribute__ ((unused)) int   arg2,
+                            __attribute__ ((unused)) void *ud) {
+
+  char message[] = "Hi from Tock!";
+  usb_keyboard_hid_send_string_sync(keyboard_buffer, message, strlen(message));
+}
+
+int main(void) {
+  int err;
+
+  err = button_subscribe(button_callback, NULL);
+  if (err < 0) return err;
+
+  button_enable_interrupt(0);
+
+  return 0;
+}
--- a/examples/tests/keyboard_hid/main.c
+++ b/examples/tests/keyboard_hid/main.c
@@ -15,7 +15,12 @@ static void button_callback(__attribute__ ((unused)) int   btn_num,
                             __attribute__ ((unused)) void *ud) {
 
   char message[] = "Hi from Tock!";
-  usb_keyboard_hid_send_string_sync(keyboard_buffer, message, strlen(message));
+  int ret;
+
+  ret = usb_keyboard_hid_send_string_sync(keyboard_buffer, message, strlen(message));
+  if (ret < 0) {
+    printf("ERROR sending string with USB keyboard HID: %i\n", ret);
+  }
 }
 
 int main(void) {
--- a/examples/tests/keyboard_hid/main.c
+++ b/examples/tests/keyboard_hid/main.c
@@ -1,3 +1,4 @@
+#include <stdio.h>
 #include <string.h>
 
 #include <button.h>
--- a/libtock/usb_keyboard_hid.c
+++ b/libtock/usb_keyboard_hid.c
@@ -70,6 +70,9 @@ static int to_hid_keycode(char c, uint8_t* modifier, uint8_t* key)
 {
   uint8_t shift = 2;  // KB_MODIFIER_LEFT_SHIFT = 2
 
+  // Clear modifier.
+  *modifier = 0;
+
   if ((c >= 'A') && (c <= 'Z')) {
     c         = tolower(c);
     *modifier = shift;
--- a/libtock/usb_keyboard_hid.c
+++ b/libtock/usb_keyboard_hid.c
@@ -1,0 +1,229 @@
+#include "tock.h"
+#include "usb_keyboard_hid.h"
+
+struct usb_keyboard_hid_data {
+  bool fired;
+  int callback_type;
+};
+
+static struct usb_keyboard_hid_data result = { .fired = false };
+
+// Internal callback for faking synchronous reads
+static void usb_keyboard_hid_upcall(int                          callback_type,
+                                    __attribute__ ((unused)) int unused1,
+                                    __attribute__ ((unused)) int unused2,
+                                    void*                        ud) {
+  struct usb_keyboard_hid_data* data = (struct usb_keyboard_hid_data*) ud;
+  data->callback_type = callback_type;
+  data->fired         = true;
+}
+
+
+int usb_keyboard_hid_set_callback(subscribe_upcall callback, void* callback_args) {
+  subscribe_return_t sval = subscribe(DRIVER_NUM_USBKEYBOARDHID, 0, callback, callback_args);
+  return tock_subscribe_return_to_returncode(sval);
+}
+
+int usb_keyboard_hid_set_receive_buffer(uint8_t* buffer, uint32_t len) {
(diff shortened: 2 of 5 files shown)


bradjc:
We can merge this now!

examples/tests/keyboard_hid/main.c
```diff
@@ -0,0 +1,30 @@
...
+// Callback for button presses.
+//   btn_num: The index of the button associated with the callback
+//   val: 1 if pressed, 0 if depressed
+static void button_callback(__attribute__ ((unused)) int   btn_num,
+                            __attribute__ ((unused)) int   val,
+                            __attribute__ ((unused)) int   arg2,
+                            __attribute__ ((unused)) void *ud) {
+
+  char message[] = "Hi from Tock!";
+  usb_keyboard_hid_send_string_sync(keyboard_buffer, message, strlen(message));
```
alistair23, line 18: Shouldn't you check `val` here?

libtock/usb_keyboard_hid.c
```diff
@@ -0,0 +1,229 @@
...
+      return 0;
+    case '/':   *key = 0x38;
+      return 0;
+  }
+  return -1;
+}
+
+
+
+int usb_keyboard_hid_send_letter_sync(uint8_t* buffer, char letter) {
```
alistair23, line 186: How do we know how long `buffer` is? This seems risky just indexing into it

libtock/usb_keyboard_hid.c
```diff
@@ -0,0 +1,229 @@
...
+  buffer[1] = 0;
+  buffer[2] = key;
+  buffer[3] = 0;
+  buffer[4] = 0;
+  buffer[5] = 0;
+  buffer[6] = 0;
+  buffer[7] = 0;
+
+  // Send key press.
+  err = usb_keyboard_hid_send_sync(buffer, 64);
```
alistair23, line 204: Why 64 bytes long? Does `buffer` need to be 64 bytes long?

alistair23 (changes requested):

libtock/usb_keyboard_hid.c
bradjc: Yes.

bradjc: The usb_hid capsule only works on 64 byte buffers.

examples/tests/keyboard_hid/main.c
```diff
@@ -0,0 +1,35 @@
+#include <string.h>
+
+#include <button.h>
+#include <usb_keyboard_hid.h>
+
+
+uint8_t keyboard_buffer[64];
```
alistair23, line 7: This 64 should be a macro

examples/tests/keyboard_hid/main.c
```diff
@@ -0,0 +1,35 @@
...
+
+// Callback for button presses.
+//   btn_num: The index of the button associated with the callback
+//   val: 1 if pressed, 0 if depressed
```
alistair23, line 11: Shouldn't `val` be checked to make sure the button was released?

libtock/usb_keyboard_hid.c
alistair23: The documentation of `usb_keyboard_hid_send_sync()` doesn't mention anything about 64 bytes though

libtock/usb_keyboard_hid.c
```diff
@@ -0,0 +1,232 @@
...
+  allow_rw_return_t aval = allow_readwrite(DRIVER_NUM_USBKEYBOARDHID, 1, (void*) buffer, len);
+  return tock_allow_rw_return_to_returncode(aval);
+}
+
+int usb_keyboard_hid_send(void) {
+  syscall_return_t cval = command(DRIVER_NUM_USBKEYBOARDHID, 1, 0, 0);
+  return tock_command_return_novalue_to_returncode(cval);
+}
+
+int usb_keyboard_hid_send_sync(uint8_t* buffer, uint32_t len) {
```
alistair23, line 45: If `len` must be 64 bytes why not enforce that with an assert?

libtock/usb_keyboard_hid.c
```diff
@@ -0,0 +1,232 @@
...
+
+  return RETURNCODE_SUCCESS;
+
+}
+
+int usb_keyboard_hid_send_string_sync(uint8_t* buffer, char* str, int length) {
+  int i = 0;
+  int err;
+
+  for (i = 0; i < length; i++) {
```
alistair23, line 225: We don't check that length is smaller or equal to `64`

libtock/usb_keyboard_hid.h
```diff
@@ -0,0 +1,27 @@
...
+int usb_keyboard_hid_set_receive_buffer(uint8_t* buffer, uint32_t len);
+int usb_keyboard_hid_set_send_buffer(uint8_t* buffer, uint32_t len);
+int usb_keyboard_hid_send(void);
+
+int usb_keyboard_hid_send_sync(uint8_t* buffer, uint32_t len);
+
+// Send one ASCII character. `buffer` must be at least 64 bytes.
+int usb_keyboard_hid_send_letter_sync(uint8_t* buffer, char letter);
+
+// Send an array of ASCII characters. `buffer` must be at least 64 bytes.
```
bradjc, line 21: Doc is here.

alistair23: There is nothing for `usb_keyboard_hid_send_sync()` though

libtock/usb_keyboard_hid.h
```diff
@@ -0,0 +1,27 @@
...
+#endif
+
+#define DRIVER_NUM_USBKEYBOARDHID 0x90005
+
+int usb_keyboard_hid_set_callback(subscribe_upcall callback, void* callback_args);
+int usb_keyboard_hid_set_receive_buffer(uint8_t* buffer, uint32_t len);
+int usb_keyboard_hid_set_send_buffer(uint8_t* buffer, uint32_t len);
+int usb_keyboard_hid_send(void);
+
+int usb_keyboard_hid_send_sync(uint8_t* buffer, uint32_t len);
```
bradjc, line 16:
```suggestion
// Send a raw keyboard HID packet. `buffer` must be at least 64 bytes.
int usb_keyboard_hid_send_sync(uint8_t* buffer, uint32_t len);
```

examples/tests/keyboard_hid/main.c
bradjc: I don't see what the extra layer of indirection accomplishes.

alistair23: Because you do this later

```c
err = usb_keyboard_hid_send_sync(buffer, 64);
```

If someone changes the size here that code will be incorrect. If you use a macro then the size of `keyboard_buffer` and the length supplied to functions will always match.

libtock/usb_keyboard_hid.c
alistair23: Does the capsule only work with 64 bytes? The documentation says at least 64 bytes

hudson-ayers:
This generally looks good to me, but I do not understand why the CI is failing

bradjc:
Fixed a bug where all of my characters were in caps.

libtock/usb_keyboard_hid.c
bradjc: It doesn't have to be less than 64. Only `buffer` needs to be 64 bytes.

libtock/usb_keyboard_hid.h
bradjc: This I fixed.

examples/tests/keyboard_hid/main.c
brghena: I think this is fine as-is. Spurious GPIO activations aren't critical to avoid here.

libtock/usb_keyboard_hid.c
bradjc: Mostly because I don't _really_ want to encode this length in the API. This limitation is a by-product of how the capsule is implemented, and from an aspirational point of view, we would fix the capsule rather than have all of these specific size requirements in userland.

I wouldn't want someone to see this code and copy these rigid length expectations as that isn't how we should design syscall APIs.

brghena (approved):

brghena:
I think this is good enough now. We can go ahead with it and make updates to improve later

PR #327, opened 2023-06-26
make: add default space in footer

bradjc:
We need space to be able to add credentials or new TBF headers.

I picked 3000 bytes pretty arbitrarily.


Configuration.mk
```diff
@@ -143,9 +143,13 @@ endif
 #
 # Provide the name, memory sizes, and required kernel version as arguments to
 # elf2tab so it can include the parameters in the TBF header.
+#
+# Add by default space in the footer that can be used for credentials. This
+# simplifies allowing tockloader to add credentials after a tbf is compiled.
 ELF2TAB_ARGS += -n $(PACKAGE_NAME)
 ELF2TAB_ARGS += --stack $(STACK_SIZE) --app-heap $(APP_HEAP_SIZE) --kernel-heap $(KERNEL_HEAP_SIZE)
 ELF2TAB_ARGS += --kernel-major $(KERNEL_MAJOR_VERSION) --kernel-minor $(KERNEL_MINOR_VERSION)
+ELF2TAB_ARGS += --minimum-footer-size 3000
```
alistair23, line 152: Should this be a flag option? As it adds a chunk of code size if it isn't being used

bradjc: Well, that is just one more thing for users to have to do if they want to do something like add credentials.

I think adding a flag to turn it off that an app makefile could include would be pretty reasonable, however.

hudson-ayers:
We will have to update github actions to use a version of elf2tab that supports this flag so that CI can pass

ppannuto:
Blocked on new `elf2tab` release. Latest version on crates.io is 0.10.12 which doesn't support this flag.

bradjc:
Seems like we should do an elf2tab release.

brghena (approved):

brghena (approved):

PR #329, opened 2023-07-06
newlib: Bump to 4.3.0.20230120

alistair23:


alistair23:
Urgh, the Ubuntu RISC-V toolchain seems too old

```
error: ../../newlib/rv32/rv32imac/libc.a(libc_a-exit.o): Mis-matched ISA version for 'i' exetension. 2.1 vs 2.0
```

I'll try rebuilding with something older

alistair23:
Hmm... After more testing I'm seeing odd crashes. It seems to be with the new newlib. I'm going to close this until I figure out what's going on

PR #335, opened 2023-07-25
HOTP Security Token Tutorial App

brghena:
HOTP Security Token tutorial.

This is in conjunction with tutorial text which will soon be pulled into the Tock Book. Multiple folders here are for checkpoints/milestones (whatever you want to call them) along the way. It is a little annoying that there's duplicated code across the apps, but at least we do check that each compiles. I've been careful to make sure that changes are synced across all of them.

Relies on tock/tock#3562 for the kernel side of things.


brghena:
I've lost the thread on some of the issues we've run into lately and I'm not sure why CI is broken for this. Was there some PR that fixes RISC-V boards failing with `"thread 'main' panicked at 'called Result::unwrap() on an Err value: Custom { kind: InvalidInput, error: "protected_region_size = 96 is too small for the TBF headers. Header size: 100" }',"`? Maybe tock/elf2tab#74 or tock/elf2tab#75?

bradjc:
Oh, because we ran CI on #327 it updated elf2tab and now no CI will pass.

Merging #327 will fix this. Basically for riscv builds we aren't reserving enough space for the header any more with the fixed addresses.

bradjc (approved):

PR #340, opened 2023-08-25
unit test: delay before discover

bradjc:
Ensure the service has time to register.

With KV testing I've had to manually add this delay to get the unit test framework to work.


ppannuto (approved):
The delay is a bit hack-y; in a real app probably should be some kind of retry loop, but for a unit test I think this is fine.

Maybe someday IPC should have an async / upcall / notify when service exists kind of thing, but the queue of IPC reform is long 😬

PR #344, opened 2023-08-30
lora: Initial support for sensor example

alistair23:


ppannuto (changes requested):
Super cool to have a LoRa example!!

examples/lora/build-RadioLib.sh
```diff
@@ -0,0 +1,22 @@
...
+mkdir -p build
+cd build
+
+cmake -G "CodeBlocks - Unix Makefiles" ..
+
+# Build the Tock example application
+# This will fail to link, as it can't find the libtock-c libraries
+# That's fine for us though, as we just need to build the RadioLib
+# library, not the entire example application
+make -j4 2> /dev/null
```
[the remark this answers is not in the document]
brghena: Combined with a good readme about how to get radiolib and build it yourself, this seems like it would be sufficient to me.

alistair23:
I think I have addressed all of the issues

examples/lora/sensor-receive/Makefile
```diff
@@ -0,0 +1,33 @@
...
+override CPPFLAGS += -I../RadioLib/examples/NonArduino/Tock
+
+# Include the base of libtock-c to fix the libtock/ includes from RadioLib
+override CPPFLAGS += -I$(TOCK_USERLAND_BASE_DIR)/
+
+override LEGACY_LIBS_cortex-m += ../RadioLib/examples/NonArduino/Tock/build/RadioLib/libRadioLib.a
+override LEGACY_LIBS_cortex-m0 += ../RadioLib/examples/NonArduino/Tock/build/RadioLib/libRadioLib.a
+override LEGACY_LIBS_cortex-m3 += ../RadioLib/examples/NonArduino/Tock/build/RadioLib/libRadioLib.a
+override LEGACY_LIBS_cortex-m4 += ../RadioLib/examples/NonArduino/Tock/build/RadioLib/libRadioLib.a
+override LEGACY_LIBS_cortex-m7 += ../RadioLib/examples/NonArduino/Tock/build/RadioLib/libRadioLib.a
```
bradjc, line 26:
```suggestion
```

Don't we just need to add it to the common variable?

examples/lora/sensor-transmit/Makefile
```diff
@@ -0,0 +1,33 @@
+# Makefile for user application
+
+# Specify this directory relative to the current application.
+TOCK_USERLAND_BASE_DIR = ../../../
+
+# We only compile this for Cortex-M platforms because of compiler
+# errors when building lua on risc-v.
```
bradjc, line 7:
```suggestion
# We only compile this for Cortex-M platforms because we only have
# libraries for Cortex-M.
```

examples/lora/sensor-receive/Makefile
```diff
@@ -0,0 +1,33 @@
+# Makefile for user application
+
+# Specify this directory relative to the current application.
+TOCK_USERLAND_BASE_DIR = ../../../
+
+# We only compile this for Cortex-M platforms because of compiler
+# errors when building lua on risc-v.
```
bradjc, line 7:
```suggestion
# We only compile this for Cortex-M platforms because we only have
# libraries for Cortex-M.
```

bradjc (commented):
I don't like submodules, so why not just commit the built libraries and relevant headers? That makes it easy for new users.

examples/lora/sensor-receive/Makefile
alistair23: There isn't one, at least not that I can find

alistair23:
> I don't like submodules, so why not just commit the built libraries and relevant headers? That makes it easy for new users.

I don't think we should do that. RadioLib is constantly adding new features, such as the work-in-progress [LoRaWAN support](https://github.com/jgromes/RadioLib/blob/master/examples/LoRaWAN/LoRaWAN_End_Device/LoRaWAN_End_Device.ino).

If we check in binaries they are going to become old and stale and make it difficult for users to add features, get updates or make changes. The `Makefile` will automatically build RadioLib so it should be easy for new users to get started, while still allowing them access to the RadioLib libraries. Just running `make` is all it takes to build these examples

As for the submodule, I understand you don't like them. But submodules are easier then having weird Makefile scripts to clone things. It is what we already do for lua

alistair23:
Urgh! Except that it doens't build in the CI. It looks like the cmake doesn't like the toolchain location

alistair23:
CI is all fixed, this is ready to go again

ppannuto (approved):
🎉

alistair23:
Ping!

alistair23:
Ping

alistair23:
Ping!

PR #347, opened 2023-09-14
Update the HOTP App to use Key Value Store

bradjc:
This changes the HOTP app to use KV rather than app state.

Advantages of KV:
1. Doesn't change the binary, so if we add credentials they remain valid.
2. Can be supported in libtock-rs.

This also just prints the code if usb-hid isn't set up.

I tested this with an nrf52840dk board and it seemed to work as expected. The programmed keys continued to work after resets.

### DIFF
--- a/examples/tutorials/hotp/hotp_oracle_complete/Makefile
+++ b/examples/tutorials/hotp/hotp_oracle_complete/Makefile
@@ -6,6 +6,9 @@ TOCK_USERLAND_BASE_DIR = ../../../..
 # Which files to compile.
 C_SRCS := $(wildcard *.c)
 
+# Make sure we have storage permissions.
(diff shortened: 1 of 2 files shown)


alistair23 (approved):

alistair23:
Looks ok besides the CI failure.

Should we add a README to this?

bradjc:
> Should we add a README to this?

Readme is part of the tutorial in the tock book.

brghena:
It looks like there's already a README which points to the Tock Book: https://github.com/tock/libtock-c/blob/3c7a601d06612c9a1132fd85ebd91f79ff0bfae0/examples/tutorials/hotp/hotp_oracle_complete/README.md

brghena (approved):

alistair23 (approved):

PR #355, opened 2023-11-07
libtock: sys: Implement _gettimeofday

alistair23:
Implement the _gettimeofday() function, which fixes build issues with C code that uses the gettimeofday syscall.

### DIFF
--- a/libtock/alarm.h
+++ b/libtock/alarm.h
@@ -31,6 +31,8 @@ extern "C" {
 
 #include "tock.h"
 
+#include <sys/time.h>
+
 /** \brief Opaque handle to a single-shot alarm.
  *
  * An opaque handle to an alarm created by `alarm_at` or `alarm_in`. Memory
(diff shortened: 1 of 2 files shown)


bradjc:
What would a caller do with the results from this function? I'm a little wary of implementing this in the core library and having it not really do the correct thing. Plus we do have a RTC interface, so in theory this could use the RTC and return a valid timestamp.

bradjc:
Seems like there are two competing ills: not having `_gettimeofday()` causing compilation failures, and having `gettimeofday()` not actually return the time of day (when a user might expect it to). I think I would be frustrated to find out that `gettimeofday()` compiles but doesn't actually do what the docs say it does.

The simple solution to my issue is just to have each library implement `_gettimeofday()` according to whatever requirements that library needs (perhaps just returning zeros is ok, maybe ticks is ok).

I would be supportive of a middle ground, where we keep the implementation in this PR but call it something else (maybe `gettimeasticks()` or something), and then have individual libraries include the code:

```c
int _gettimeofday(struct timeval *tv, void *tzvp) {
  return gettimeasticks(tv, tzvp);
}
```

if they want.

bradjc (approved):

twilfredo (approved):

alistair23:
This is ready to merge

PR #363, opened 2024-01-19
Add u8g2 library support

bradjc:
[u8g2](https://github.com/olikraus/u8g2) is a C library for creating text and graphics on monochrome displays.

This uses the screen driver with the u8g2 library.

I also included a handful of demo apps.


alevy (approved):
I think there is very little to approve here. The changes to libtock are minimal. The main body of code is in `u8g2/u8g2-tock.c`, which specializes the u8g2 library for Tock. Here, the main thing that is not straightforward is dynamically setting the screen info based on the what the driver responds in `u8x8_d_ssd1306_tock`.

The code here is not very self explanatory, and I needed to dig through the u8g2 library code to understand the flow of operations.

brghena:
I haven't looked yet, but given that Amit approves and Tyler has tested this, I think we can safely `last-call` it.

PR #368, opened 2024-02-09
make: tocklibrary: never build with -Werror

bradjc:
Our build all file adds the `-Werror` CFLAG, which in general is good, but when compiling a library we don't necessarily have control over the library not having any warnings. This makes CI fail when using external libraries.

## Todo

But, we want to enforce no warnings on libtock. Hmm...


lschuermann:
Can't we include these libraries using `-Isystem`? I believe that should stop the compiler from reporting warnings for them.

bradjc:
But in this case we are compiling the library in CI, not just using it.

ppannuto:
I suspect if we added a `$$((LIBNAME)_CPPFLAGS)`, which makes sense for other reasons probably, and added that after all the other flags here https://github.com/tock/libtock-c/blob/master/TockLibrary.mk#L102-L109, then you could add `-Wno-error` to those library flags. IIRC, last-flag has precedence in gcc when there is otherwise conflicts.

lschuermann (approved):

PR #370, opened 2024-02-09
libtock: Rewrite and standardize API

bradjc:
Rewrite libtock-c with a standardized format. See the guide.md file.

## Progress

- [x] chips
  - [x] remove all chips
- [x] crypto
  - [x] aes
  - [x] hmac (@bradjc)
  - [x] sha (@bradjc)
- [x] display
  - [x] screen
  - [x] text_screen
- [x] interface
  - [x] button
  - [x] buzzer
  - [x] console **tricky** (@alevy)
  - [x] led
  - [x] usb_keyboard_hid
- [ ] kernel
  - [ ] ipc **tricky** (@alevy)
  - [x] read only state (@bradjc)
- [ ] net
  - [ ] ble (@alevy)
  - [x] ieee802154 (@tyler-potyondy)
  - [x] lora_phy (@bradjc)
  - [x] nrf51_serialization **even worth doing??**
  - [x] udp (@tyler-potyondy)
- [ ] peripherals
  - [x] adc (@alevy)
  - [x] analog_comparator
  - [x] dac (@bradjc)
  - [x] gpio (@bradjc)
  - [x] gpio_async (@bradjc)
  - [ ] i2c_master (@alevy)
  - [ ] i2c_master_slave (@alevy)
  - [x] rng
  - [x] rtc
  - [x] spi (@hudson-ayers)
  - [x] spi_peripheral (@alevy)
  - [x] usb (@bradjc)
- [x] sensors
  - [x] ambient_light
  - [x] humidity
  - [x] ninedof
  - [x] pressure
  - [x] proximity
  - [x] sound_pressure
  - [x] temperature
  - [x] touch
- [x] storage
  - [x] app_state
  - [x] kv
  - [x] sdcard
  - [x] nonvolatile_storage
- [x] OTHER (not sure how to group)
  - [x] alarm/timer **tricky** (@hudson-ayers)
  - [x] crc
  - [x] unit_test

## Apps

- [x] tests/ieee802154
  - [x] tests/ieee802154/radio_tx
  - [x] tests/ieee802154/radio_ack
  - [x] tests/ieee802154/radio_rxtx
  - [x] tests/ieee802154/radio_rx
- [x] tests/udp
  - [x] tests/udp/udp_virt_app_kernel
  - [x] tests/udp/udp_virt_rx_tests/app2
  - [x] tests/udp/udp_virt_rx_tests/app1
  - [x] tests/udp/udp_send
  - [x] tests/udp/udp_rx
  - [x] tests/udp/udp_virt_app_tests/app2
  - [x] tests/udp/udp_virt_app_tests/app1
- [x] courses/2018-11-SenSys/sensys_udp_rx
- [x] courses/2018-11-SenSys/important-client/app2
- [x] courses/2018-11-SenSys/important-client/app1
- [x] ip_sense
- [x] tests/i2c
  - [x] tests/i2c/i2c_master_ping_pong
  - [x] tests/i2c/i2c_master_slave_ping_pong
- [x] tutorials/03_ble_scan
- [x] tests/imix
- [x] ./tests/sdcard
- [x] ./tests/adc_continuous
- [x] ./tests/hail
- [x] ./tests/adc_single_samples
- [x] ./tests/adc
- [x] ./tests/microbit_v2
- [x] ./adc
- [x] lvgl
- [x] music
- [x] witenergy
- [x] unit_tests/nrf52840dk-test
- [x] touch
- [x] tests/lsm303dlhc
- [x] lora/sensor-transmit
- [x] lora/sensor-receive
- [x] find_north
- [x] rtc_app
- [x] services/ble-env-sense
- [x] services/ble-env-sense/test-with-sensors
- [x] tutorials/05_ipc/led
- [x] tutorials/05_ipc/rng
- [x] tutorials/02_button_print
- [x] tutorials/hotp
  - [x] tutorials/hotp/hotp_starter
  - [x] tutorials/hotp/hotp_oracle_complete
  - [x] tutorials/hotp/hotp_milestone_three
  - [x] tutorials/hotp/hotp_milestone_two
  - [x] tutorials/hotp/hotp_milestone_one
- [x] screen
- [x] tests/aes
- [x] security_app
- [x] ble-uart
- [x] tests/multi_alarm_test
- [x] tests/analog_comparator
- [x] tests/app_state
- [x] tests/touch
- [x] tests/crc
- [x] tests/hail
- [x] tests/hmac
- [x] ./tests/ble/ble_nrf5x_concurrency/Advertiser1
- [x] ./tests/ble/ble_nrf5x_concurrency/Advertiser3
- [x] ./tests/ble/ble_nrf5x_concurrency/Advertiser4
- [x] ./tests/ble/ble_nrf5x_concurrency/Advertiser2
- [x] ./tests/ble/ble_advertise
- [x] ./tests/ble/ble_test_tx_power
- [x] ./tests/ble/ble_gap_library
- [x] ./tests/u8g2-demos/led-menu
- [x] ./tests/console_timeout
- [x] ./tests/number_guess_game
- [x] ./tests/screen-tock
- [x] ./tests/console


brghena:
I like this idea

bradjc:
Note: I don't know where to put crc. In the kernel syscalls its under "Cryptography", but that seems not right.

doc/guide.md
```diff
@@ -0,0 +1,241 @@
...
+  result.fired = true;
+  result.result = ret;
+}
+```
+
+Define the external function in the `libtocks_` name space for the sync
+operation:
+
+```c
+int libtocks_sensor_read(int* val) {
```
bradjc, line 227: This should probably return a `returncode_t`.

bradjc:
I've gone through and updated temperature and button to match the guide I wrote.

doc/guide.md
```diff
@@ -0,0 +1,241 @@
...
+| Header File Name | `libtock/[category]/internal/[name]_internal.h` |
+
+### Upcalls
+
+Each supported upcall must have its own function.
+
+The signature is:
+
+```c
+returncode_t set_upcall(subscribe_upcall callback, void* callback_args);
```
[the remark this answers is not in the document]
alistair23: `opaque` is a common name in C code

doc/guide.md
```diff
@@ -0,0 +1,241 @@
...
+
+## Internal APIs
+
+All direct system calls must be wrapped in simple functions.
+
+| Characteristic   | Value                                           |
+|------------------|-------------------------------------------------|
+| Location         | `libtock/[category]/internal`                   |
+| Source File Name | `libtock/[category]/internal/[name].c`          |
+| Header File Name | `libtock/[category]/internal/[name]_internal.h` |
```
tyler-potyondy, line 23: What is the reasoning for `[name]_internal.h` vs `[name].h`?

doc/guide.md
```diff
@@ -0,0 +1,241 @@
...
+```c
+bool [name]_exists(void) {
+  return driver_exists(DRIVER_NUM_[NAME]);
+}
+```
+
+
+## Asynchronous APIs
+
+All common system call operations should have asynchronous versions.
```
tyler-potyondy, line 146: I think it might be helpful to describe more what constitutes a "common system call operation".

doc/guide.md
tyler-potyondy, line 227: Is `libtocks_` the proposed convention for synchronous functions?

tyler-potyondy:
@bradjc this looks great and does an excellent job formalizing the convention in `guide.md`. I'm in favor of these changes and think it would mark a good step forward towards having more accessible documentation.

doc/guide.md
tyler-potyondy: I'm in favor of this.

bradjc: Yeah, I don't love it though.

doc/guide.md
bradjc: I think I was thinking so it wouldn't conflict with the "non internal" header. But the directory structure should take care of that.

doc/guide.md
brghena: Even `libtocksync` would be better

brghena (commented):
This represents a significant improvement to libtock-c and I strongly support it.

I read through all of the guide and I'm for it. We can bikeshed some names for a bit, but the plan it lays out resonates with me.

doc/guide.md
bradjc: Definitely syscalls, and its probably worth not having multiple files with the same name, even if we could get away with it.

bradjc:
Couple things:
1. Should we namespace libtock functions with `libtock_`?
2. Should we put all header files in `#include <libtock/sensors/temperature.h>`? If so, how?
3. We probably need to specify the cpp wrapper for the header files.

tyler-potyondy:
>     1. Should we namespace libtock functions with `libtock_`?

I am in support of this.
 
>     2. Should we put all header files in `#include <libtock/sensors/temperature.h>`? If so, how?
> 
>     3. We probably need to specify the cpp wrapper for the header files.

Could you provide some more context with this. I'm not sure I follow with what this would be / how it would look.

bradjc:
What about putting all header files in `libtock/include/libtock/[category]/`? Then our apps would do `#include <libtock/interface/led.h>`.

This is what zephyr does: https://github.com/zephyrproject-rtos/zephyr/tree/main/include/zephyr

alistair23:
The current method of including headers causes conflicts with other libraries. For example `#include <alarm.h>` is pretty common.

So putting it under libtock would be a great idea (the categories also help fix this issue)

brghena:
While working on the redesign: is this the time to approach how to handle multiple instantiations of a single syscall driver?

Came up again in https://github.com/tock/tock/pull/3867

This could be handled with a struct for each driver that encapsulates its instantiated syscall number. Or it could be handled in the build system possibly, allowing us to include duplicate copies of a C file with a different `#define` value and name?

Or we could keep punting on it as an orthogonal problem that mostly occurs for downstream users (who could just make a duplicate C file with a different name if they wanted to).

bradjc:
> This could be handled with a struct for each driver that encapsulates its instantiated syscall number. Or it could be handled in the build system possibly, allowing us to include duplicate copies of a C file with a different `#define` value and name?

In these approaches, how would we default to what we have now? It seems like either we would have to add additional code in the app to choose the driver numbers, or apps could configure themselves in a makefile and completely change what they are doing with no changes to code. Neither seems desirable.

doc/guide.md
```diff
@@ -0,0 +1,296 @@
...
+```
+
+### Upcalls
+
+Each supported upcall must have its own function.
+
+The signature is:
+
+```c
+returncode_t libtock_[name]_set_upcall(subscribe_upcall callback, void* opaque);
```
bradjc, line 61: I think I want to change this to:
```suggestion
returncode_t libtock_[name]_set_upcall(subscribe_upcall upcall, void* opaque);
```

bradjc:
We need to figure out an allow buffer management strategy. What layer is responsible for un-allowing the buffer? Are buffers passed back in callbacks? Or should users of the library hold their own copy of the pointer to the buffer?

bradjc:
The unit test framework stores state. Where do we put it?

bradjc:
Three more issues that have come up:

1. We need to use more of `status_to_returncode`
2. Should the upcall format be `int`? Why not just `uint32_t`?
3. Should we document that all tock.h functions start with `tock_`? What about `subscribe()`, etc.?

bradjc:
I would also like to revisit what goes in `examples/` and what goes in `examples/tests`.

bradjc:
@tyler-potyondy @alevy Anything that would help with this?

tyler-potyondy:
> @tyler-potyondy @alevy Anything that would help with this?

Working on this now. Will push changes for UDP today and 15.4 tomorrow.

doc/guide.md
```diff
@@ -0,0 +1,353 @@
...
+};
+
+static struct data result = { .fired = false };
+```
+
+Define a static callback function to pass to the asynchronous implementation:
+
+```c
+static void sensor_cb(returncode_t ret, int val) {
+  result.temp = val;
```
tyler-potyondy, line 329: To match the struct defined.
```suggestion
  result.val = val;
```

tyler-potyondy (dismissed):
A few small fixes after using the guide.

bradjc:
Ping @alevy @hudson-ayers @tyler-potyondy

hudson-ayers:
I am noticing that a lot of drivers have been ported but not the corresponding examples. what is the plan there?

bradjc:
I added the list of apps that don't build (for me in my one test on 6b80b89c at least) to the main PR description.

The consensus on the call today is once this PR builds to go ahead and merge it.

bradjc:
I edited the commits in this PR to remove duplicates and clean things up. Sorry if that messes with branches, but you should be able to just cherrypick any changes.

bradjc:
I don't know what to do with UDP. The driver had some hidden state. I removed it so we can avoid warnings, but we need an actual fix.

doc/guide.md
```diff
@@ -0,0 +1,353 @@
...
+
+```c
+static void sensor_cb(returncode_t ret, int val) {
+  result.val = val;
+  result.fired = true;
+  result.result = ret;
+}
+```
+
+Define the external function in the `libtocksync_` name space for the sync
```
Samir-Rashid, line 335: Why do you want every function to have `libtock_` prepended? This seems more verbose for no benefit. I understand the purpose of putting `sync`, but not `libtock`.

brghena: C doesn't really give you a choice since it doesn't have a namespace concept. If your driver is implementing a timer, `libtock_timer` is a unique name but `timer` is not.

bradjc: If you are writing a "pure" libtock-c app, where you are only calling libtock functions, I agree the prefix is a bit redundant. But for more complicated apps that use multiple libraries, it becomes very convenient to have namespacing to make the code easier to read.

I think we should get to a place where someone is writing a c app with some other library (e.g. lora or graphics or ML) and then just happens to use libtock, maybe for sensor data.

libtock/net/syscalls/eui64_syscalls.c
```diff
@@ -0,0 +1,7 @@
+#include "eui64_syscalls.h"
+
```
bradjc, line 2: Need driver exists.

libtock/net/eui64.h
```diff
@@ -0,0 +1,14 @@
+#pragma once
+
+#include "tock.h"
+#include "syscalls/eui64_syscalls.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+returncode_t get_eui64(uint64_t* eui64);
```
bradjc, line 10: Needs to be `libtock_eui64_get(...)`.

libtock/net/eui64.c
```diff
@@ -0,0 +1,5 @@
+#include "eui64.h"
+
+returncode_t get_eui64(uint64_t *eui64) {
+  return libtock_eui64_command_getter(eui64);
```
bradjc, line 4: -ter

bradjc:
Ok we're down to just the networking and SPI test apps that do not compile.

bradjc:
I added some documentation for libtock and libtock-sync/services.

I moved the printf/_write use of yield to libtock-sync. I moved read only state yield to libtock-sync. I moved unit test to libtock-sync services. The only remaining uses of yield in libtock are libraries which have not been updated.

bradjc:
I cleaned up the commits.

I also fixed up the makefile `-I` flags so that apps have to use the `libtock/` prefix in their includes.

I think the remaining todos are:
1. alarm
2. more tests on nrf52840dk

bradjc:
I'm ready to merge this.

ppannuto (approved):
This is massive, and in pretty good shape / clean at the moment. I think we should merge this, and then if there's anything missing we can clean it up with patches later. I can't see this getting much better than it is in the short term.

ppannuto:
Pushed a tag for users who would like to defer updating to the new API here: https://github.com/tock/libtock-c/releases/tag/2.x-legacy-api — merging!

PR #371, opened 2024-02-13
15.4 Send Raw Syscall Command

tyler-potyondy:
Tock / libtock-c currently provide support for forming and securing 15.4 packets. The `ieee802154_send(..)` method accepts a payload in addition to configuration fields (seen below):

```
int ieee802154_send(unsigned short   addr,
                    security_level_t level,
                    key_id_mode_t    key_id_mode,
                    unsigned char *  key_id,
                    const char *     payload,
                    unsigned char    len)
```

This provides a helpful feature to users desiring to easily send 15.4 packets without implementing header forming / parsing. However, this does not allow for user processes to fully form 15.4 packets. 

In OpenThread's platform abstraction layer, OpenThread passes a fully formed 15.4 packet to the PAL function interfacing with the radio. This creates a challenge as there does not exist a syscall command for sending preformed packets. The current implementation would require the PAL libtock-c function deconstruct the fully formed 15.4 packet to retrieve and pass the fields into `ieee802154_send(..)`.

This PR adds a libtock-c syscall command for direct sending (i.e. sending a packet fully formed by the user process). This PR coincides with https://github.com/tock/tock/pull/3851.

To demonstrate this use and for testing, a test has been added to  `examples/tests/ieee802154/radio_tx_direct`.

(As a side note, I also updated the READMEs in the 15.4 test directory to avoid repeating the README in each subdirectory).


examples/tests/ieee802154/README.md
```diff
@@ -0,0 +1,9 @@
+This is one of five  applications to test 802.15.4 packet reception
+and transmission. The five apps are:
+
+radio_ack: Sends packets, using a printf to signal whether they were
+           acknowledged. Also receives packets.
+radio_rx: Receives packets only.
+radio_rxtx: Sends and receives packets.
+radio_tx: Sends packets only.
+radio_tx_direct: Send packets fully formed by userprocess. This example sends an ACK packet. For forming headers, use the generic `ieee802154_send(..)` method.
```
brghena, line 9: Rename to "raw" here to match the kernel. (Plus, I think it makes more sense)

This same comment applies to everywhere the word "direct" appears.

examples/tests/ieee802154/radio_tx_direct/main.c
```diff
@@ -0,0 +1,40 @@
...
+#include "led.h"
+#include "timer.h"
+#include "tock.h"
+
+// IEEE 802.15.4 sample packet transmission app.
+// Continually transmits frames at the specified short address to the specified
+// destination address.
+
+#define BUF_SIZE 3
+char packet[BUF_SIZE] = {2, 0, 1};
```
brghena, line 15: Is this a valid and meaningful 15.4 packet? It would be nice for it to be.

brghena (commented):
Looks good apart from the naming changes

examples/tests/ieee802154/radio_tx_direct/main.c
tyler-potyondy: Yes, this is a simple ACK packet. I will update this to have a comment mentioning this.

brghena (approved):

tyler-potyondy:
> Looks like the formatting is bad though.

Yeah, I made a mistake and I'm trying to fix now... I'll make a comment once everything is updated and good to go (sorry for the messy force pushes and updates. Tracking down all the direct renames proved more challenging than I initially expected).

tyler-potyondy:
Alright everything should be good to go now and rebased to master. @brghena @bradjc should I squash the unnecessary commits before we merge to clean up the history?

brghena (approved):

brghena:
I don't know that we have a policy. Personally, I'd squash these commits since it's two merge commits, and two small bugfixes. It's not really history of a feature in a meaningful way that someone might want to look at in prior history.

PR #376, opened 2024-03-19
IEEE 802.15.4 Updates for Kernel Ring Buffer

tyler-potyondy:
This PR updates the libtock-c 15.4 functions to use the receive ring buffer the kernel now expects. I updated the relevant 15.4 libtock functions in addition to the various test apps. Additionally, I attempted to update logic using hard coded values for the frame meta data length.

The test apps are currently operating as expected with these changes.


ppannuto (approved):

tyler-potyondy:
Looks like this PR failed when running the final checks before merging. @ppannuto  should we run the CI again or was this due to some actual issue?

ppannuto:
image
Looks like some weird (hopefully transient) infra error — trying again..

PR #380, opened 2024-03-23
OpenThread Port

tyler-potyondy:
This PR provides a functional OpenThread implementation for Tock. This is not the final version of OpenThread in Tock, but represents a stable state. As is, this PR adds functionality for joining an existing network (as a minimal thread device), remaining attached to the network, and sending and receiving UDP packets. 

The openthread_hello test application provides this functionality. The accompanying README provides in depth details to the current design and known issues that remain. To setup and test this device in a thread network, follow the instructions in the README.

Some issues that may occur if using/testing this current version:
- Child detachment: after ~7 minutes, the Tock device does not send a heart beat to remain attached to the network. This is occurring due to an alarm implementation issue. 
- 6LoWPAN capsule panic: The 6LoWPAN capsule has a bug which on occasion causes a panic.

Todos:

- [x] Update makefiles and build structure from https://github.com/tock/libtock-c/tree/ot-bradjc
- [x] Need to actually figure out what flags we need set. I included some in the makefile while debugging, which may be redundant or unwanted. Should document which flags we are setting in the makefile(s). I think we don't set any flags, and just use the header files.
- [x] I had to hack around the cortex-m0 for mbedtls. I tried to make it clear how I modified the header file. I think it should be possible to use #undef and include the version from openthread instead, and then unset the asm flag on thumb1. (in ot-bradjc)
- [x] I think we should create a new helper makefile in libtock-c/openthread that apps can include which setup all of the EXTERN_LIBS needed for openthread. This means we don’t have to copy the list to every ot app. We also can’t use the makefile.app because the extern_libs variable is not handled recursively.
- [x] Rename folder to just openthread (or I suppose libopenthread would also be consistent).
- [x] Simplify the platform directory, remove empty/unneeded files, remove folder nesting.
- [x] Rename from nrf52840 to tock in the platform library.
- [x] Move config headers to their own directory. We should include them consistently in all of the builds.


alevy:
So, just to clarify, there are basically two "real" bugs remaining in this current status, but neither are to do with any of the files added/touched in this PR.

One is in the kernel's 6lowpan driver (?).

One is in libtock-c's timer/alarm subsystem.

ot-tock/Makefile
bradjc, line 1: Why not just call this folder `openthread`? "ot" likely means "opentitan" in the Tock context.

ot-tock/create_ot-tock
bradjc, line 1: needs an extension

.gitmodules
```diff
@@ -7,3 +7,6 @@
 [submodule "examples/lora/RadioLib"]
 	path = examples/lora/RadioLib
 	url = https://github.com/jgromes/RadioLib.git
+[submodule "ot-tock/openthread"]
+	path = ot-tock/openthread
+	url = https://github.com/openthread/openthread.git
```
bradjc, line 12: What ensures you have the git submodule checked out?

ot-tock/Makefile
tyler-potyondy: The convention within OpenThread ports is typically of the form "ot-nrf528xx" (that was the reason for ot-tock). My hesitation calling this directory openthread is that the `openthread` repo is a subdirectory within the current `ot-tock` directory. 

I agree `ot` can easily be confused with "opentitan" so we should change this.

.gitmodules
tyler-potyondy: Should we add `git submodule update --init' to the openthread build script `create_ot-tock`?

tyler-potyondy:
> So, just to clarify, there are basically two "real" bugs remaining in this current status, but neither are to do with any of the files added/touched in this PR.
> 
> One is in the kernel's 6lowpan driver (?).
> 
> One is in libtock-c's timer/alarm subsystem.

Yes that is correct. (caveated by being unsure if the alarm bug is due to the OpenThread PAL, the libtock alarm subsystem, or a combination of both)

There is a brief PR I just created for the 15.4 capsule receive fixing a data offset miscalculation (so block on testing with this implementation until https://github.com/tock/tock/pull/3933 merges). 

I'll copy the more extensive todo list from the README here so it is more easily viewable: 

> ### TODOs and Shortcomings
> 
> - The Tock Thread device disconnects from the network (i.e. does not send child update request 
> and subsequently times out) after ~7 minutes. This bug has been traced to an issue with the alarm
> implementation but has not been resolved.
> - The 6LoWPAN capsule on occassion panics (this behavior is observed semi frequently)
> - The code size is quite large as is for the current implementation. There are numerous 
> opportunities to reduce this (use hardware crypto instead of mbed crypto library).
> - The openthread_hello `Makefile` will build all needed OpenThread and platform dependencies. 
> However, the `make` build system, as is, does not rebuild the platform or OpenThread library 
> if changes are made to either the platform or openthread directories.
> - Channel switching is not yet implemented
> - The RSSI value is hard coded

ot-tock/Makefile
bradjc: Yeah within the openthread ecosystem I guess "ot" is no longer ambiguous.

.gitmodules
bradjc: I should be able to type `make` in an app and everything should just work.

tyler-potyondy: The scripts are updated now such that you should be able to do this.

openthread-tock/platform/platform-tock/alarm.c
```diff
@@ -0,0 +1,68 @@
...
+uint32_t otPlatAlarmMilliGetNow(void) {
+  // TODO: there is a bug in this implementation causing the timer value to 
+  // wrap around 500,000 milliseconds. This may be an issue with the 
+  // `gettimeasticks` function or in the logic here.
+  struct timeval tv;
+  gettimeasticks(&tv, NULL);    // second arg is unused
+
+  uint32_t nowSeconds    = tv.tv_sec;
+  uint32_t nowMicro      = tv.tv_usec;
+  uint32_t nowMilli32bit = (nowSeconds * 1000) + (nowMicro / 1000);
```
alistair23, line 65: This will loose precision, that might be the issue you are seeing

openthread-tock/platform/platform-tock/misc.c
```diff
@@ -0,0 +1,18 @@
+// TODO COPYRIGHT
+
+#include <openthread/platform/misc.h>
+
+void otPlatReset(otInstance *aInstance) {
+  // TODO
```
alistair23, line 6: You can `exit()`? Tock can be set to reset apps on exit

alistair23 (changes requested):

openthread-tock/CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
+
+#  Copyright (c) 2019, The OpenThread Authors.
```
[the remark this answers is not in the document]
tyler-potyondy: This `CMakeLists.txt` is used for building the openthread static library (separate from the app in examples). This is used as the starting point to initiate the openthread build system in conjunction with `create_openthread-tock.sh`.

Since `examples/openthread` directory is separate from this as it links against the static library that is built, I think it makes more sense to keep the openthread library in one place and the application specific implementation in examples.

openthread-tock/CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
...
+# Compile for ARM
+include(arm-none-eabi.cmake)
+
+project(openthread_dummy)
+
+add_subdirectory(openthread)
+
+target_compile_definitions(ot-config INTERFACE
+    OPENTHREAD_PLATFORM_CORE_CONFIG_FILE="openthread-core-tock-config.h"
+    # OPENTHREAD_CORE_CONFIG_PLATFORM_CHECK_FILE="openthread-core-cc2538-config-check.h"
```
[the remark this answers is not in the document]
tyler-potyondy: Without this, we are unable to specify a custom config file (i.e. that can specify which flash interface to use). The commented out line 40 should be removed.

Can you elaborate on why we don't need line 38/39?

openthread-tock/platform/platform-tock/alarm.c
```diff
@@ -0,0 +1,68 @@
...
+  if ((aT0 > UINT32_MAX) || (aDt > UINT32_MAX)) {
+    assert(false);
+    return;
+  }
+
+  // inputs aT0, aDt are in ms, but libtock call must be in clock ticks
+  uint32_t ref = milliToTicks(aT0);
+  uint32_t dt  = milliToTicks(aDt);
+
+  // NOTE: This is a problem that will have to be fixed. Thread wants to
```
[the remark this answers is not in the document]
tyler-potyondy: I am assuming you are referring to setting a large `ref` for the case just above this in the guard statement 

```rust
  // check that aT0 and aDt are both less than max
  // uint32_t. this is a temporary fix until we implement
  // the wrap around functionality
  if ((aT0 > UINT32_MAX) || (aDt > UINT32_MAX)) {
    assert(false);
    return;
  }
```

We could just set a large `ref` for that instance. However, the behavior is undefined (as we are implementing) for those behaviors, so just returning should suffice (i.e. let openthread think it has set the alarm when in actuality no alarm has been set). I will remove the assert.

openthread-tock/platform/platform-tock/alarm.c
tyler-potyondy: Although this loses precision, as far as I can tell, the precision that is lost is the 3 least significant digits (micro -> milli). This should be acceptable since these functions assume millisecond precision.

openthread-tock/platform/platform-tock/misc.c
tyler-potyondy: This is a good idea! (I'm planning to implement these todos in a follow up PR and will look further into this then)

tyler-potyondy:
Update: I believe I have addressed all comments either in the most recent commits or responded to them above. (with the exception of the copyright). Looking at other platforms that have ported openthread, it appears they use the openthread copyright statement. If someone else can confirm that this seems reasonable, I'll push changes including the copyright.

openthread-tock/create_openthread-tock.sh
```diff
@@ -0,0 +1,79 @@
...
+
+
+# Confirm that the git submodule has been checked out.
+git submodule status -- openthread >/dev/null 2>&1
+
+if [ $? -ne 0 ]; then
+    echo "${red}Error: git submodule 'openthread' not found.${normal}" 
+    echo "To checkout the submodule, use the command ${green}$ git submodule update --init -- openthread${normal}"
+    exit 
+fi
```
bradjc, line 24: If we're going to use a submodule this needs to "just work", aka do this for me.

openthread-tock/Makefile
bradjc, line 1: Can we just name this directory `openthread`? None of our other libraries include the `-tock` suffix.

openthread-tock/platform/platform-tock/alarm.c
```diff
@@ -0,0 +1,68 @@
+// TODO ADD COPYRIGHT ETC
```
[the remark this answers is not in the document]
bradjc: You don't need to worry about this.

openthread-tock/platform/platform-tock/diag.c
```diff
@@ -0,0 +1,3 @@
+// TODO
+
+// This file is being kept as a place holder, but likely is not necessary for the PAL implementation.
```
bradjc, line 3: ?

openthread-tock/platform/platform-tock/flash.c
```diff
@@ -0,0 +1,73 @@
...
+#include <string.h>
+
+const uint32_t SWAP_SIZE = 512;
+uint8_t swapPart0[512];
+uint8_t swapPart1[512];
+
+// NOTE: This is just a temporary version of this file. OpenThread will
+// not operate properly if it is unable to "save" the state. This provides
+// an in memory representation of "flash" that will be lost across power
+// cycles.
```
bradjc, line 13: I think we should do this.

openthread-tock/platform/platform-tock/flash.c
```diff
@@ -0,0 +1,73 @@
...
+    arr[i] = 0;
+  }
+}
+
+void otPlatFlashInit(otInstance *aInstance) {
+  OT_UNUSED_VARIABLE(aInstance);
+
+  // initialize swapPart0 and swapPart1
+  zeroArray(swapPart0, SWAP_SIZE);
+  zeroArray(swapPart1, SWAP_SIZE);
```
bradjc, line 29:
```suggestion
  memset(swapPart0, 0, SWAP_SIZE);
  memset(swapPart1, 0, SWAP_SIZE);
```

openthread-tock/platform/platform-tock/openthread-core-tock-config.h
```diff
@@ -0,0 +1,338 @@
...
+#endif
+
+/**
+ * @def OPENTHREAD_CONFIG_PLATFORM_INFO
+ *
+ * The platform-specific string to insert into the OpenThread version string.
+ *
+ */
+#ifndef OPENTHREAD_CONFIG_PLATFORM_INFO
+#define OPENTHREAD_CONFIG_PLATFORM_INFO "NRF52840"
```
bradjc, line 62:
```suggestion
#define OPENTHREAD_CONFIG_PLATFORM_INFO "TOCK"
```

openthread-tock/platform/platform-tock/platform-config.h
bradjc, line 1: Do we need this file?

openthread-tock/platform/platform-tock/radio.c
```diff
@@ -0,0 +1,281 @@
...
+// nrf52840 ACK on transmit is currently unimplemented. We fake this 
+// for now by always saying the sent packet was ACKed by the receiver.
+static uint8_t ack_mPSdu[ACK_SIZE] = {0x02, 0x00, 0x00};
+static otRadioFrame ackFrame = {
+  .mPsdu   = ack_mPSdu,
+  .mLength = 3
+};
+
+void otPlatRadioGetIeeeEui64(otInstance *aInstance, uint8_t *aIeeeEui64) {
+  // TODO HARDCODED FOR NOW
```
bradjc, line 27: I assume this won't work for a network of devices?

openthread-tock/platform/platform-tock/uart.c
```diff
@@ -0,0 +1 @@
+// TODO.. leaving this file as a placeholder
```
bradjc, line 1: Probably just remove

openthread-tock/lib/libtcplp/Makefile.app
bradjc, line 1: Why do we need all of these folders?

bradjc (changes requested):

openthread-tock/platform/platform-tock/radio.c
tyler-potyondy: Long term this isn't great. Interestingly, this looks like it _will work_ for a network of devices since OpenThread generates a somewhat random ipv6 address for the device from this eui64 (said differently, I have not seen the same ipv6 address repeated while testing).

Adding support for obtaining the eui64 is on my todo list for this week.

openthread-tock/lib/libtcplp/Makefile.app
tyler-potyondy: This arose due to the libtock-c build system (as I understand how it works). I would be very much in favor of a cleaner solution.

OpenThread generates 5 static libraries as part of the build system (libtcplp being one of these). We need to include these 5 different external libraries in the OpenThread makefile. I did not see a way in the current build system to include multiple external libraries from the same directory since the libtock-c external libraries build system expects a very specific directory structure. @bradjc do we have a better method for doing this?

alistair23:
> Update: I believe I have addressed all comments either in the most recent commits or responded to them above. (with the exception of the copyright). Looking at other platforms that have ported openthread, it appears they use the openthread copyright statement. If someone else can confirm that this seems reasonable, I'll push changes including the copyright.

I'm assuming you copied OpenThread code and then modified it, those files should then include the original copyright. If you made major changes you can append your copyright as well.

If you did a full re-write or a clean implementation then you can add a your own instead of OpenThread.

A lot of libtock-c doesn't specify anything in the file. Which I **think** means it's just copyrighted by you and dual licensed. It's probably best to be specific here as it's a mix between your work and OpenThread. That hopefully makes it clear

openthread-tock/create_openthread-tock.sh
bradjc: No. Let's just use a .zip. Then we also get the benefit of working if someone isn't using git and just downloads the code.

lschuermann: I disagree. This is distinct from how we distribute toolchains, where ZIPs are arguably the best solution. Here we just want to just some pre-existing codebase, and with a ZIP we lose all of the benefits of using the code directly, such as the connection to a specific git commit, being able to immediately tell the authenticity of the code we pull in, etc. The recent attack on XZ arguably has shown us the importance on reducing reliance on non-auditable blobs.

bradjc: We also get all of the downsides of git submodules, too. The clunky interface, the challenges when developing on multiple branches, the inability to have multiple versions concurrently, the frustration with even discovering you _need_ a submodule, the conflicts if you change something in the submodule and the referenced commit changes, the inability to just download the library.

I agree that a zip makes it harder to tell if the zip is from the intended source. I'm not particularly concerned with this compared to ease of use. I disagree with the other two points (we use zips for libraries, not toolchains, and the commit is _glaringly_ obvious (much more so than with git and its hidden files). What are the others (etc)?

bradjc: I'm less and less convinced we can have users compile this when they build an openthread app. I think we need to provide precompiled libraries ala newlib. If we do that, that largely makes this issue moot.

bradjc:
I would like to propose we compile openthread in advance and provide pre-compiled binaries (like we do with newlib/libcpp/picolib). I do not think we can provide a quality user experience with our current build system (and interfacing with openthread's build system) to compile the library at the same time as compiling apps. Thoughts?

tyler-potyondy:
> I would like to propose we compile openthread in advance and provide pre-compiled binaries (like we do with newlib/libcpp/picolib). I do not think we can provide a quality user experience with our current build system (and interfacing with openthread's build system) to compile the library at the same time as compiling apps. Thoughts?

My preference is to use git submodules here, but I understand your frustrations / concerns @bradjc. In my opinion, the submodule is nice for openthread since the code base will likely not be altered and it serves as an explicit pointer to the detached head at the commit we set in the `libtock-c` repo. Additionally, the submodule in my opinion is more explicit for people navigating through the codebase in github as a pointer to the openthread repo.

If the consensus is for using a zip over the submodule, this would not be an issue as we could obtain a zip of the openthread repo. However, I think using an entirely precompiled binary would be a mistake. The biggest downside I see to precompiled binaries surrounds customization of openthread. OpenThread provides a high degree of customizable build flags that can be specified in config header files and also as CMake arguments. Using a precompiled binary pigeonholes us to our specific build configuration for OpenThread. A good example of this is the build flag for logging and whether to compile the Full Thread Device vs Minimal Thread Device. 

If we remove the submodule, the entire library and repo will build using `make` from within the directory. The need to use make to kick off a script that then launches a CMake build system is convoluted, however, I do not see another way to allow for customizable OpenThread builds that use developer specific build flags.

openthread-tock/platform/platform-tock/diag.c
tyler-potyondy: I'll go ahead and remove these.

openthread-tock/platform/platform-tock/flash.c
tyler-potyondy: We have an implementation of this that uses non-volatile flash (which is accessible by all apps). @Samir-Rashid and @atar13 are working on implementing a non-volatile flash capsule with isolation.

For this PR, do we want to just use the memory representation? It seems placing a networkkey in flash accessible by all apps is antithetical to Tock's separation.

tyler-potyondy:
Also just as a comment, the build system in the current state should work as expected. However, I need to push some changes to bring the branch in this PR up to date with changes we have made in the kernel.

bradjc:
> A good example of this is the build flag for logging and whether to compile the Full Thread Device vs Minimal Thread Device.

I think for three examples we just build the library three or four times and include pre-compiled libraries for each (or maybe there are four: FTD/MTD x LOG/NOLOG). Are there more configurations that are realistic? We can compile everything in the open thread source, and allow the linker to remove dead code once it knows what the applications are. I understand the desire to support customizability, but 2-4 options is a pretty small number. Are there configurations which conflict? The [master](https://developer.nordicsemi.com/nRF_Connect_SDK/doc/2.6.0/kconfig/#CONFIG_OPENTHREAD_NORDIC_LIBRARY_MASTER) config option I've uses suggests openthread is ok with just compiling everything.





> the entire library and repo will build using `make` from within the directory

How are we going to support cortex-m0/m3/m4/m7/rv32i/rv32imac? And different versions of libcpp and newlib? Picolib? precompiling helps us avoid these issues by doing a more manual process offline. If we have users build on the fly, we need some make structure to handle all of that.

> Using a precompiled binary pigeonholes us to our specific build configuration for OpenThread.

I think this is a good thing for most users who just want something that works. For others, we document everything, include scripts and makefiles, include dockerfiles that show the _exact_ steps. Same as we do for newlib, which someone may want to customize and certainly can, but most people just need something that works.

openthread-tock/platform/platform-tock/flash.c
bradjc: I'm wary of open todos and a not quite complete port of openthread. This is by no means directed at you, or anyone specifically, but urgency/code updates have a way of, shall we say, diminishing after a tutorial finishes.

bradjc:
I can't even build with cmake :(

```
./create_openthread-tock.sh
-- The C compiler identification is GNU 13.2.0
-- The CXX compiler identification is GNU 13.2.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - failed
-- Check for working C compiler: /usr/local/bin/arm-none-eabi-gcc
-- Check for working C compiler: /usr/local/bin/arm-none-eabi-gcc - broken
CMake Error at /usr/local/Cellar/cmake/3.25.2/share/cmake/Modules/CMakeTestCCompiler.cmake:70 (message):
  The C compiler

    "/usr/local/bin/arm-none-eabi-gcc"

  is not able to compile a simple test program.

  It fails with the following output:

    Change Dir: /Users/bradjc/temp/libtock-c/openthread-tock/build/CMakeFiles/CMakeScratch/TryCompile-Ga8ybZ

    Run Build Command(s):/usr/bin/make -f Makefile cmTC_dd9b9/fast && /Library/Developer/CommandLineTools/usr/bin/make  -f CMakeFiles/cmTC_dd9b9.dir/build.make CMakeFiles/cmTC_dd9b9.dir/build
    Building C object CMakeFiles/cmTC_dd9b9.dir/testCCompiler.c.obj
    /usr/local/bin/arm-none-eabi-gcc   -frecord-gcc-switches -gdwarf-2 -Os -fdata-sections -ffunction-sections -fstack-usage -D_FORTIFY_SOURCE=2 -fPIC -mthumb -mfloat-abi=soft -msingle-pic-base -mpic-register=r9 -mno-pic-data-is-text-relative -mcpu=cortex-m4 -isystem ../lib/libtock-newlib-4.2.0.20211231/arm/arm-none-eabi/include -isystem ../lib/libtock-libc++-10.5.0/arm/arm-none-eabi/include/c++/10.5.0 -isystem ../lib/libtock-libc++-10.5.0/arm/arm-none-eabi/include/c++/10.5.0/arm-none-eabi -std=gnu11  -o CMakeFiles/cmTC_dd9b9.dir/testCCompiler.c.obj -c /Users/bradjc/temp/libtock-c/openthread-tock/build/CMakeFiles/CMakeScratch/TryCompile-Ga8ybZ/testCCompiler.c
    Linking C executable cmTC_dd9b9
    /usr/local/Cellar/cmake/3.25.2/bin/cmake -E cmake_link_script CMakeFiles/cmTC_dd9b9.dir/link.txt --verbose=1
    /usr/local/bin/arm-none-eabi-gcc -frecord-gcc-switches -gdwarf-2 -Os -fdata-sections -ffunction-sections -fstack-usage -D_FORTIFY_SOURCE=2 -fPIC -mthumb -mfloat-abi=soft -msingle-pic-base -mpic-register=r9 -mno-pic-data-is-text-relative -mcpu=cortex-m4 -isystem ../lib/libtock-newlib-4.2.0.20211231/arm/arm-none-eabi/include -isystem ../lib/libtock-libc++-10.5.0/arm/arm-none-eabi/include/c++/10.5.0 -isystem ../lib/libtock-libc++-10.5.0/arm/arm-none-eabi/include/c++/10.5.0/arm-none-eabi -std=gnu11  -frecord-gcc-switches -gdwarf-2 -Os -fdata-sections -ffunction-sections -fstack-usage -D_FORTIFY_SOURCE=2 -fPIC -mthumb -mfloat-abi=soft -msingle-pic-base -mpic-register=r9 -mno-pic-data-is-text-relative -mcpu=cortex-m4 -isystem ../lib/libtock-newlib-4.2.0.20211231/arm/arm-none-eabi/include -isystem ../lib/libtock-libc++-10.5.0/arm/arm-none-eabi/include/c++/10.5.0 -isystem ../lib/libtock-libc++-10.5.0/arm/arm-none-eabi/include/c++/10.5.0/arm-none-eabi -specs=nano.specs -specs=nosys.specs  CMakeFiles/cmTC_dd9b9.dir/testCCompiler.c.obj -o cmTC_dd9b9
    arm-none-eabi-gcc: fatal error: cannot read spec file 'nano.specs': No such file or directory
    compilation terminated.
    make[2]: *** [cmTC_dd9b9] Error 1
    make[1]: *** [cmTC_dd9b9/fast] Error 2





  CMake will not be able to correctly generate this project.
Call Stack (most recent call first):
  CMakeLists.txt:34 (project)


-- Configuring incomplete, errors occurred!
See also "/Users/bradjc/temp/libtock-c/openthread-tock/build/CMakeFiles/CMakeOutput.log".
See also "/Users/bradjc/temp/libtock-c/openthread-tock/build/CMakeFiles/CMakeError.log".
cp: build/openthread/src/core/libopenthread-mtd.a: No such file or directory
make: *** [all] Error 1
```

tyler-potyondy:
> I can't even build with cmake :(

That's strange. Let me look into this....

tyler-potyondy:
A couple of updates....

- Following up on the CMake issues, it seems this is an issue on MacOS with the arm-gcc provided. @bradjc was able to fix this locally, but we will need a long term fix before merging this.
- The most recent push brings this branch up to date with the current master branch in Tock. I need to finish going through and incorporating the remainder of the feedback and comments above.

alistair23:
> > I would like to propose we compile openthread in advance and provide pre-compiled binaries (like we do with newlib/libcpp/picolib). I do not think we can provide a quality user experience with our current build system (and interfacing with openthread's build system) to compile the library at the same time as compiling apps. Thoughts?

I think that is a bad idea. I feel like the xz drama clearly shows why. It's difficult to audit binary blobs that we ship to users.

> 
> My preference is to use git submodules here, but I understand your frustrations / concerns @bradjc. In my opinion, the submodule is nice for openthread since the code base will likely not be altered and it serves as an explicit pointer to the detached head at the commit we set in the `libtock-c` repo. Additionally, the submodule in my opinion is more explicit for people navigating through the codebase in github as a pointer to the openthread repo.

+1 Submodules are clunky, but they are commonly used and industry standard. As you say it makes it easy to track future changes and does explicitly show we are using the openthread repo and the exact commit.

> 
> If the consensus is for using a zip over the submodule, this would not be an issue as we could obtain a zip of the openthread repo. However, I think using an entirely precompiled binary would be a mistake. The biggest downside I see to precompiled binaries surrounds customization of openthread. OpenThread provides a high degree of customizable build flags that can be specified in config header files and also as CMake arguments. Using a precompiled binary pigeonholes us to our specific build configuration for OpenThread. A good example of this is the build flag for logging and whether to compile the Full Thread Device vs Minimal Thread Device.

+1

bradjc:
> > > I would like to propose we compile openthread in advance and provide pre-compiled binaries (like we do with newlib/libcpp/picolib). I do not think we can provide a quality user experience with our current build system (and interfacing with openthread's build system) to compile the library at the same time as compiling apps. Thoughts?
> 
> I think that is a bad idea. I feel like the xz drama clearly shows why. It's difficult to audit binary blobs that we ship to users.

It's a very advanced attack, but I doubt the defense is expecting end users to audit source code. I also doubt that debian/ubuntu and every other app store-like environment are going to stop sending compiled binaries to their users. I think we may be able to learn from the xz attack, but I think we have to take on the challenge, not defer to users to audit code.

alistair23:
I maybe should have been more clear.

I don't think everyone is going to switch to Gentoo.

I do think it shines a light on random developers committing non-reproducible binary blobs. I suspect users will start to expect a bit more then "I built this on my random development machine and now you are running it, trust me it's fine, I provided a SHA as proof".

Building something on the Debian infrastructure as part of a reproducible build system and signing it with a Debian trusted key isn't really the same as someone uploading some build artifacts to libtock-c.

alistair23:
> I maybe should have been more clear.
> 
> I don't think everyone is going to switch to Gentoo.
> 
> I do think it shines a light on random developers committing non-reproducible binary blobs. I suspect users will start to expect a bit more then "I built this on my random development machine and now you are running it, trust me it's fine, I provided a SHA as proof".
> 
> Building something on the Debian infrastructure as part of a reproducible build system and signing it with a Debian trusted key isn't really the same as someone uploading some build artifacts to libtock-c.

It's not just @lschuermann  and I who think this either

```
It has now alerted the entire Linux community to this new type of attack. We now know that binary blobs even in test-files (which aren't shipped to end users) can still infect a project, if the groundwork has been done in the project's open source release-scripts, such as in this case. Projects can now start demanding that every binary blob added to any project must be reproducible from scratch.
```
https://www.reddit.com/r/linux/comments/1bvdjgs/xz_shows_our_strength/

bradjc:
Admittedly this packaging discussion has been a bit of a moving target as I've worked hard to figure out what is possible. I now believe we can build openthread with our existing libtock-c build system (branch: https://github.com/tock/libtock-c/compare/master...ot-bradjc), so we do not need to precompile libraries for our supported platforms.

I do ask that going forward we all keep our users, specifically new users in mind. Attacks are shiny, but we can't ignore our user base and what supports them and what ensures they have a good experience using Tock (and for the first time). This necessarily leads to tradeoffs which must be balanced.

bradjc:
Todos:

1. Update makefiles and build structure from https://github.com/tock/libtock-c/tree/ot-bradjc
1. Need to actually figure out what flags we need set. I included some in the makefile while debugging, which may be redundant or unwanted. Should document which flags we are setting in the makefile(s).
2. I had to hack around the cortex-m0 for mbedtls. I tried to make it clear how I modified the header file. I think it should be possible to use #undef and include the version from openthread instead, and then unset the asm flag on thumb1.
3. I think we should create a new helper makefile in libtock-c/openthread that apps can include which setup all of the EXTERN_LIBS needed for openthread. This means we don’t have to copy the list to every ot app. We also can’t use the makefile.app because the extern_libs variable is not handled recursively.
4. Rename folder to just `openthread` (or I suppose `libopenthread` would also be consistent).
5. Simplify the platform directory, remove empty/unneeded files, remove folder nesting.
6. Rename from nrf52840 to tock in the platform library.

tyler-potyondy:
Most recent push completes todos 5,6,7.

> 
> - [x]  Rename folder to just openthread (or I suppose libopenthread would also be consistent).
> - [x]  Simplify the platform directory, remove empty/unneeded files, remove folder nesting.
> - [x]  Rename from nrf52840 to tock in the platform library.

tyler-potyondy:
Testing https://github.com/tock/libtock-c/tree/ot-bradjc the openthread library compiles using the new build system, but the flashed app does not operate correctly (i.e. unable to join thread network). I'll look more into what is causing this, but we will need to resolve this issue before incorporating the new build system here.

bradjc:
I've updated ot-bradjc with including the correct config file and refactoring how we modify the mbedtls-config.

tyler-potyondy:
@bradjc I've merged  your changes into my local version of this branch and OpenThread is working properly. I see that some of the ot-bradjc branch changes are to the TockLibrary makefiles. Should I block on pushing a merged version here until we incorporate the Tocklibrary changes into master?

tyler-potyondy:
I merged the changes from the ot-bradjc branch (the non openthread commits from brad should all be already merged into master). As is, this will fail CI since we have a warning for `stack-usage` being unbounded in `libtcplp`. I could not find a way to disable this warning.

libopenthread/platform/Makefile
```diff
@@ -0,0 +1,17 @@
...
+override CFLAGS += -I$(TOCK_USERLAND_BASE_DIR)/libopenthread/openthread/include
+override CFLAGS += -I$(TOCK_USERLAND_BASE_DIR)/libopenthread/openthread/src/core
+override CFLAGS += -I$(TOCK_USERLAND_BASE_DIR)/libtock
+
+$(LIBNAME)_DIR := $(TOCK_USERLAND_BASE_DIR)/libopenthread/platform
+$(LIBNAME)_SRC_ROOT := $(TOCK_USERLAND_BASE_DIR)/libopenthread/platform
+$(LIBNAME)_SRCS :=				\
+	$(wildcard $($(LIBNAME)_DIR)/*.c)	\
+	$(wildcard $($(LIBNAME)_DIR)/*.cxx)	\
+	$(wildcard $($(LIBNAME)_DIR)/*.s)
```
bradjc, line 15:
```suggestion
$(LIBNAME)_SRCS := $(wildcard $($(LIBNAME)_SRC_ROOT)/*.c)
```

bradjc:
Ok the last piece on the build side is the controversial #383. That will enable the libraries to automatically compile when trying to compile the app.

libopenthread/Makefile
bradjc, line 1: This is fairly fiendish; running make in the libopenthread directory deletes all built output. Do we need this Makefile for some reason?

libopenthread/README.md
bradjc, line 1: This file is out of date

libopenthread/libmbedtls/Makefile
```diff
@@ -0,0 +1,29 @@
...
+override CPPFLAGS += -I$($(LIBNAME)_SRC_ROOT)
+override CPPFLAGS += -I$($(LIBNAME)_DIR)
+
+# Include our standard OpenThread configuration to ensure all flags are set
+# consistently.
+override CPPFLAGS += -I$(TOCK_USERLAND_BASE_DIR)/libopenthread/platform
+override CPPFLAGS += -DOPENTHREAD_PLATFORM_CORE_CONFIG_FILE=\"openthread-core-tock-config.h\"
+
+# Include our config file for mbedtls which makes one minor modification.
+override CPPFLAGS += -DMBEDTLS_CONFIG_FILE=\"libtock-mbedtls-config.h\"
```
bradjc, line 22: We should probably use `CPPFLAGS_$(LIBNAME)` for these since we have the variable.

libopenthread/libmbedtls/Makefile.app
bradjc, line 1: Should just be includes directories.

libopenthread/libopenthread-ftd/Makefile
```diff
@@ -0,0 +1,39 @@
...
+override CPPFLAGS += -I$($(LIBNAME)_SRC_ROOT)/src/core
+override CPPFLAGS += -I$($(LIBNAME)_SRC_ROOT)/include
+override CPPFLAGS += -I$($(LIBNAME)_SRC_ROOT)/third_party/mbedtls/repo/include
+override CPPFLAGS += -I$($(LIBNAME)_SRC_ROOT)/third_party/mbedtls
+
+override CPPFLAGS += -DPACKAGE_NAME=\"tock-openthread-ftd\"
+override CPPFLAGS += -DPACKAGE_VERSION=\"0.1\"
+override CPPFLAGS += -DOPENTHREAD_FTD=1
+override CPPFLAGS += -DOPENTHREAD_PLATFORM_CORE_CONFIG_FILE=\"openthread-core-tock-config.h\"
+override CPPFLAGS += -DMBEDTLS_CONFIG_FILE=\"mbedtls-config.h\"
```
bradjc, line 33: `CPPFLAGS_$(LIBNAME)`

libopenthread/libopenthread-mtd/Makefile
```diff
@@ -0,0 +1,50 @@
...
+override CPPFLAGS += -I$(TOCK_USERLAND_BASE_DIR)/libopenthread/platform
+
+# Package settings for the Tock version of OpenThread.
+override CPPFLAGS += -DPACKAGE_NAME=\"tock-openthread-mtd\"
+override CPPFLAGS += -DPACKAGE_VERSION=\"0.1\"
+# Specify we want a minimal thread device build of the library.
+override CPPFLAGS += -DOPENTHREAD_MTD=1
+# Configuration header files we need to specify.
+override CPPFLAGS += -DOPENTHREAD_PLATFORM_CORE_CONFIG_FILE=\"openthread-core-tock-config.h\"
+override CPPFLAGS += -DMBEDTLS_CONFIG_FILE=\"mbedtls-config.h\"
```
bradjc, line 38: `CPPFLAGS_$(LIBNAME)`

libopenthread/platform/radio.c
```diff
@@ -0,0 +1,279 @@
...
+  OT_UNUSED_VARIABLE(aInstance);
+  aIeeeEui64[0] = 0xf4;
+  aIeeeEui64[1] = 0xce;
+  aIeeeEui64[2] = 0x36;
+  aIeeeEui64[3] = 0x67;
+  aIeeeEui64[4] = 0x13;
+  aIeeeEui64[5] = 0x12;
+  aIeeeEui64[6] = 0x3f;
+  aIeeeEui64[7] = 0xa6;
+}
```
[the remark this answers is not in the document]
tyler-potyondy: I added eui64 support in the `libtock-folders` branch since I wrote the eui64 syscall in the new format.

libopenthread/Makefile
tyler-potyondy: Having make delete the build directory is rather silly. I'll go ahead and update this!

tyler-potyondy:
@bradjc does the revised libtock build system support this now?

>  I think we should create a new helper makefile in libtock-c/openthread that apps can include which setup all of the EXTERN_LIBS needed for openthread. This means we don’t have to copy the list to every ot app. We also can’t use the makefile.app because the extern_libs variable is not handled recursively.

bradjc:
> @bradjc does the revised libtock build system support this now?
> 
> > I think we should create a new helper makefile in libtock-c/openthread that apps can include which setup all of the EXTERN_LIBS needed for openthread. This means we don’t have to copy the list to every ot app. We also can’t use the makefile.app because the extern_libs variable is not handled recursively.

It's not really something the build system needs to support, we just need to make the makefile(s).

examples/openthread/hello/Makefile
```diff
@@ -0,0 +1,17 @@
...
+
+# Specify this directory relative to the current application.
+TOCK_USERLAND_BASE_DIR = ../../..
+
+# Specify this app depends on the MTD OpenThread library.
+include $(TOCK_USERLAND_BASE_DIR)/libopenthread/libopenthread-mtd.mk
+
+# Arbitrarily set this for now. We will need to revisit these values.
+#APP_HEAP_SIZE:=5000
+STACK_SIZE:=5000
```
bradjc, line 11: What should these be?

examples/openthread/hello/README.md
```diff
@@ -0,0 +1,231 @@
...
+and subsequently times out) after ~7 minutes. This bug has been traced to an issue with the alarm
+implementation but has not been resolved.
+- The 6LoWPAN capsule on occassion panics (this behavior is observed semi frequently)
+- The code size is quite large as is for the current implementation. There are numerous 
+opportunities to reduce this (use hardware crypto instead of mbed crypto library).
+- The openthread_hello `Makefile` will build all needed OpenThread and platform dependencies. 
+However, the `make` build system, as is, does not rebuild the platform or OpenThread library 
+if changes are made to either the platform or openthread directories.
+- Channel switching is not yet implemented
+- The RSSI value is hard coded
```
bradjc, line 41: I think these are all in the process of being resolved?

libopenthread/Makefile
```diff
@@ -0,0 +1,23 @@
...
+mtd: dependencies
+	make -C libopenthread-mtd
+
+dependencies:
+	make -C libmbedtls
+	make -C libtcplp
+	make -C platform
+
+subcmd:
+	echo "hi"
```
bradjc, line 16: ?

libopenthread/Makefile
```diff
@@ -0,0 +1,23 @@
+.PHONY: clean default
+default:
```
bradjc, line 2: I haven't tested using this makefile, but it should just build everything if you type `make`.

libopenthread/libmbedtls/Makefile.setup
bradjc, line 1: All of these Makefile.setup files need to be replaced with something that just checks if the submodule is checked out, and if not, checks out the submodule.

libopenthread/platform/plat.h
```diff
@@ -0,0 +1,12 @@
+#ifndef PLAT_H
+#define PLAT_H
```
bradjc, line 2: Just use `#pragma once`

libopenthread/platform/plat.h
```diff
@@ -0,0 +1,12 @@
...
+#include<stdbool.h>
+
+typedef struct {
+    char *buffer;
+    int write_index;
+    int read_index;
+    bool new;
+} ring_buffer;
```
bradjc, line 10: What is this for?

libopenthread/Makefile
tyler-potyondy: Artifact from testing that should have been removed.

examples/openthread/hello/Makefile
tyler-potyondy: I picked a larger than the default value for testing somewhat arbitrarily. For our version we upstream, I'm not sure the best way determine what these should be. @bradjc do you have thoughts on a good way to determine what this value should be?

libopenthread/Makefile
tyler-potyondy: Only reason this is empty was so that you explicitly specify `make mtd` or `make ftd`. Perhaps printing an error message prompting to run `make mtd` or `make ftd` would suffice?

bradjc: Why not just build both?

libopenthread/platform/settings.c
```diff
@@ -0,0 +1,63 @@
+#include <openthread/platform/settings.h>
+
+// Our current approach is to use the flash.c API rather than the settings flash API (this
+// is configured in openthread-system.h). This file is left for now as a place holder.
```
bradjc, line 4: Is this file actually needed?

examples/openthread/hello/README.md
tyler-potyondy: Yes, with the exception of channel switching.

LQI/RSSI should be merged soon (https://github.com/tock/tock/pull/3972).
6LoWPAN bug is resolved.
The disconnect bug is mostly fixed (blocking on https://github.com/tock/tock/pull/3973)

Channel switching for now is not implemented (and in my opinion should not be a blocker on this PR. This is something we can add support for in the 15.4 capsule rewrite)

libopenthread/platform/settings.c
tyler-potyondy: No it is not. I will remove this.

libopenthread/platform/alarm.c
```diff
@@ -0,0 +1,66 @@
...
+
+void otPlatAlarmMilliStop(otInstance *aInstance) {
+  OT_UNUSED_VARIABLE(aInstance);
+  alarm_cancel(&alarm);
+}
+
+uint32_t otPlatAlarmMilliGetNow(void) {
+  // TODO: there is a bug in this implementation causing the timer value to 
+  // wrap around 500,000 milliseconds. This may be an issue with the 
+  // `gettimeasticks` function or in the logic here.
```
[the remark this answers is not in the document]
tyler-potyondy: This fix is blocking on (https://github.com/tock/tock/pull/3973)

libopenthread/platform/plat.h
tyler-potyondy: This is being used to solve the race conditions and problems we faced with receive buffers and the receive upcall. 

There is now a ring buffer in the capsule and also a ring buffer in the openthread app. The app's ring buffer performs a sync copy from the allowed kernel ring buffer and completes the entire operation before passing the packet to openthread (some of openthread's operations may yield which was causing numerous issues with handling packets being passed to the process). 

This userprocess ring buffer could be removed in the future with a kernel allow ring buffer syscall.

libopenthread/README.md
```diff
@@ -0,0 +1,17 @@
...
+- libmbedtls.a
+- libtcplp.a
+
+There are two methods to compile the `libopenthread` library files:
+- Include the libopenthread static libraries in the app's Makefile (see `examples/openthread/hello/Makefile`). Upon running `make` in the app directory, libopenthread will be built.
+- Run `make ftd` or `make mtd` in the `libtock-c/libopenthread`
+
+Given the size of the OpenThread library, it may be beneficial to only build the library for your needed target as opposed to the default libtock-c option of building all architectures. To specify this, add `TOCK_TARGETS=cortex-m4` as an argument when invoking `make` or to the Makefile.
+
+For now, `libopenthread` is only supported for the nrf52840dk board.
```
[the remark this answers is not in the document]
tyler-potyondy: I have removed this comments since the build system now supports all platforms (although not all platforms may have a radio etc).

tyler-potyondy:
Update on the state of the PR... 

@bradjc I believe I have addressed all outstanding feedback and todos with the exception of:
- timer bug (blocking on https://github.com/tock/tock/pull/3973 and updating openthread logic accordingly)
-  eui64 support (#409)
- lqi/rssi (https://github.com/tock/tock/pull/3972)

libopenthread/platform/system.c
```diff
@@ -0,0 +1,180 @@
...
+
+  int res = ieee802154_receive(rx_callback, otTockInstance.kernel_rx_buf, NULL);
+
+  otError result = OT_ERROR_NONE;
+
+  if (res != 0) {
+    result = OT_ERROR_FAILED;
+  }
+
+  return result;
```
bradjc, line 179:
```suggestion
  if (res != RETURNCODE_SUCCESS) return OT_ERROR_FAILED;

  return OT_ERROR_NONE;
```

bradjc (changes requested):

libopenthread/platform/radio.c
```diff
@@ -0,0 +1,279 @@
...
+  // convert aExtAddress to be big endian in temp variable
+  otExtAddress temp;
+  for (int i = 0; i < 8; i++) {
+    temp.m8[i] = aExtAddress->m8[7 - i];
+  }
+
+  int retCode = ieee802154_set_address_long((unsigned char*) &temp);
+  assert(retCode == 0);
+
+  if (retCode != 0) printf("Error setting long address");
```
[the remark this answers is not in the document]
tyler-potyondy: This can be removed.

libopenthread/Makefile
```diff
@@ -0,0 +1,27 @@
...
+
+ftd: dependencies
+	make -C libopenthread-ftd
+
+mtd: dependencies
+	make -C libopenthread-mtd
+
+# confirm openthread git submodule is checked out.
+$(TOCK_USERLAND_BASE_DIR)/libopenthread/openthread/README.md:
+	git submodule update --init -- openthread
```
bradjc, line 15: I'm fairly sure we can remove this and just rely on the `Makefile.setup` files.

tyler-potyondy:
The `alarm.c` file in the most recent commit was outdated. Updating now and will push soon.

tyler-potyondy:
The most recent commits add the logic needed to resolve the alarm overflow bug leading to the device dropping off the network at 7 minutes on the nordic boards. 

This will be blocking on the left padding/left justified alarm capsule PRs. Testing with https://github.com/tock/tock/pull/3973, this bug is now resolved.

tyler-potyondy:
TODOs remaining:

- LQI (https://github.com/tock/tock/pull/3972) - will add this to the OpenThread app once this PR is merged.
- EUI64 (https://github.com/tock/libtock-c/pull/409) @bradjc I believe your feedback is addressed, unless I am still missing some details. I will remove the hardcoded values and update accordingly once merged.
- Separate apps as mentioned above. 

In regards to separating apps, there are helper functions I defined that will be used across openthread apps. Do we want to place these helper functions in `libopenthread` as `libtock_openthread_*`?

libopenthread/platform/radio.c
```diff
@@ -0,0 +1,278 @@
+#include <ieee802154.h>
+#include <openthread-system.h>
+#include <openthread/platform/radio.h>
+#include <plat.h>
+#include <stdio.h>
```
bradjc, line 5:
```suggestion
#include <stdio.h>

#include <openthread-system.h>
#include <openthread/platform/radio.h>

#include <ieee802154.h>

#include "plat.h"
```

libopenthread/platform/system.c
```diff
@@ -0,0 +1,208 @@
+#include <ieee802154.h>
+#include <openthread-system.h>
+#include <openthread/instance.h>
+#include <openthread/platform/alarm-milli.h>
+#include <openthread/platform/radio.h>
+#include <plat.h>
+#include <stdio.h>
```
bradjc, line 7:
```suggestion
#include <stdio.h>

#include <openthread-system.h>
#include <openthread/instance.h>
#include <openthread/platform/alarm-milli.h>
#include <openthread/platform/radio.h>

#include <ieee802154.h>

#include "plat.h"

```

bradjc (changes requested):
Apps, readme, libtock-c rewrite

tyler-potyondy:
Rebased to the current master to hopefully make the libtock rewrite merging easier. EUI64 is now added, apps are separated, and READMEs are updated.

libopenthread/platform/radio.c
```diff
@@ -0,0 +1,281 @@
...
+  // TODO HARDCODED FOR NOW
+  OT_UNUSED_VARIABLE(aInstance);
+  aIeeeEui64[0] = 0xf4;
+  aIeeeEui64[1] = 0xce;
+  aIeeeEui64[2] = 0x36;
+  aIeeeEui64[3] = 0x67;
+  aIeeeEui64[4] = 0x13;
+  aIeeeEui64[5] = 0x12;
+  aIeeeEui64[6] = 0x3f;
+  aIeeeEui64[7] = 0xa6;
```
bradjc, line 35: Not dynamic?

libtock/services/alarm.c
bradjc, line 1: ?

libopenthread/platform/radio.c
tyler-potyondy: This change got lost in the rebase/merge. Updated now.

libtock/services/alarm.c
tyler-potyondy: Same as above.

bradjc (dismissed):

bradjc (approved):

PR #383, opened 2024-03-30
make: support automatically building libraries when compiling apps

bradjc:
This adds a `Makefile.setup`, which, if provided, is used by `AppMakefile` to fetch and then build the library for an app that requires it.

The key is this in `AppMakefile`:

```make
# Rules to call library makefiles to build required libraries. We use a group target
# dependent on all library .a files.
#
# We also support an optional Makefile.setup which is responsible for retrieving
# the source used to build the library.
#
# Arguments:
# - $(1): The path to the build directory.
# - $(2): The path to the library.
define EXTERN_LIB_BUILD_RULE

ifneq "$$(wildcard $(2)/Makefile.setup)" ""
$(1):
	# Since a Makefile.setup exists, do any setup steps needed to fetch the
	# library.
	$$(MAKE) -C $(2) -f Makefile.setup all
	$$(MAKE) -C $(2) -f Makefile all
else
$(1):
	$$(MAKE) -C $(2) -f Makefile all
endif

endef
```

which runs `make Makefile.setup` and then `make` in the library directory if the build directory does not exist. The setup script downloads the library source. The regular makefile builds it. These are split because the regular build makefile needs all of the source files to be present to create the build rules.

### DIFF
--- a/AppMakefile.mk
+++ b/AppMakefile.mk
@@ -51,11 +51,11 @@ ifneq "$$(wildcard $(2)/Makefile.setup)" ""
 $(1):
 	# Since a Makefile.setup exists, do any setup steps needed to fetch the
 	# library.
-	$(MAKE) -C $(2) -f Makefile.setup all
-	$(MAKE) -C $(2) -f Makefile all
+	$$(MAKE) -C $(2) -f Makefile.setup all
+	$$(MAKE) -C $(2) -f Makefile all
 else
 $(1):
-	$(MAKE) -C $(2) -f Makefile all
+	$$(MAKE) -C $(2) -f Makefile all
 endif
 
 endef
--- a/AppMakefile.mk
+++ b/AppMakefile.mk
@@ -36,6 +36,43 @@ include $(TOCK_USERLAND_BASE_DIR)/libtock/Makefile
 include $(TOCK_USERLAND_BASE_DIR)/Program.mk
 
 
+# Rules to call library makefiles to build required libraries. We use the build
+# directory as the indicator if the library exists or not.
+#
+# We also support an optional Makefile.setup which is responsible for retrieving
+# the source used to build the library.
+#
+# Arguments:
+# - $(1): The path to the build directory.
+# - $(2): The path to the library.
+define EXTERN_LIB_BUILD_RULE
+
+ifneq "$$(wildcard $(2)/Makefile.setup)" ""
+$(1):
+	# Since a Makefile.setup exists, do any setup steps needed to fetch the
+	# library.
+	$(MAKE) -C $(2) -f Makefile.setup all
+	$(MAKE) -C $(2) -f Makefile all
+else
+$(1):
+	$(MAKE) -C $(2) -f Makefile all
+endif
+
+endef
+
+# Rule to build each architecture-specific library archive (.a file). To ensure
+# each library only has a single build rule, we just make each archive dependent
+# on the build directory itself. Creating the build directory is responsible for
+# building the entire library.
+#
+# Arguments:
+# - $(1): The path to the built library .a file.
+# - $(2): The path to the build directory for the library.
+define EXTERN_LIB_BUILD_RULES
+
+$(1): | $(2)
+
+endef
 
 # Rules to incorporate external libraries
 define EXTERN_LIB_RULES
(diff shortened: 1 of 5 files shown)


alistair23 (approved):

bradjc:
I'm more convinced we need this.

https://github.com/tock/libtock-c/blob/1f5621438940635d04efa26a803ec9665ae59cbf/AppMakefile.mk#L31-L33

is wrong, and should be:

```make
# By default, have all apps depend on libtock.
EXTERN_LIBS += $(TOCK_USERLAND_BASE_DIR)/libtock
```

The current approach only works with a single library because we can't include TockLibrary.mk more than once.

AppMakefile.mk
```diff
@@ -45,6 +45,15 @@ include $(TOCK_USERLAND_BASE_DIR)/Program.mk
...
 # Rules to incorporate external libraries
 define EXTERN_LIB_RULES
+
+# If there is a Makefile.setup, then we try to automatically build the library.
+ifneq "$$(wildcard $(1)/Makefile.setup)" ""
+  # Do any setup steps needed to fetch the library.
+  DEVNULL := $$(shell make -C $(1) -f Makefile.setup all)
+  # Build the library if needed.
+  DEVNULL := $$(shell make -C $(1) all)
+endif
```
ppannuto, line 55: This is a red flag. Generally, our build system tries to follow the ethos of [recursive make considered harmful](https://accu.org/journals/overload/14/71/miller_2004/). Everything else carefully stitches together rules so that there is only ever one invocation of the `make` program which knows the full build.

I don't see anything in the `u8g2` example with this PR that couldn't be done as rules. Probably something like have included makefiles add their final targets to a variable like `PREBUILD_DEPENDENCIES`, and then add `PREBUILD_DEPENDENCIES` as another dependency roughly everywhere the `BUILD_DIR` dependency already is.

ppannuto (changes requested):

AppMakefile.mk
bradjc: One issue:
1. We need to download a .zip file.
2. That .zip file contains the c files that are picked up by the wildcard.

Is there a way to populate the SRCs variable after a recipe has run?

bradjc: Second issue: we need to build N libraries. We only have one TockLibrary makefile. That makefile uses `$(LIBNAME)`, which can only be set once.

bradjc: However, `$(LIBNAME)` is used in almost every line in that makefile so it's probably gonna get gnarly.

ppannuto: While I haven't tested it in a while, the multiple library name issue should already be addressed, that's what https://github.com/tock/libtock-c/blob/master/TockLibrary.mk#L134-L148 is about.

`$(LIBNAME)` is only used while `TockLibrary.mk` is generating rules, so if you redefine `LIBNAME` (must use `:=`) and include `TockLibrary.mk` again, things should work.

bradjc: _Should_, but only if the second library contains no files with the same name as the first library.

ppannuto: > Is there a way to populate the SRCs variable after a recipe has run?

This one is a bigger and more fundamental issue. Make expects to first fully define the dependency graph and then start resolving it. It does not, by design, allow for the dependency graph to be modified once it starts running rules.

---

A _hack_ of an answer would be something like

```
LIBNAME_SRCS := $(shell test -d unzipped_dir || (wget ZIPFILE && unzip ZIPFILE) && echo unzipped_dir/*.c)
```

ppannuto: This is why all the other libraries that we're interested in building at runtime are submodules; lets version control solve the "get the source files there before running make" problem.

Probably best to do the same for `u8g2`, _or_ to simply commit pre-compiled objects like we do for newlib/etc.

bradjc:
>This is why all the other libraries that we're interested in building at runtime are submodules; lets version control solve the "get the source files there before running make" problem.

What if you want to download libtock-c without git? Or who checks out the submodule?

ppannuto:
> What if you want to download libtock-c without git? Or who checks out the submodule?

I don't think we have a mechanism to do that currently. But if we added one, I think the generated `libtock-c-srcs.zip` or whatever would have to include all the submodule sources as well. (i.e., the tool that takes `libtock-c` out of git is responsible for making sure it takes everything out of git, including the submodules).

bradjc:
Well, we do have this:
image

But I still don't understand who ensures you have the submodule checked out and to the correct hash.

Would it be more kosher to just have a bash script, ala the `fetch_*` scripts for prebuilt libraries?

bradjc:
This gets messy because we use all these global variables (eg CFLAGS) everywhere. But I think this is possible.

I have to say, however, that running separate instances of make sure does seem like the better option. E.g., what if two libraries need the same #define flag, but with different values?

bradjc:
Updated version here: https://github.com/tock/libtock-c/tree/make-tocklib-paths-compile-libs

This PR is blocked on #391

bradjc:
Ok here is my next attempt. I'll try to recap where we are:

**Goal:** Enable a two step process: 1) clone libtock-c, and 2) run make in an app, and everything just works. Also, without committing or providing built output.

Various options:
- Include all of the rules to build the library in the app build. This is elegant, but has two issues. First, we end up trying to format the source code for the libraries. Second, we end up with a single list of `CFLAGS` (and others) for building everything (all libraries and the app).
- Call `make` from make to build the library as a separate make execution. Fixes above issues, but isn't as clean. This is the approach in this PR. Another drawback is we aren't tracking dependencies correctly. So changes in the library won't trigger a rebuild.

bradjc:
Whatever magic exists in #405 that tells make to run the fetch script only once for a list of targets, does _not_ seem to work if the rule is generated using `$eval` as in https://github.com/tock/libtock-c/pull/383/commits/c67be382eb4b32f62b3832d3165e57313945a62e.

bradjc:
Ok! This builds again.

ppannuto (approved):
I tested this for a bunch of permutations of partial builds, parallelism, etc tonight and couldn't get anything to break. The dependency generation all makes sense to me as well. Sorry to push on this a few times, but I think this is much more robust now.

PR #387, opened 2024-04-01
Dynamic App Loader Helper

viswajith-g:
This PR introduces the helper app to validate the kernel's dynamic process loading functionality ([tock/tock#3941](https://github.com/tock/tock/pull/3941)). The helper app triggers the dynamic process loading with a button press. 

This application is located in `/examples/tests/app_loader` and was tested on the nRF52840DK. A previous version of the dynamic process loader was tested with the same application on an Imix board as well.


examples/ota_app_wip/main.c
```diff
@@ -0,0 +1,5246 @@
...
+  0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0,
+  0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0,
+  0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0,
+  0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0,
+  0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0,
+  0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0,
+  0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0,
+  0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0};
+
+const uint8_t bleapp_bin[] = {0x2, 0x0, 0x40, 0x0, 0x0, 0x0, 0x1, 0x0, 0x1, 0x0, 0x0, 0x0, 0x0C, 0x5A, 0x21, 0x7A,
```
alistair23, line 282: This should probably be split out at least into it's own file. GitHub is struggling just reading this :)

alistair23 (changes requested):

examples/ota_app_wip/main.c
viswajith-g: These files seem to have been updated when I ran `make prepush` from the stable branch. This is not my code. 

The code is located [here](examples/tests/app_loader/main.c).

examples/ota_app_wip/main.c
```diff
@@ -0,0 +1,5246 @@
+#include <button.h>
```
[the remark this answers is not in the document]
viswajith-g: This is not my code. It was formatted when I ran `make prepush`.

examples/ota_app_wip/main.c
```diff
@@ -0,0 +1,5246 @@
...
+#include <internal/ota.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <string.h>
+#include <timer.h>
+#include <unistd.h>
+
+#define Yes 0
```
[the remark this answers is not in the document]
viswajith-g: Also not my code.

examples/ota_app_wip/main.c
alistair23: I'm not sure what you mean, I can see you added this in this code in this PR

You should squash your commits to remove this if you just want to remove it later

libtock/internal/app_loader.h
```diff
@@ -0,0 +1,99 @@
...
+
+/*
+ * Function to enable interrupts for the specified button number.
+ */
+int button_enable_interrupt(int button_num);
+
+/*
+ * Function to disable interrupts for the specified button number.
+ */
+int button_disable_interrupt(int button_num);
```
alistair23, line 82: This doesn't seem like the right place for this to be

libtock/internal/app_loader.c
```diff
@@ -0,0 +1,60 @@
...
+
+int button_count(int* count) {
+  syscall_return_t res = command(DRIVER_NUM_BUTTON, 0, 0, 0);
+  return tock_command_return_u32_to_returncode(res, (uint32_t*) count);
+}
+
+int button_enable_interrupt(int button_num) {
+  syscall_return_t res = command(DRIVER_NUM_BUTTON, 1, button_num, 0);
+  return tock_command_return_novalue_to_returncode(res);
+}
```
alistair23, line 50: This should be it's own file

examples/tests/app_loader/main.c
```diff
@@ -0,0 +1,298 @@
...
+#include <tock.h>
+
+#define FLASH_BUFFER_SIZE 512
+#define RETURNCODE_SUCCESS 0
+
+static bool done  = false;    // to check if writing to flash is done
+uint32_t app_size = 0;        // variable to store app size
+
+// Tock Application Binary to be flashed.
+const uint8_t app_binary[] = {0x2, 0x0, 0x34, 0x0, 0x0, 0x8, 0x0, 0x0, 0x1, 0x0, 0x0, 0x0, 0xD7, 0x75, 0x50, 0x6E,
```
alistair23, line 25: This application can build for any architecture. How does this data then work?

examples/ota_app_wip/main.c
viswajith-g: Gotcha, this was a legacy implementation from the person who worked on this project before me. I reworked the project, but forgot to remove those directories from my repo before committing and opening the PR is what I meant. 
Sorry for the confusion!

libtock/internal/app_loader.c
viswajith-g: Yeah, I duplicated this from `button.h`, but now I removed it and just included `button.h` in the `main.c` file.

examples/ota_app_wip/main.c
alistair23: Ah ok. That makes sense. You should squash the commits then. We don't need the legacy implementation committed and then removed when this PR is merged

examples/tests/app_loader/main.c
```diff
@@ -0,0 +1,298 @@
+// \file
```
alistair23, line 1: I think this app is complex enough that it deserves a README describing what it is doing

examples/tests/app_loader/main.c
```diff
@@ -0,0 +1,298 @@
...
+  // btn_num: The index of the button associated with the callback
+  // val: 1 if pressed, 0 if depressed
+  if ((btn_num == 0) && (val == 1)) {
+    printf("[Event] Button Pressed!\n");
+
+    int ret0;
+    ret0 = app_loader_exists();
+    if (ret0 == RETURNCODE_SUCCESS) {
+      printf("[Success] Dynamic Apploader driver exists.\n");
+    }
```
alistair23, line 188: We could do this as part of `main()` so that it is only called once at startup

examples/tests/app_loader/main.c
```diff
@@ -0,0 +1,298 @@
...
+
+    int ret  = 0;
+    int ret1 = 0;
+    int ret2 = 0;
```
alistair23, line 192: You can just reuse the same `ret`, no need to declare multiple variables

examples/tests/app_loader/main.c
```diff
@@ -0,0 +1,298 @@
...
+
+    app_size = sizeof(app_binary);
+    ret      = app_loader_command_setup(app_size); // asks the capsule to set up for app flash
+    if (ret != RETURNCODE_SUCCESS) {
+      printf("[Error] Setup Failed: %d.\n", ret);
+    }
+
+    if (ret == RETURNCODE_SUCCESS) {
```
alistair23, line 200: If `app_loader_command_setup()` fails then we won't run this, but we will run the next if as `RETURNCODE_SUCCESS` is `0`.

You should check for an error and then log and return if there is one

examples/tests/app_loader/main.c
```diff
@@ -0,0 +1,298 @@
...
+      } else {
+        printf("[Success] Process created successfully.\n");
+      }
+    }
+    app_size = 0; // reset app_size
+  }
+
+  if ((btn_num == 1) && (val == 1)) {
+    printf("Button 2 pressed!\n");
+  }
```
alistair23, line 222: Why not just support any button? It's not always obvious which button is button 1 on a board

examples/tests/app_loader/main.c
```diff
@@ -0,0 +1,298 @@
...
+  int ret;
+  uint32_t write_count = 0;
+  uint8_t write_buffer[FLASH_BUFFER_SIZE];
+  uint32_t flash_offset = 0;
+
+  write_count = sizeof(app_binary) / FLASH_BUFFER_SIZE;
+
+  ret = app_loader_write_buffer(write_buffer, FLASH_BUFFER_SIZE);   // set the write buffer
+  if (ret != RETURNCODE_SUCCESS) {
+    printf("[Error] Failed to set the write buffer.\n");
```
alistair23, line 243: You should print the failure reason

alistair23: or just return the error code directly as it's printed by the caller

examples/tests/app_loader/main.c
```diff
@@ -0,0 +1,298 @@
...
+    return -1;
+  }
+
+  for (uint32_t offset = 0; offset < write_count; offset++) {
+    memcpy(write_buffer, &app_binary[FLASH_BUFFER_SIZE * offset], FLASH_BUFFER_SIZE);     // copy binary to write buffer
```
alistair23, line 248: What if `app_binary` isn't a multiple of `FLASH_BUFFER_SIZE`?

examples/tests/app_loader/main.c
```diff
@@ -0,0 +1,298 @@
...
+    return err1;
+  }
+
+  int err2 = button_subscribe(button_callback, NULL);
+  if (err2 != 0) {
+    printf("[Error] Failed to set button callback.\n");
+    return err2;
+  }
+
+  while (1) {
```
alistair23, line 295: Maybe report to the user that we are waiting for a button press

libtock/internal/app_loader.c
alistair23: These are still here though

libtock/internal/app_loader.c
```diff
@@ -0,0 +1,60 @@
+#include <stdio.h>
+#include <string.h>
```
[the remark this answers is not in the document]
viswajith-g: Yes. 
We need the `stdio.h` or the compiler complains about implicit declaration of `printf`s.
We need `string.h` to use `memcpy`.

examples/tests/app_loader/main.c
viswajith-g: In this case, I used button 1 to load the app, and button 2 to print the statement to ensure I could get the buttons to work independently. I am removing it, but it can be added back anytime.

libtock/internal/app_loader.c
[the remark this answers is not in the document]
viswajith-g: Oh! my bad, I misread it as main.c. Yeah, I removed those, thanks!

libtock/internal/app_loader.c
viswajith-g: I have not pushed it yet. I am waiting to accumulate changes before pushing.

bradjc:
Can you rebase this on master and

- move the syscall wrappers to their own file
- move app_binaries and the python script to the test directory

viswajith-g:
> Can you rebase this on master and
I believe the repo was already rebased on master last time i pushed it.

> * move the syscall wrappers to their own file
Done. I also renamed the functions to start with libtock_ to keep in line with the other applications' format.
> * move app_binaries and the python script to the test directory
Done

examples/tests/app_loader/button-press-loading/app_binaries.c
```diff
@@ -0,0 +1,1304 @@
+#include <examples/tests/app_loader/button-press-loading/app_binaries.h>
```
bradjc, line 1:
```suggestion
#include "app_binaries.h"
```

libtock/internal/app_loader/app_loader.c
bradjc, line 1: Move these to `libtock/kernel`

libtock/internal/app_loader/syscalls/app_loader_syscalls.c
```diff
@@ -0,0 +1,46 @@
+#include "app_loader_syscalls.h"
+// #include "libtock/tock.h"
+
+// bool libtock_app_loader_exists(void) {
+//   syscall_return_t res = command(DRIVER_NUM_APP_LOADER, 0, 0, 0);
+//   return tock_command_return_novalue_to_returncode(res);
+// }
```
bradjc, line 7:
```suggestion
```

lvgl/lvgl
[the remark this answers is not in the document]
viswajith-g: i did not touch this though?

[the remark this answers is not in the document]
viswajith-g: nvm i see it now

viswajith-g:
ok, given the kernel part is now merged, i will leave this comment here: 

I am all for removing `button-press-loading` and renaming `abort-test` to `dynamic-process-loading-test` because abort test is an extension of button press. We add the abort function and test it is all. I don't think it is worth two separate examples and a whole lot of code repeat for that reason. Thoughts @bradjc?

libtock/kernel/app_loader.h
```diff
@@ -0,0 +1,98 @@
...
+#define BUTTON1 0
+#define BUTTON2 1
+#define BUTTON3 2
+
+/*
+ * Function to setup the callback from capsule.
+ * This function takes in the function that will be executed
+ * when the callback is triggered.
+ */
+returncode_t libtock_app_loader_set_setup_upcall(subscribe_upcall cb, void* userdata);
```
bradjc, line 20: These need to be updated to match https://github.com/tock/libtock-c/blob/3ff91f69e51ee305ab37e0878c88ba8a3a2b4ac2/doc/guide.md

examples/tests/app_loader/tock-dpl-hello/main.c
bradjc, line 1: What is this app for? It seems the same as c_hello.

bradjc (commented):
I think we should have one interactive app meant for trying out DPL, eg button-press-loading. The others should be more friendly for CI (ie not require user input). I think the abort test should be able to run and test abort without requiring a user to get the button timing right.

I also think we should come up with a folder and more general structure for storing the .tbf binaries. We should also make it clear the binaries are cortex-m4 binaries. They should match what tockloader generates.

examples/tests/app_loader/tock-dpl-hello/main.c
viswajith-g: This is the app that's compiled and fed to `tock-
-welcomes-dpl`.

viswajith-g:
Okay, I agree with that

viswajith-g:
> I also think we should come up with a folder and more general structure for storing the .tbf binaries. We should also make it clear the binaries are cortex-m4 binaries. They should match what tockloader generates.

Okay, so I gave it some thought and the way I have it working right now is to move the binaries to a different directory and create a symlink so the `Makefile` can find and compile the binaries for each example. I still have them stored in a single `app_binaries.c` file but I am considering splitting each binary into its own file. This should hopefully reduce the size of the compiled test app, with people getting to choose and test their own binaries easily downstream.

The structure looks like this right now:

```
│ 
├─ app-binaries/
│   ├─ cortex-m4/
│   	   ├─ app_binaries.h
│	   ├─ app_binaries.c
│
├─ abort-test/
│   ├─ app_binaries.c -> symlink
│   ├─ main.c
│   └─ Makefile
│
├─ button-press-loading/
│   ├─ app_binaries.c -> symlink
│   ├─ main.c
│   └─ Makefile
```

viswajith-g:
These pushes only address the directory structure changes and not the ci-ease concern that was raised. That will come with a separate commit

viswajith-g:
The current directory structure looks like this: 

```
│ 
├─ app-binaries/
│   ├─ cortex-m4/
│   	   ├─ tock-apps.h
│	   ├─ tock-dpl-hello.c
│         ├─ blink.c
│         ├─ adc.c
│
├─ abort-test/
│   ├─ tock-dpl-hello.c  -> symlink
│   ├─ blink.c                 -> symlink
│   ├─ adc.c                   -> symlink
│   ├─ main.c
│   ├─ Readme
│   └─ Makefile
│
├─ button-press-loading/
│   ├─ blink.c  -> symlink
│   ├─ adc.c    -> symlink
│   ├─ main.c
│   ├─ Readme
│   └─ Makefile
```

The binaries match the ones generated by tockloader. I pulled from the examples in the repo for `blink` and `adc`. I wrote my own example for the third, but it is a simple variation of `c-hello`.

For the abort-test, the user can press the button whenever, and it'll still abort. I've tested it multiple times, the NonvolatileStorage driver is just fast. 

However, I've changed the interface so that the app is listening for a console command instead of a button press. Essentially, the characters '0', '1', or '2' load a different app based on the command. This should eliminate those pesky debounce issues and whatnot. Plus it is easy to expand this anyway regardless of the board button support capabilities, while introducing the undefined behavior when multiple apps are listening however. 
'2' installs `adc`, but the app is programmed so that at the 50% mark, the app sends an abort signal internally. One could write a script to automate this.

bradjc:
Why does the abort test need to be interactive at all? It is simpler and easier to run if all you have to do is install it and see if it prints success or not.

You could add a timer (potentially with some random delay) if you want to not run immediately at boot or randomize when the abort signal happens.

bradjc:
> Okay, so I gave it some thought and the way I have it working right now is to move the binaries to a different directory and create a symlink so the `Makefile` can find and compile the binaries for each example. I still have them stored in a single `app_binaries.c` file but I am considering splitting each binary into its own file. This should hopefully reduce the size of the compiled test app, with people getting to choose and test their own binaries easily downstream. The downside to this would be multiple symlinks.

Why can't we just include in the makefile which binary apps to include?

viswajith-g:
I don't remember exactly why but the compiler kept complaining when I included the binaries in the make file. Something about a path issue. So I ended up making symlinks. I can try again.

examples/tests/app_loader/tool/README.md
lschuermann, line 1: Can you maybe rename this directory to something other than `tool`? That doesn't really suggest much of what it does.

examples/tests/app_loader/app_binaries/cortex-m4/adc.c
lschuermann, line 1: Do we need to check in those binaries? This seems like something that'll go out of sync with the rest of the repository quickly. It would be nice if, instead, these could be automatically generated upon compilation of the DPL tests.

Also, using symlinks to reference these files makes me skittish; I don't think that typically interacts well with C include paths, no? It might be better to modify the Makefile to add these files into the include path (or, if you're generating them on the fly as part of compiling the DPL binaries, you could simply place them in the `build/` output path).

lschuermann (commented):
Very cool.

@bradjc @viswajith-g Do you think it would make sense to also have some higher-level, reusable interface for process loading, such as the `write_app` function of the `button-press-loading` example available to all apps?

examples/tests/app_loader/app_binaries/cortex-m4/adc.c
viswajith-g: They aren't generated on the fly. We have a complementary tockloader command that helps convert a `.tab` file into a struct whose contents can be copied over. I do think symlinks may cause issues with pathings across OSes. 
These binaries are essentially c arrays that are passed as references to whatever app is loading them, they don't come from the original binaries in the example/tests directories. 

I'm having some troubles with getting the makefile to recognize the path to these files because of the way it builds. I haven't fully delved into it yet.

viswajith-g:
> Do you think it would make sense to also have some higher-level, reusable interface for process loading, such as the `write_app` function of the `button-press-loading` example available to all apps?

Hmm. I think there is merit in introducing an interface for the whole subsystem (setup + flash + load). That might reduce the implementation repeat. Another option, and something I am heavily leaning towards is to use a standalone `app_loader` application that functions as a background service (by making it a sticky app) which you can invoke via IPC.

examples/tests/app_loader/app_binaries/cortex-m4/adc.c
lschuermann: Unfortunately, I'm no C expert so probably not the best person to ask about this. But I do think that adjusting the include paths is better than using symlinks.

> They aren't generated on the fly. We have a complementary tockloader command that helps convert a .tab file into a struct whose contents can be copied over. I do think symlinks may cause issues with pathings across OSes.

I thought this was the job of the "tool" included in this PR?

My point is that I don't see a reason why they _shouldn't_ be generated on the fly, as part of building the DPL example applications themselves. You should be able to simply run another `make` command in the respective application directories as part of the Makefiles of the DPL example applications themselves.

Pinging @bradjc and @ppannuto who know more about the Makefile magic here and how this could be supported most elegantly.

lschuermann:
> Hmm. I think there is merit in introducing an interface for the whole subsystem (setup + flash + load). That might reduce the implementation repeat. Another option, and something I am heavily leaning towards is to use a standalone app_loader application that functions as a background service (by making it a sticky app) which you can invoke via IPC.

I don't think these are mutually exclusive: you could have a high-level interface, and an IPC service that can perform app loading for other applications. However, I think that given the current state of IPC (both concerning problems with the interface, and the app loader application not being able to reliably determine whether the IPC client is a trustworth app), having this be a reusable function that you can embed in your own app seems more attractive. This app can then be verified by the kernel as a whole, before being given permission to manage applications on the board.

examples/tests/app_loader/app_binaries/cortex-m4/adc.c
bradjc: I think there is something elegant about having the binaries be generated on the fly, and it would be cool to have that. But, given we want tutorial participants to be able to use this, and to add their own binaries for flashing (with things like different signatures), I think we need this to be simple like what is here now.

ppannuto: Okay, so I was starting down the road of making this more ergonomic / automatic with make stuff [which I can still do] — oorrrrrr, [we can bump the `Configuration.mk` to C23 and use the `#embed` directive](https://stackoverflow.com/a/79366789/358675).

For N=1 apps, switching from `-gnu11` to `-gnu23` worked no-problem (and out-of-the-box on the compiler that happened to be on this machine (`arm-none-eabi-gcc (GCC) 14.2.0`, not the most new by any means). I'm testing this a bit more thoroughly now and will make a separate PR to libtock-c in a moment if everything is smooth.

bradjc:
@ppannuto what is the status for a better method of including apps?

ppannuto:
Fell off my mental list, will work on it today, sorry

ppannuto:
@viswajith-g I forced-pushed an update to your branch here, so that I could rebase on latest `master` and pick up the `C23` changes.

This is otherwise currently a minimal update relative to what you had (i.e., it doesn't delete any of the hex-print code from the Python tool).

What it does do is remove the `app_binaries/` folder — instead, it will build on-demand a `build/cortex-m4/blink.embed` image to be embedded for each app.

Right now, the Makefile explicitly restricts `TOCK_TARGETS := cortex-m4` for each of the example apps. This reflects some hard-coded assumptions from the existing support tool I did not update (i.e., it never worked for anything but cortex-m4, this just makes that explicit).

Finally, for the moment, it assumes that `blink` and `adc` have been built already in your `libtock-c` checkout. That felt like a sane stopping point for a second until we decide how we want to "ship" pre-built images (if at all).

I think there's a lot of machinery here that could continue to be removed, but I wanted to get minimal use of `#embed` up for you in case it was blocking.

I'll try to keep improving this a little today, but won't touch things after 5pm pacific.

ppannuto:
Okay, actually, thinking about it for just a second more; now that we're just extracting the `.tbf` from the `.tab`, we can just use `tar` directly; which this now does.

So, the real question ergonomics-wise is "how do we want to expect people to get a baseline copy of `blink` or `adc`?"

I'm going to punt on that for _just_ a second to go ahead and fix the multi-arch, because that's trivial now...

examples/tests/app_loader/abort-test/main.c
```diff
@@ -0,0 +1,273 @@
+// \file
+
```
ppannuto, line 2:
```suggestion
```
?

examples/tests/app_loader/abort-test/README.md
```diff
@@ -0,0 +1,15 @@
+# Dynamic App Loader Abort Test (Helper App)
+============================================
+
+This app waits upon console input. It only recognizes three characters - 
+'0', '1' and '2'
```
ppannuto, line 5: What’s the long-term plan for `abort-test`? Currently the readme and the top comment in the file talk about integrating with the console, three app options, one that aborts; but the code just has the ADC app (and doesn’t seem to abort)?

My view is that the `button-press-loader` covers the “pick an app” case, and `abort-test` should reduce to just loading one specific app, and actually aborting the load — that seems easiest, and in line with the nature of a 'test' app to focus down on specific functionality

examples/tests/app_loader/button-press-loading/main.c
```diff
@@ -0,0 +1,281 @@
+// \file
+
```
ppannuto, line 2:
```suggestion
```

examples/tests/app_loader/button-press-loading/main.c
```diff
@@ -0,0 +1,281 @@
...
+/******************************************************************************************************
+* Callback functions
+*
+* 1. Callback to let us know when the capsule is done writing data to flash
+* 2. Set button callback to initiate the dynamic app load process on pressing button 1 (on nrf52840dk)
+*
+******************************************************************************************************/
+
+// static void nop_callback(int a __attribute__((unused)), int b __attribute__((unused)), int c __attribute__((unused)), void *d __attribute__((unused))) {}
```
ppannuto, line 58:
```suggestion
```

examples/tests/app_loader/abort-test/README.md
viswajith-g: this is exactly what is currently happening. i should change the readme. the code does what you describe

examples/tests/app_loader/button-press-loading/main.c
```diff
@@ -0,0 +1,281 @@
...
+  if (pressed == 1 && !button_press) {
+    // Note: this variable is introduced
+    // because the current alarm upcall implementation
+    // results in panic when a button is pressed within
+    // the debounce period if the debounce period is long.
+    // Setting the debounce interval to 100ms seems to work
+    // but setting it to 200ms and rapidly clicking buttons
+    // leads to the kernel panicking.
+    button_press = true;
+    libtocksync_alarm_delay_ms(200); // debounce
```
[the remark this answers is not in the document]
viswajith-g: i agree, i wrote this so long ago and it's gone through so many changes that i cannot even reason why this structure exists anymore. I'll move the logic to main

examples/tests/app_loader/app_binaries/cortex-m4/adc.c
ppannuto: This now will use apps built elsewhere in a libtock-c checkout, and will build them on-demand if-needed.

examples/tests/app_loader/button-press-loading/main.c
```diff
@@ -0,0 +1,287 @@
...
+          app_name = "adc";
+          app_data = APP_ADC;
+          app_size = sizeof(APP_ADC);
+          break;
+        default:
+          printf("[Log] Unsupported Button.\n");
+          app_load = false;
+          return;
+      }
+    }
```
ppannuto, line 124:
```suggestion
    switch (btn_num) {
      case BUTTON1:
        app_name = "blink";
        app_data = APP_BLINK;
        app_size = sizeof(APP_BLINK);
        app_load = true;
        break;
      case BUTTON2:
        app_name = "adc";
        app_data = APP_ADC;
        app_size = sizeof(APP_ADC);
        app_load = true;
        break;
      default:
        printf("[Log] Unsupported Button.\n");
        return;
    }
```

I think this can still be way simpler and remove the weird debounce thing if you just treat `app_load` as a "already busy, ignore button" flag.

viswajith-g: yeah i forgot to push that change lol

viswajith-g: ok i think i remember why i did not push it. it was because in the unsupported case, the debounce still exists and prints logs the unsupported button case multiple times. i found it to be quite ugly

ppannuto: I might push for this change still. As-written here, for the "pushed a valid button case", the bounces are dropped silently. An extra print or two for the invalid button case isn't really something that users will see often, and I think having a simpler app is more valuable.

bradjc:
There are few updates we need to make to `libtock_app_loader`. 

- Remove the duplicated function comments from app_loader.c.
- The functions in libtock/kernel/app_loader should be semantically meaningful, not just duplicates of the syscall interface. For example, there should be a write function which takes a buffer, offset, length, and callback and does the entire write operation, including the allow and subscribe.
- Remove the button `#defines`

examples/tests/app_loader/support/AppLoaderSupport.mk
bradjc, line 1: We already have this file.

bradjc: That's fine, I don't think we necessarily have a great way of handling this already.

viswajith-g:
Ok currently what I have works and is an improvement over what we had before. However, the write() function still requires for the entire app binary to be passed to it, making it unfriendly to streams of data coming in. I will be working on that change next.

viswajith-g:
updated the tutorial code, passes build now

libtock/kernel/app_loader.c
```diff
@@ -1,96 +1,156 @@
...
+    memset(write_buffer + chunk_len, 0, FLASH_BUFFER_SIZE - chunk_len);
+  }
+
+  ret = libtock_app_loader_command_write(offset, FLASH_BUFFER_SIZE);
+  if (ret != RETURNCODE_SUCCESS) {
+    printf("[Error] Flash write failed at offset 0x%lx: err %d\n", offset, ret);
+    return ret;
+  }
+
+  yield_for(&write_done);
```
bradjc, line 83: Can't call yield in `libtock` (async only)`.

libtock/kernel/app_loader.h
```diff
@@ -8,89 +8,17 @@ extern "C"
 #include "libtock/kernel/syscalls/app_loader_syscalls.h"
 #include "libtock/tock.h"
 
-#define BUTTON1 0
-#define BUTTON2 1
-#define BUTTON3 2
+returncode_t libtock_app_loader_setup(uint32_t app_length, subscribe_upcall cb);
```
bradjc, line 11: The callback needs to be customized to this driver.

libtock/kernel/syscalls/app_loader_syscalls.h
```diff
@@ -69,23 +48,26 @@ returncode_t libtock_app_loader_command_finalize(void);
...
+ * Functions to setup the callback from capsule.
  * This function takes in the function that will be executed
  * when the callback is triggered.
  */
+returncode_t libtock_app_loader_set_setup_upcall(subscribe_upcall cb, void* userdata);
+returncode_t libtock_app_loader_set_write_upcall(subscribe_upcall cb, void* userdata);
+returncode_t libtock_app_loader_set_finalize_upcall(subscribe_upcall cb, void* userdata);
+returncode_t libtock_app_loader_set_load_upcall(subscribe_upcall cb, void* userdata);
 returncode_t libtock_app_loader_set_abort_upcall(subscribe_upcall cb, void* userdata);
+returncode_t libtock_app_loader_set_uninstall_upcall(subscribe_upcall cb, void* userdata);
```
bradjc, line 70: The function docs should be in the .h file and not in the .c file.

viswajith-g: huh? isn't this the .h file?

PR #389, opened 2024-04-02
Fixed documentation on behaviour on return from main

Ioan-Cristian:
### Pull request overview

The documentation regarding the behaviour on return from main was outdated. This pull request fixes it.

### TODO

Nothing

### Documentation

+ [x] Updated relevant files.

### Format

+ [x] Run `./format_all.sh`

### DIFF
--- a/libtock/crt0.c
+++ b/libtock/crt0.c
@@ -160,7 +160,6 @@ void _start(void* app_start __attribute__((unused)),
     "movs r0, r6\n"             // first arg is app_start
     "movs r1, r7\n"             // second arg is mem_start
     "bl _c_start_pic\n"
-    "bkpt #255\n"
     );
 
 #elif defined(__riscv)
--- a/doc/overview.md
+++ b/doc/overview.md
@@ -12,21 +12,20 @@ method with the following signature:
 int main(void);
 ```
 
-Applications **should** return 0 from `main`. Returning non-zero is undefined and the
-behavior may change in future versions of `libtock`.
-Today, `main` is called from `_start` and includes an implicit `while()` loop:
+The return value of the main function is used as an argument of the `exit` system call:
+
+If an application is long running, then it should set up a series of event subscriptions
+and add an explicit `while` loop at the end of the `main` function:
 
 ```c
-void _start(void* text_start, void* mem_start, void* memory_len, void* app_heap_break) {
-  main();
-  while (1) {
-    yield();
-  }
+int main(void) {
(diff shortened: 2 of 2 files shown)


libtock/crt0.c
```diff
@@ -160,7 +160,6 @@ void _start(void* app_start __attribute__((unused)),
     "movs r0, r6\n"             // first arg is app_start
     "movs r1, r7\n"             // second arg is mem_start
     "bl _c_start_pic\n"
-    "bkpt #255\n"
```
Ioan-Cristian, line 163: `_c_start_pic` is marked as `noreturn`, so this instruction is useless and  never reached. This change is consistent with RISC-V variant.

alistair23 (approved):

libtock/crt0.c
ppannuto: _Technically_, a compiler _could_ ignore the `noreturn` attribute. This is a 2-byte hedge for a corner case of something going particularly wonky (see the comment above explaining its purpose). I'd rather _add_ this to RISC-V than remove it here.

ppannuto (changes requested):
Looks like this got missed in #246 — thanks!

libtock/crt0.c
Ioan-Cristian: This looks mostly like a defensive approach. `_c_start_pic` and `_c_start_nopic` are the functions marked as `noreturn`. Most end users will never have to touch these functions. Furthermore, modern compilers are good enough to determine if a `noreturn` function has an execution path that leads to exiting the function. The only way one could return from a `noreturn` function is through inline assembly, but at that point everything is possible, including modifying the return address to point past the break point making `bkpt` and `ebreak` useless.

ppannuto (approved):
I'd still rather be defensive, as I've been the recipient of the person who overrode startup symbols (but didn't say that) but then sent a bug report of things not working..., however, it's enough of a rare case that I'd rather just have the doc fix here and move on from my prior traumas 😓.

Ioan-Cristian:
Tock doesn't use either [bkpt](https://github.com/tock/tock/blob/1d9c6412f95bd8a098b3a378a7b2f91779a3ed9b/arch/cortex-m/src/lib.rs#L191) or [ebreak](https://github.com/tock/tock/blob/1d9c6412f95bd8a098b3a378a7b2f91779a3ed9b/arch/rv32i/src/lib.rs#L138) even though in theory the main function could return through inline assembly.

PR #390, opened 2024-04-02
make: TockLibrary: unconditionally set build dir

bradjc:
This fixes an issue when multiple Tock libraries are used in the same build. With `?=`, the value is "reset" for each instantiation of TockLibrary, causing the build directory for all libraries to be set to whichever library is included last. With `:=` this does not occur.


ppannuto (approved):
Oh make... nice catch

PR #394, opened 2024-04-05
Add support for transmitting and receiving GPS location over LoRa

alistair23:
This uses https://github.com/tock/tock/pull/3881 to read a NMEA location and transmit it over LoRa


alistair23:
Ping!

PR #395, opened 2024-04-05
timer: handle timer overflow

Samir-Rashid:
PR fixes https://github.com/tock/tock/issues/3879
Tock does not support setting alarms more than 2^32 ticks in the future as the `alarm_at` syscall takes in ticks. For the default nrf52840 tock configuration at 32KHz, the greatest possible value for the syscall alarm is 37 hours.

`timer_in` takes in milliseconds, but is flawed in that it is just a wrapper around `alarm_at` which can only accept a max value of 2^32 ticks. Thus, calling `timer_in` longer than 2^32 ticks in milliseconds will not have the expected behavior. `delay_ms` calls `timer_in`, so it is flawed for the same reason.

We fixed this by modifying `timer_in`. If the timer length is longer than the time that `alarm_at` can count up to, then `timer_in` will schedule multiple alarms to reach the full length. We calculate the number of full overflows and the remainder ticks to reach the target length of time. The overflows use the `overflow_callback` for each intermediate overflow.

## Testing

We verified this code works by creating a test program `timer_overflow` that simulates setting a timer longer than the threshold of what would overflow and verifies the total timer length was correct. To test overflow you need to edit `#define MAX_TICKS` in `alarm_timer.c` to overflow in a reasonable amount of time. This tests the overflow logic, which works correctly.

### DIFF
--- a/libtock/services/alarm.h
+++ b/libtock/services/alarm.h
@@ -50,9 +50,16 @@ typedef struct alarm {
  * An opaque handle to a repeating alarm created by `libtock_alarm_repeating_every`.
  */
 typedef struct alarm_repeating {
-  uint32_t interval;
-  libtock_alarm_callback cb;
-  void* ud;
+  // Length of timer in milliseconds.
+  uint32_t interval_ms;
+  // Number of times the underlying counter will overflow
+  // before hitting the target interval_ms.
+  uint32_t overflows_left;
+  // Number of ticks remaining after the last time the
+  // counter overflows.
+  uint32_t remaining_ticks;
+  libtock_alarm_callback callback;
+  void* user_data;
   libtock_alarm_t alarm;
 } libtock_alarm_repeating_t;
 
@@ -60,17 +67,21 @@ typedef struct alarm_repeating {
 /** \brief Create a new alarm to fire at a particular clock value.
  *
  * The `alarm` parameter is allocated by the caller and must live as long as
- * the alarm is outstanding.
+ * the alarm is outstanding. `reference` and `dt` are in terms of the tick
+ * time.
  *
- * \param reference: the reference time from which the alarm is being set
- * \param dt: the time after reference that the alarm should fire
+ * Alarms longer than 2^32 ticks should use `timer_in`.
+ *
+ * \param reference the reference time from which the alarm is being set in ticks.
+ * \param dt the time after reference that the alarm should fire in ticks.
  * \param callback a callback to be invoked when the alarm expires.
  * \param userdata passed to the callback.
- * \param a pointer to a new alarm_t to be used by the implementation to keep
+ * \param alarm pointer to a new alarm_t to be used by the implementation to keep
  *        track of the alarm.
  * \return An error code. Either RETURNCODE_SUCCESS or RETURNCODE_FAIL.
  */
-int libtock_alarm_at(uint32_t reference, uint32_t dt, libtock_alarm_callback cb, void* opaque, libtock_alarm_t* alarm);
+int libtock_alarm_at(uint32_t reference, uint32_t dt, libtock_alarm_callback callback, void* opaque,
+                     libtock_alarm_t* alarm);
 
 /** \brief Cancels an existing alarm.
  *
@@ -95,13 +106,19 @@ int libtock_alarm_gettimeasticks(struct timeval* tv, void* tzvp);
 
 /** \brief Create a new alarm to fire in `ms` milliseconds.
  *
+ * The `timer` parameter is allocated by the caller and must live as long as
+ * the alarm is outstanding.
+ *
+ * This should be used for long periods. For more precision, but less time
+ * range, use `alarm_at`.
+ *
  * \param ms the number of milliseconds to fire the alarm after.
- * \param callback a callback to be invoked when the alarm expires.
+ * \param cb a callback to be invoked when the alarm expires.
  * \param opaque pointer passed to the callback.
- * \param A handle to the alarm that was created.
+ * \param alarm handle to the alarm that was created.
  * \return An error code. Either RETURNCODE_SUCCESS or RETURNCODE_FAIL.
  */
-int libtock_alarm_in_ms(uint32_t ms, libtock_alarm_callback cb, void* opaque, libtock_alarm_t* alarm);
+int libtock_alarm_in_ms(uint32_t ms, libtock_alarm_callback cb, void* opaque, libtock_alarm_repeating_t* alarm);
 
 /** \brief Create a new repeating alarm to fire every `ms` milliseconds.
  *
@@ -109,13 +126,13 @@ int libtock_alarm_in_ms(uint32_t ms, libtock_alarm_callback cb, void* opaque, li
  * the repeating alarm is outstanding.
  *
  * \param ms the interval to fire the alarm at in milliseconds.
- * \param callback a callback to be invoked when the alarm expires.
+ * \param cb a callback to be invoked when the alarm expires.
  * \param opaque pointer passed to the callback.
- * \param a pointer to a new alarm_repeating_t to be used by the implementation to
+ * \param alarm_repeating pointer to a new alarm_repeating_t to be used by the implementation to
  *        keep track of the alarm.
  */
-void libtock_alarm_repeating_every(uint32_t ms, libtock_alarm_callback cb, void* opaque,
-                                   libtock_alarm_repeating_t* alarm_repeating);
+void libtock_alarm_repeating_every_ms(uint32_t ms, libtock_alarm_callback cb, void* opaque,
+                                      libtock_alarm_repeating_t* alarm_repeating);
 
 /** \brief Cancels an existing repeating alarm.
  *
--- a/libtock/services/alarm.h
+++ b/libtock/services/alarm.h
@@ -2,7 +2,7 @@
  * This module allows the client to initiate alarms and receive
  * callbacks when those alarms have expired. Clients can set one-shot alarms to
  * fire at particular clock values (`libtock_alarm_at`) or periodic alarms
- * (`libtock_alarm_repeating_every`)
+ * (`libtock_alarm_repeating_every_ms`)
  *
  * The client should not assume anything about the underlying clock used by an
  * implementation other than that it is running at sufficient frequency to
@@ -43,13 +43,14 @@ typedef struct alarm {
   void* ud;
   struct alarm* next;
   struct alarm* prev;
-} libtock_alarm_t;
+} libtock_alarm_ticks_t;
 
 /** \brief Opaque handle to a repeating alarm.
  *
- * An opaque handle to a repeating alarm created by `libtock_alarm_repeating_every`.
+ * An opaque handle to an alarm created by `libtock_alarm_repeating_every_ms`
+ * and `libtock_alarm_in_ms`.
  */
-typedef struct alarm_repeating {
+typedef struct alarm_data {
   // Length of timer in milliseconds.
   uint32_t interval_ms;
   // Number of times the underlying counter will overflow
@@ -60,8 +61,8 @@ typedef struct alarm_repeating {
   uint32_t remaining_ticks;
   libtock_alarm_callback callback;
   void* user_data;
-  libtock_alarm_t alarm;
-} libtock_alarm_repeating_t;
+  libtock_alarm_ticks_t alarm;
+} libtock_alarm_t;
 
 
 /** \brief Create a new alarm to fire at a particular clock value.
@@ -70,7 +71,7 @@ typedef struct alarm_repeating {
  * the alarm is outstanding. `reference` and `dt` are in terms of the tick
  * time.
  *
- * Alarms longer than 2^32 ticks should use `timer_in`.
+ * Alarms longer than 2^32 ticks should use `libtock_alarm_in_ms`.
  *
  * \param reference the reference time from which the alarm is being set in ticks.
  * \param dt the time after reference that the alarm should fire in ticks.
@@ -81,7 +82,7 @@ typedef struct alarm_repeating {
  * \return An error code. Either RETURNCODE_SUCCESS or RETURNCODE_FAIL.
  */
 int libtock_alarm_at(uint32_t reference, uint32_t dt, libtock_alarm_callback callback, void* opaque,
-                     libtock_alarm_t* alarm);
+                     libtock_alarm_ticks_t* alarm);
 
 /** \brief Cancels an existing alarm.
  *
@@ -89,7 +90,7 @@ int libtock_alarm_at(uint32_t reference, uint32_t dt, libtock_alarm_callback cal
  *
  * \param alarm
  */
-void libtock_alarm_cancel(libtock_alarm_t*);
+void libtock_alarm_cancel(libtock_alarm_ticks_t* alarm);
 
 // Use this to implement _gettimeofday yourself as libtock-c doesn't provide
 // an implementation.
@@ -118,27 +119,27 @@ int libtock_alarm_gettimeasticks(struct timeval* tv, void* tzvp);
  * \param alarm handle to the alarm that was created.
  * \return An error code. Either RETURNCODE_SUCCESS or RETURNCODE_FAIL.
  */
-int libtock_alarm_in_ms(uint32_t ms, libtock_alarm_callback cb, void* opaque, libtock_alarm_repeating_t* alarm);
+int libtock_alarm_in_ms(uint32_t ms, libtock_alarm_callback cb, void* opaque, libtock_alarm_t* alarm);
 
 /** \brief Create a new repeating alarm to fire every `ms` milliseconds.
  *
- * The `alarm_repeating` parameter is allocated by the caller and must live as long as
+ * The `alarm` parameter is allocated by the caller and must live as long as
  * the repeating alarm is outstanding.
  *
  * \param ms the interval to fire the alarm at in milliseconds.
  * \param cb a callback to be invoked when the alarm expires.
  * \param opaque pointer passed to the callback.
- * \param alarm_repeating pointer to a new alarm_repeating_t to be used by the implementation to
+ * \param alarm pointer to a new libtock_alarm_t to be used by the implementation to
  *        keep track of the alarm.
  */
 void libtock_alarm_repeating_every_ms(uint32_t ms, libtock_alarm_callback cb, void* opaque,
-                                      libtock_alarm_repeating_t* alarm_repeating);
+                                      libtock_alarm_t* alarm);
 
-/** \brief Cancels an existing repeating alarm.
+/** \brief Cancels an existing alarm set in milliseconds.
  *
- * \param alarm_repeating
+ * \param alarm to cancel.
  */
-void libtock_alarm_repeating_cancel(libtock_alarm_repeating_t* alarm_repeating);
+void libtock_alarm_ms_cancel(libtock_alarm_t* alarm);
 
 #ifdef __cplusplus
 }
--- a/examples/tests/alarm_in_overflow/Makefile
+++ b/examples/tests/alarm_in_overflow/Makefile
@@ -1,0 +1,11 @@
+# Makefile for user application
+
+# Specify this directory relative to the current application.
+TOCK_USERLAND_BASE_DIR = ../../..
+
+# Which files to compile.
+C_SRCS := $(wildcard *.c)
+
+# Include userland master makefile. Contains rules and flags for actually
+# building the application.
+include $(TOCK_USERLAND_BASE_DIR)/AppMakefile.mk
--- a/examples/tests/alarm_in_overflow/README.md
+++ b/examples/tests/alarm_in_overflow/README.md
@@ -1,0 +1,33 @@
+Test Alarm Overflow
+====================
+
+This tests the libtock alarm implementation's ability to handle situations
+where setting an alarm would overflow the underlying hardware counter.
+
+The userspace `libtock_alarm_in_ms` function expects a number of milliseconds when a timer
+should fire in the future. This argument is a `uint32_t` which means that an
+alarm should be able to be set $2^{32}$ ms in the future (about 49 days). However,
+`libtock_alarm_in_ms` calls to `libtock_alarm_at` which can only set alarms up to $2^{32}$ clock ticks
+(only about 36 hours with a 32768Hz oscillator). `libtock_alarm_in_ms` handles the underlying
+overflow of the alarm and makes the timer appear as if it can set longer timers.
+Under the hood, it keeps track of how many times the alarm needs to overflow until
+the target time is reached (how many $2^{32}$ tick intevals need to pass). Only after
+the last alarm does it call the original callback provided to `libtock_alarm_in_ms` while
+hiding the effects of the intermediate alarms that were scheduled.
+
+## Example Output
+
+To replicate the behavior of the following example output, modify the "MAX_TICKS"
+macro in "libtock/services/alarm.c" to whatever the frequency of your board is. 
+For example, if the frequency is 32768Hz and we set "MAX_TICKS" to 32768 we can simulate 
+the counter overflowing after 1 second.
+Once you have simulated your board overflowing after 1 second, you can set "TIMER_LEN_MS"
+in the test `main.c` to have the timer fire after 2 seconds. This corresponds to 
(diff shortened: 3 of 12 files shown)


libtock/alarm_timer.c
```diff
@@ -106,6 +122,79 @@ int alarm_at(uint32_t reference, uint32_t dt, subscribe_upcall cb, void* ud, ala
...
+      ud->original_ud,
+      ud->alarm
+      );
+  } else {
+    // schedule next intermediate alarm that will overflow
+    ud->overflows_left--;
+
+    alarm_at(
+      last_timer_fire_time,
+      UINT_MAX,
```
Samir-Rashid, line 164: Is it alright to use `UINT_MAX` for the max value of `uint32_t`? They have the same value for me, but maybe not for everyone?

tyler-potyondy: Could you use `UINT32_MAX`?

tyler-potyondy (changes requested):
In addition to testing the wrap around logic, it would likely prove useful to test the timing precision of this implementation (i.e. by calling `timer_in` in conjunction with `gettimeasticks` to see if the timer fired near when it was expected).

libtock/alarm_timer.c
Samir-Rashid: Yes, just fixed it. I tried that and it didn't compile earlier, not sure why.

ppannuto:
Following on Tyler's comment, it would be nice to capture the more complete testing program in the `tests/` folder, so that we have something to validate with down the road as well.

libtock/alarm_timer.c
```diff
@@ -106,6 +131,84 @@ int alarm_at(uint32_t reference, uint32_t dt, subscribe_upcall cb, void* ud, ala
...
+  uint32_t now;
+  int ret = alarm_internal_read(&now);
+  if (ret != RETURNCODE_SUCCESS) return ret;
+
+  // If `dt_ms` is longer than the time that an alarm can count up to, then `timer_in` will
+  // schedule multiple alarms to reach the full length. We calculate the number of full overflows
+  // and the remainder ticks to reach the target length of time. The overflows use the
+  // `overflow_callback` for each intermediate overflow.
+  uint32_t max_ticks_in_ms = ticks_to_ms(MAX_TICKS);
+  if (dt_ms > max_ticks_in_ms) {
```
alistair23, line 193: We should report this back to the user somehow. This could have a large impact on power usage for long sleeping devices as we wake up the CPU a lot more then we should. Maybe a `bool` to enable this feature?

examples/tests/timer_overflow/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+This tests the libtock timer implementation's ability to handle situations
+where setting a timer would overflow the underlying hardware counter.
+
+The userspace `timer_in` function expects a number of milliseconds when a timer
+should fire in the future. This argument is a `uint32_t` which means that an
+alarm should be able to be set 2^32 ms in the future (about 49 days). However,
+`timer_in` calls to `alarm_at` which can only set alarms up to 2^32 clock ticks
+(only about 36 hours with a 32768Hz oscillator). Therefore, `timer_in` has been
+modified to handle the underlying overflow of the alarm and make the timer appear
```
tyler-potyondy, line 12:
```suggestion
(only about 36 hours with a 32768Hz oscillator). `timer_in` handles the underlying overflow of the alarm and makes the timer appear
```

examples/tests/timer_overflow/main.c
```diff
@@ -0,0 +1,80 @@
...
+{
+  // need to allocate memory for a timer before giving a pointer to `timer_in`
+  tock_timer_t* timer = malloc(sizeof(tock_timer_t));
+
+  uint32_t timer_start_ticks;
+  alarm_internal_read(&timer_start_ticks);
+  testing_ud_t* ud = malloc(sizeof(testing_ud_t));
+  ud->timer_len_ms             = TIMER_LEN_MS;
+  ud->timer_start_ticks        = timer_start_ticks;
+  ud->expected_timer_end_ticks = timer_start_ticks + ms_to_ticks(TIMER_LEN_MS);
```
tyler-potyondy, line 63: Does the current design require this information be passed to `timer_in`? Ideally, `timer_in` would abstract needing to allocate this struct and just be passed a time in ms and a timer object.

libtock/timer.h
```diff
@@ -45,14 +45,21 @@ typedef struct tock_timer {
 
 
 /** \brief Create a new alarm to fire in `ms` milliseconds.
+ * 
+ * The `timer` parameter is allocated by the caller and must live as long as
+ * the alarm is outstanding.
+ * 
+ * This should be used for long periods. Due to no reference time, there could
+ * be up to a few milliseconds of inaccuracy. For more accuracy, but less time
+ * range, use `alarm_at`.
```
tyler-potyondy, line 54: Did the testing show a few milliseconds of inaccuracy? Our timer should have at least millisecond accuracy (we say elsewhere in libtock-c docs that this is an assumption). 

I assume the testing for the accuracy was using `printf` which introduces inherent latency. @ppannuto @bradjc have we benchmarked timers before for accuracy?

examples/tests/timer_overflow/main.c
Samir-Rashid: You can ignore the `ud` (user data) part of the timer by passing in a nullptr. `timer_in` does not use this at all - it is mechanism to pass data to your callback. As `ud` is effectively an optional parameter, I do not see any reason to add another function stub that omits `ud`.
Using `ud` in this test means that we also check that we do not corrupt this field.

Samir-Rashid: As for the usage of `ud` here, these are values that we wanted to send to the callback for our testing setup. Originally we were printing out these values in the callback but we moved the prints to the end of the file after the `yield_for` so these don't need to be passed as user data.

libtock/alarm_timer.c
atar13: For the common case of alarms less than 2^32 ticks (which is about 36 hrs for a 32KHz clock), no additional alarms will be scheduled. The behavior will match how `timer_in` worked before these changes. We've verified this with our testing program (found at `examples/test/timer_overflow/main.c`) and it can be reproduced by setting `TIMER_LEN_MS` to a length of time that will not overflow the counter.

For the uncommon case of alarms greater than 2^32 ticks there will be additional overhead. In our testing we have observed it to be around 1ms. This can be reproduced by setting `MAX_TICKS` in `libtock/alarm_timer.c` to something smaller than `TIMER_LEN_MS` in our testing file. For our Nordic NRF52840 board with an oscillator of 32768Hz, we set  `MAX_TICKS` to 32768Hz which will simulate overflowing every 1 second. In our testing file we set an alarm of 2 seconds (setting `TIMER_LEN_MS` to 2000) which means that we will schedule two 1 second long alarms to reach our target time of 2 seconds. As mentioned, the overhead of scheduling an additional intermediate alarm is only 1ms. This is tolerable since this will only happen when setting long running timers (over 36 hours with the 32KHz oscillator).

examples/tests/timer_overflow/main.c
atar13: To avoid any confusion we've removed this `testing_ud_t` struct and are passing in a single integer as user data. There's also a comment that explains how the user_data parameter is completely optional and that we are only using it for printing out values in the callback. The comment mentions that they can pass NULL as `ud` if they don't need to pass any user data.

libtock/timer.h
atar13: The testing program has also been updated to print timer error in milliseconds. See my comment to Allistair for a breakdown of our results and how to reproduce the testing: https://github.com/tock/libtock-c/pull/395#discussion_r1569271372

libtock/alarm_timer.c
atar13: Also, with regards to waking up the CPU more often. The CPU will not be waken up any more than the previous version of `timer_in` for the common case of alarms less than 2^32 ticks. 

The alternative to having this feature is that `timer_in` will behave incorrectly when the user tries to set a long running alarm that will overflow the counter. We think that by default `timer_in` should count the correct amount of time.

libtock/alarm_timer.c
```diff
@@ -4,11 +4,36 @@
...
 #include <stdlib.h>
 
+#define MAX_TICKS UINT32_MAX
+
 // Returns 0 if a <= b < c, 1 otherwise
 static int within_range(uint32_t a, uint32_t b, uint32_t c) {
   return (b - a) < (b - c);
 }
 
+inline static uint32_t ticks_to_ms(uint32_t ticks) {
```
ppannuto, line 14: Does the `inline` keyword actually do anything meaningful on modern compilers? With `static`, this will already be limited to this compilation unit and the compiler can make the inlining decision.

ppannuto (dismissed):
One last documentation ask, then I think this is good to go. Implementation looks sound.

libtock/alarm_timer.c
```diff
@@ -106,6 +131,84 @@ int alarm_at(uint32_t reference, uint32_t dt, subscribe_upcall cb, void* ud, ala
...
+      MAX_TICKS,
+      (subscribe_upcall*) overflow_callback,
+      (void*) ud,
+      ud->alarm
+      );
+  }
+}
+
+// Timer implementation
+int timer_in(uint32_t dt_ms, subscribe_upcall cb, void* ud, tock_timer_t *tock_timer) {
```
bradjc, line 182: Why did this function move in the file?

libtock/alarm_timer.c
```diff
@@ -106,6 +131,84 @@ int alarm_at(uint32_t reference, uint32_t dt, subscribe_upcall cb, void* ud, ala
...
+  int ret = alarm_internal_read(&now);
+  if (ret != RETURNCODE_SUCCESS) return ret;
+
+  // If `dt_ms` is longer than the time that an alarm can count up to, then `timer_in` will
+  // schedule multiple alarms to reach the full length. We calculate the number of full overflows
+  // and the remainder ticks to reach the target length of time. The overflows use the
+  // `overflow_callback` for each intermediate overflow.
+  uint32_t max_ticks_in_ms = ticks_to_ms(MAX_TICKS);
+  if (dt_ms > max_ticks_in_ms) {
+    overflow_ud_t* tmp_ud = malloc(sizeof(overflow_ud_t));
```
bradjc, line 194: I don't see a `free` in this patch.

In general, we want to move away from libtock holding internal state. Can this just be part of the timer object?

bradjc (changes requested):

libtock/alarm_timer.c
Samir-Rashid: Fixed.

libtock/alarm_timer.c
Samir-Rashid: > I don't see a free in this patch.

Sorry! Can't believe I missed the oldest trick in the book.

> Can this just be part of the timer object?

Good point. This could be added to the timer object. This will require changing how the timer length is stored from `uint32_t interval;` to `int overflows_left;` plus `int remainder_ticks;`.

atar13: Resolved in https://github.com/tock/libtock-c/pull/395/commits/f0b0b4691fad3a091da2768953feb9783234ddca

We no longer have any frees/mallocs. Just modified the fields of `tock_timer_t` to support the overflow handling.

atar13: We verified that our test with `timer_in` still works as intended and that `timer_every` still works by running `examples/tests/multi_alarm_test`

examples/tests/timer_overflow/main.c
bradjc, line 1: I don't understand what this test does. Setting a timer for 2 seconds will work just fine.

I think having to modify the libtock library is a high burden for running this test. I think it would be better to calculate how long until a overflow will happen, and set a timer for longer than that.

We then only run this test on a kernel with a very fast clock.

examples/tests/timer_overflow/main.c
```diff
@@ -0,0 +1,71 @@
...
+  fired = true;
+
+  uint32_t expected_timer_end_ms = *((uint32_t*)user_data);
+  uint32_t actual_fire_time_ms   = get_time_ms();
+  uint32_t difference_ms         = actual_fire_time_ms - expected_timer_end_ms;
+
+  printf("Timer Fired!\n");
+  printf("\tExpected timer to fire at %ld ms\n", expected_timer_end_ms);
+  printf("\tTimer actually fired at %ld ms\n", actual_fire_time_ms);
+  printf("\tDifference of %ld ms\n", difference_ms);
```
bradjc, line 48: I think this should somehow indicate test success.

libtock/alarm_timer.c
```diff
@@ -123,45 +243,89 @@ void alarm_cancel(alarm_t* alarm) {
...
+  }
 }
 
 // Timer implementation
-
-int timer_in(uint32_t ms, subscribe_upcall cb, void* ud, tock_timer_t *timer) {
-  uint32_t frequency;
-  alarm_internal_frequency(&frequency);
-  uint32_t interval = (ms / 1000) * frequency + (ms % 1000) * (frequency / 1000);
+int timer_in(uint32_t dt_ms, subscribe_upcall cb, void* ud, tock_timer_t *tock_timer) {
```
bradjc, line 282: Why did the argument name need to change? Now it doesn't match the header file.

libtock/alarm_timer.c
```diff
@@ -123,45 +243,89 @@ void alarm_cancel(alarm_t* alarm) {
...
+                              __attribute__ ((unused)) int unused1,
+                              __attribute__ ((unused)) int unused2,
+                              void*                        overflow_ud) {
+  tock_timer_t* tock_timer = (tock_timer_t*)overflow_ud;
+
+  if (tock_timer->overflows_left == 0) {
+    // no overflows left, schedule last alarm with original callback
+    alarm_at(
+      last_timer_fire_time,
+      tock_timer->remaining_ticks,
```
bradjc, line 262: If this alarm ends up set in the past (ie because remaining ticks happens to be its lowest possible value, and it took awhile for the kernel to get around to actually running this upcall), is everything ok?

libtock/alarm_timer.c
atar13: Resolved here: https://github.com/tock/libtock-c/pull/395/commits/6277a8924e0d4f3b0942deb1e032ce9314784136

`timer_in` arguments are consistent with the header.

examples/tests/timer_overflow/main.c
atar13: That sounds reasonable.

I have 2 questions:

1. How do we know the number of milliseconds that it will take to overflow? We know that overflows happen after $2^{32}$ ticks, however the number of milliseconds that takes depends on the frequency of the clock. We can query this value for the purpose of the test using `alarm_internal_frequency` function, but we were told not to use the `alarm_internal_*` functions in our testing code. We could hard code this value for the NRF52840dk board or just set a timer for an absurdly large number (like $2^{32}$ ms) likely overflow.
2. How can a user customize their kernel to run with a fast clock? Does the kernel already allow users to customize the clock speed?

libtock/alarm_timer.c
atar13: In this case, the upcall should fire immediately once the kernel gets around to scheduling it. We think this is the behavior based on [this comment and implementation in the alarm capsule](https://github.com/tock/tock/blob/master/capsules/core/src/alarm.rs#L91-L107).

For this case where we're setting the last alarm that won't overflow, this will be fine. The user's callback will get scheduled as soon as possible due to the behavior of the alarm capsule.

examples/tests/timer_overflow/main.c
atar13: Addressed here: https://github.com/tock/libtock-c/pull/395/commits/39c4c5b30cee19c8a396ceff55df7b6e579de94f

examples/tests/timer_overflow/main.c
tyler-potyondy: 1. Once the left aligned frequency changes are merged (https://github.com/tock/tock/pull/3975), we will be able to determine the wrap point. I don't think we should block on waiting for that PR. For now, I think it's reasonable to just set this to some "large value" i.e. 10 minutes and place a comment saying this time value is board specific. 

2. Yes, this can be configured within the `main.rs` initialization.

libtock/alarm_timer.c
```diff
@@ -4,11 +4,131 @@
...
+  // This calculation is mathematically equivalent to doing:
+  //
+  //   `leftover_ticks` / (`frequency` / `milliseconds_per_second`)
+  //
+  // This is taking the remaining unconverted ticks (leftover_ticks)
+  // and dividing by the number of ticks per millisecond
+  // (`frequency` (ticks per second) / `1000` milliseconds per second)
+  // The division is done this way because of the same argument in
+  // `ms_to_ticks`.
+  milliseconds += (leftover_ticks * milliseconds_per_second) / frequency;
```
tyler-potyondy, line 41: `leftover_ticks` is at most equal to `frequency`. For frequency values larger than 4MHz, I believe this calculation will overflow. I do not think there is a way around having a `uint64_t` to handle the intermediate calculations.

Using a `uint64_t` would allow us to support up to and beyond GHz frequencies.

libtock/alarm_timer.c
```diff
@@ -4,11 +4,131 @@
...
+  return milliseconds;
+}
+
+// WARNING: This function can silently overflow if the output
+// number of ticks is larger than `UINT32_MAX`.
```
tyler-potyondy, line 46: This does not seem ideal considering that libtock-c assumes:

```
 * The client should not assume anything about the underlying clock used by an
 * implementation other than that it is running at sufficient frequency to
 * deliver at least millisecond granularity and that it is a 32-bit clock (i.e.
 * it will wrap at 2^32 clock ticks).
 ```

Perhaps it would be better to pass a `uint32_t` reference that this function updates and instead returns a returncode?

File: libtock/alarm_timer.c [hunk 14]
```diff
@@ -4,11 +4,131 @@
...
+  // This multiplication has no error because the board frequency is given in Hertz (ticks per second).
+  uint32_t ticks = seconds * frequency;
+
+  // Now that we have converted the full seconds part of `ms` to ticks, we must
+  // convert the `leftover_millis`.
+  // The reason for splitting this conversion rather than doing a direct conversion
+  // is due to 32-bit overflow. `leftover_millis` can only take values in the range
+  // [0, frequency). Thus, this multiplication is upper bounded by `frequency`^2. This
+  // calculation will not overflow as long as
+  //   `frequency`^2 < `UINT32_MAX` -> frequency < 65535.99 -> frequency <= 2^16 - 1
```
tyler-potyondy, line 89: Same as above, I think we want this to support frequencies at least up to 100MHz considering the kernel currently provides enum variants for frequencies up to 100MHz.

I do not think there is a way around using a `u64` here.

tyler-potyondy (changes requested):

libtock/timer.h
ppannuto: `Due to no reference time,`, can we remove that language and replace it with the explanation of precision limitations [once those are more solidly established]?

libtock/alarm_timer.c
Samir-Rashid: Making the calculation use `uint64_t`s will not solve this issue because the timer interface is 32-bit. This function *must* return a 32-bit integer. It is a separate concern whether a 64-bit timer interface should be added.

libtock/alarm_timer.c
Samir-Rashid: This is fixed in the separated PR. Also, I made an error in the calculation here which is also fixed.

bradjc:
What is left to do here? Move code from openthread to the general alarm code?

libtock/alarm_timer.c
lschuermann: I'm just browsing through all this, and I think I agree with the logic here. Using `uint64_t` would also allow us to simply the logic here a bunch, and remove the split between seconds and fractional seconds, effectively making this equivalent to: `ticks_to_ms = (ticks * 1000) / freq`.

I'm wondering how much the split into seconds and fractional seconds buys us here in terms of performance in code size, vs. 64-bit arithmetic on a 32-bit platform.

lschuermann: Nevermind, I just saw that the new, merged PR has been changed to 64-bit, my bad. Part of my comment still applies, I suppose: https://github.com/tock/libtock-c/pull/434/files#r1642913413

Samir-Rashid:
Rebased on master.

libtock/services/alarm.h
```diff
@@ -47,38 +47,50 @@ typedef struct alarm {
...
+  // before hitting the target interval_ms.
+  uint32_t overflows_left;
+  // Number of ticks remaining after the last time the
+  // counter overflows.
+  uint32_t remaining_ticks;
+  libtock_alarm_callback callback;
+  void* user_data;
   libtock_alarm_t alarm;
-} libtock_alarm_repeating_t;
+} libtock_alarm_data_t;
```
bradjc, line 65: Why does this modify `libtock_alarm_repeating_t` and not `libtock_alarm_t`? Couldn't a single alarm also have an overflow?

Samir-Rashid: `libtock_alarm_at` sets a single alarm and takes in a 32 bit argument (`dt`) for the length of the alarm in ticks. This cannot overflow as the kernel guarantees that the alarm buffer will not overflow sooner than 32 bits of ticks. Only alarms longer than $2^{32} - 1$ ticks can overflow. These alarms can only be set by `libtock_alarm_in_ms`.

bradjc: Ok that makes sense.

What about renaming `libtock_alarm_t` to something like `libtock_alarm_dt_t`, creating a new `libtock_alarm_t` that supports overflows, and leaving `libtock_alarm_repeating_t` as it is? It would be nice to not have to change all of the examples and to keep the more descriptive "repeating" name. Changing the type for `libtock_alarm_at()` seems the least invasive since it isn't really used.

Samir-Rashid: I have restored the `libtock_alarm_t` name to make the renaming less invasive. We cannot keep the `libtock_alarm_repeating_t` unchanged for a few reasons:
- `libtock_alarm_repeating_t` sets repeating alarms in milliseconds, so it must support overflows. Thus we must add extra fields to support the new overflow logic.
-  `libtock_alarm_repeating_every_ms` calls `libtock_alarm_in_ms` with the alarm struct it was given. To support the less intrusive naming, both must be the same type name `libtock_alarm_t`. As the struct `libtock_alarm_repeating_t` is not relevant to the user calling these functions, I do not think it is more helpful to include the "repeating" moniker. This is why I believe there should only be two structs and why the name of the `repeating` struct must be changed.

bradjc: Ok makes sense I agree.

bradjc (dismissed):

bradjc:
@hudson-ayers Do you think there is any reason to keep the `libtock_alarm_repeating_t` type around? It would essentially be an alias for `libtock_alarm_t` right now, but if we need to keep more state for repeating alarms in the future then it would be helpful. But I can't think of what that state might be.

hudson-ayers:
I think it is fine to get rid of `libtock_alarm_repeating_t` with this change since they would become identical, I cannot think of any other data we would really want to add to repeating alarms.

libtock/services/alarm.h
```diff
@@ -109,19 +127,19 @@ int libtock_alarm_in_ms(uint32_t ms, libtock_alarm_callback cb, void* opaque, li
...
- * \param a pointer to a new alarm_repeating_t to be used by the implementation to
+ * \param alarm_repeating pointer to a new libtock_alarm_t to be used by the implementation to
  *        keep track of the alarm.
  */
-void libtock_alarm_repeating_every(uint32_t ms, libtock_alarm_callback cb, void* opaque,
-                                   libtock_alarm_repeating_t* alarm_repeating);
+void libtock_alarm_repeating_every_ms(uint32_t ms, libtock_alarm_callback cb, void* opaque,
+                                      libtock_alarm_t* alarm);
 
 /** \brief Cancels an existing repeating alarm.
```
hudson-ayers, line 138: This comment is now outdated, since this function is used to cancel repeating alarms or any alarm set in milliseconds.

bradjc (approved):

alevy (approved):

tyler-potyondy (approved):

PR #397, opened 2024-04-10
make: apps: app compilation depends on libs

bradjc:
The library must exist before the app can compile in case the app needs headers from the library.


ppannuto (approved):

ppannuto:
IIUC, this is only useful for the case where we have rules that will automatically source the library, which I think is still a semi-open question. Otherwise this is the difference between a compiler error of 'missing header' vs make error of 'no rule to make X', but the latter is probably a bit cleaner in either case.

PR #400, opened 2024-04-12
port spi to new libtock-c design + rename to spi_controller

hudson-ayers:
Targets the libtock-folders branch. Posted here for review.


libtock-sync/peripherals/spi_controller.c
```diff
@@ -0,0 +1,49 @@
+#include "spi.h"
```
bradjc, line 1:
```suggestion
#include "spi_controller.h"
```

PR #401, opened 2024-04-12
move lora examples to wip directory

bradjc:
The main change here is moving the lora examples to a folder where they are not on the critical path for building and passing CI. For #370 we are updating libtock, but these apps use code that references libtock in a different library. To make progress on #370 this makes them not have to build.

The folder doesn't have to be called "legacy", that is just the first idea I had.


ppannuto (commented):
Moving seems pragmatic. 

Re name: I think this is still somewhat fresh and recent work from Alistair, so legacy perhaps misleading.

Maybe something more explicit like a "wip" folder, with readme explaining stuff in that folder isn't tested by CI or vetted as heavily, and a short explanation in the Lora example readme of why it's wip?

alistair23:
I'm not sure I follow. Why remove this from the CI?

The LoRa stuff works really well. It's probably the most reliable wireless support Tock has. It's also not WIP, the actual LoRa code is very robust. I have a [PR open](https://github.com/tock/libtock-c/pull/394) to transmit GNSS data as well

bradjc:
> Why remove this from the CI?

We need it to support RISC-V, and we need to be able to update libtock function names without patching a different library.

alistair23:
Do you have a RISC-V board with LoRa? There isn't any reason it won't support RISC-V

Are we really changing the libtock-c APIs so often that we can't call them from external libraries? That makes using libtock-c for anything besides test examples really hard

bradjc:
> Are we really changing the libtock-c APIs so often that we can't call them from external libraries?

Yes and no. We are changing them (they really need to be updated). But an external library can pin to a version. The issue is the circle in the lora examples.

ppannuto (approved):
I think so long as we want a policy of "risc-v builds are on by default", then an example app that doesn't work on risc-v would necessarily qualify as wip [i.e., until it works on "all officially supported archs"].

PR #402, opened 2024-04-16
make: include library paths only for library build

bradjc:
In the rewrite, we want apps to do:

```
#include <libtock/sensors/temperature.h>
```

so we want to ensure that apps are not compiled with `-I` flags pointing to directories internal to the library. We do that by adding the include paths to the `CPPFLAGS_$(LIBNAME)` variable. That way those paths are only included during the library build.

Now, our apps currently expect access to headers within libtock/, so we workaround that for now. That will be removed in the rewrite.

Also fix an issue with the include paths having double `/`.


simple-ble/Makefile.app
```diff
@@ -0,0 +1,5 @@
+$(info HKJFDHSJKHFDSKJL)
```
brghena, line 1: I think this slipped by you

bradjc: ha

brghena (approved):
Looks good

This will be a positive change for header management.

PR #405, opened 2024-04-18
Real targets for LINK_LIBS (now SYSTEM_LIBS)

ppannuto:
This subsumes #398 and #399.

It has basically the same goals as those two PRs, but the implementation is a bit more generic. In particular, we no longer need to include (and remember to update!) any information in the library rule generation about valid versions for newlib or libc++.

This should correctly express all library targets. Also important, it correctly groups all targets which come from one pre-built library such that if you run a parallel make, you don't kick of multiple jobs fetching the same archive.

Tested fairly heavily with all the permutations of deleted libraries, random on/off of `TOCK_TARGETS` entries to stress re-builds where arch's need, varying degrees of parallelization of the build, and anything else I could think of.


TockLibrary.mk
```diff
@@ -113,19 +113,19 @@ $$($(LIBNAME)_BUILDDIR)/$(1)/$(LIBNAME):
 	$$(TRACE_DIR)
 	$$(Q)mkdir -p $$@
 
-$$($(LIBNAME)_BUILDDIR)/$(1)/$(LIBNAME)/%.o: %.cpp | $$($(LIBNAME)_BUILDDIR)/$(1)/$(LIBNAME) newlib-$$(NEWLIB_VERSION_$(1))
+$$($(LIBNAME)_BUILDDIR)/$(1)/$(LIBNAME)/%.o: %.cpp | $$($(LIBNAME)_BUILDDIR)/$(1)/$(LIBNAME) $$(LINK_LIBS_$(1))
```
bradjc, line 116:
```suggestion
$$($(LIBNAME)_BUILDDIR)/$(1)/$(LIBNAME)/%.o: %.cpp | $$($(LIBNAME)_BUILDDIR)/$(1)/$(LIBNAME) $$(SYSTEM_LIBS_$(1))
```
?

ppannuto: ugh, late night rebase flub, sorry — fixed.

bradjc (approved):
I think I am partial to the approach in #398 of just using the version number in Precompiled.mk (rather than the wildcard) and avoid adding the lib/newlib and lib/libc++ directories. But I haven't tested if that is even possible.

In any case I don't feel that strongly about it and this is still movement in the right direction.

bradjc (approved):

PR #406, opened 2024-04-19
make: use `$(TOCK_USERLAND_BASE_DIR)` as ROOT

bradjc:
The intent of `$($(LIBNAME)_SRC_ROOT)` was to have a path prefix that was unique for each library, so even different libraries with the same file names would compile as expected. This approach worked fine for libraries like openthread and u8g2, but for libtock and libtock-sync it does not. Those libraries have the same paths and the same files.

In hindsight, it's probably better to just use the libtock-c folder as the root, and base everything off of that. Therefore every file has a unique path.

This effectively reverts adding `$($(LIBNAME)_SRC_ROOT)` and instead just uses `$(TOCK_USERLAND_BASE_DIR)` in its place. Crucially, this allows libtock and libtock-sync to compile side-by-side.

### DIFF
--- a/TockLibrary.mk
+++ b/TockLibrary.mk
@@ -24,15 +24,9 @@ include $(TOCK_USERLAND_BASE_DIR)/Precompiled.mk
 # - `$($(LIBNAME)_DIR)`: The path to the directory the library Makefile is
 #   stored in. This is typically based on the `$(TOCK_USERLAND_BASE_DIR)`
 #   variable.
-# - `$($(LIBNAME)_SRC_ROOT)`: The path to the directory where the library sources
-#   are stored. This is typically based on the `$(TOCK_USERLAND_BASE_DIR)`
-#   variable.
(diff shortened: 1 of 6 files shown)


TockLibrary.mk
```diff
@@ -74,9 +68,9 @@ $(LIBNAME)_BUILDDIR := $($(LIBNAME)_DIR)/build
 $(LIBNAME)_SRCS_FLAT := $(notdir $($(LIBNAME)_SRCS))
 $(LIBNAME)_SRCS_DIRS := $(sort $(dir $($(LIBNAME)_SRCS))) # sort removes duplicates
 
-# Only use vpath for certain types of files
-# But must be a global list
-VPATH_DIRS += $($(LIBNAME)_SRC_ROOT)
+# # Only use vpath for certain types of files
+# # But must be a global list
```
ppannuto, line 72:
```suggestion
# Only use vpath for certain types of files
# But must be a global list
```

ppannuto (approved):

PR #407, opened 2024-04-22
update libnrfserialization to modern libtock-c make

bradjc:
The libnrfserialization library has been around awhile, and it used a rather bespoke build setup. This updates it to match the current format.

I included a new copy of the built output. We can remove that.

For users, the only change is that instead of unpacking our custom headers the build system downloads a zip of the nrf5x-base repo. If we remove the built output entirely then this system will build the library files.


bradjc:
One issue is that if we do leave the .a files committed, then Makefile.setup doesn't run, and we don't get the source, and so the apps won't find the headers.

ppannuto (approved):
I agree that removing the built archives is the way to go and makes things robust; a one-off build cost isn't bad.

PR #408, opened 2024-04-24
Update alarm to new libtock-c API

hudson-ayers:
This PR updates the alarm / timer APIs to the new libtock-c API.

### Important notes:

- The raw alarm system calls are moved into the peripherals folder, and the name `alarm.h` is used to mirror the corresponding Tock capsule
- Virtualized userspace alarms and timers, previously defined in `alarm.h` and `timer.h` and implemented in `alarm_timer.c`, now live in a new "services" folder, which is intended to hold APIs that hold internal state even in their asynchronous implementations.
- There is no longer a seperate `alarm.h` / `timer.h`, just using a single `alarm.h` and `alarm.c`. This made it easier to follow the new naming guide.
- Renamed "tock_timer_t" to "alarm_repeating_t", and replaced `timer_in()` with `libtock_alarm_at()` and `timer_every()` with `libtock_alarm_repeating_every()`. I understand that previously we had `alarm_at()` and `timer_in()` for absolute and relative alarms, respectively, because of the semantic meaning of those two words. But we also conflated the word "timer" with "repeating alarm", which led to an (IMO) confusing API that often required users to waste stack space allocating `tock_timer_t` structs just to call `timer_in()` which would ignore everything in `tock_timer_t` except its alarm field anyway. Also, it is really weird that in the old API `timer_in()` accepted a timer parameter but returned an alarm, while `timer_every()` accepted a timer parameter and returned a timer...
- `delay_ms()` and `yield_for_with_timeout()` moved into libtock_sync because they yield. I kept the old names for `delay_ms()` and `yield_for_with_timeout()` for now, mostly because every single example getting filled with calls to `libtock_sync_alarm_delay_ms()` felt painful...but I can change this if people feel otherwise!
- I fixed up libtock/unit_test.c and libtock/net/ieee802154.c to use the new interfaces, but only fixed the blink example so far. I will fix the rest if people generally approve this design.

## Potential Issue
This PR does not use the new callback format. The primary reason for this is that the virtual alarm implementation relies on `tock_enqueue()` in `libtock/tock.c`, which stores both a callback pointer and a userdata pointer and uses those such that callbacks are executed correctly when `yield()` is called. It seems to me that the new callback format is incompatible with this design, where tock_enqueue needs to be able to store (and eventually call) callbacks of arbitrary signatures.

**Currently, alarm is the only API that uses `tock_enqueue()`, so in theory we could still use the new API by having tock.h depend on `services/alarm.h`, and making it explicit that `tock_enqueue()` is only for alarms, not other callbacks. I am open to this if people think that is a better approach**


libtock/services/alarm.h
```diff
@@ -0,0 +1,124 @@
...
+//
+// See https://github.com/tock/libtock-c/pull/355#issuecomment-1841351091 for
+// more details
+//
+// ```c
+// int _gettimeofday(struct timeval *tv, void *tzvp) {
+//   return libtock_gettimeasticks(tv, tzvp);
+// }
+// ```
+int libtock_gettimeasticks(struct timeval *tv, void *tzvp);
```
bradjc, line 91: Not namespaced in `alarm`?

bradjc (commented):
I would like to see the clunky doxygen comments removed in favor of the same style as rustdoc.

Re: callbacks: the new design avoids the overhead of libtock users having to figure out what four random u32s mean. The kernel upcall mechanism requires arbitrary upcall fn signatures, why can't tock_enqueue? Can't tock_dequeue call back into alarm which then calls the user callback?

libtock/unit_test.c
```diff
@@ -230,7 +231,7 @@ void unit_test_runner(unit_test_fun *tests, uint32_t test_count,
 
   // Establish communication with the test supervisor service. First delay 10 ms
   // to ensure the supervisor service has time to register.
-  delay_ms(10);
+  //delay_ms(10); // TODO: Restore me once unit test is ported and moved
```
Samir-Rashid, line 234: Fix TODO.

Samir-Rashid (commented):
While you are refactoring all the function names, can we improve the names of these functions?
- `libtock_gettimeasticks`: No reason to mirror the `gettimeofday` naming if it doesn't do the same thing. Also, this does not actually return the *time as ticks*. We should rename this to `libtock_gettime`.
- `libtock_alarm_at` and `libtock_alarm_in` are very similarly named. It is not obvious why these functions differ. `libtock_alarm_in` could be `libtock_alarm_ms` to match `delay_ms`. It may also make sense to have `libtock_alarm_at` be `libtock_alarm_ticks`.

libtock/unit_test.c
hudson-ayers: That is a TODO intended to be resolved outside of this PR -- this PR targets the libtock-folders branch, and the PR for that branch (https://github.com/tock/libtock-c/pull/370) tracks the remaining drivers that need to be ported, including this one.

hudson-ayers:
> While you are refactoring all the function names, can we improve the names of these functions?
> 
> * `libtock_gettimeasticks`: No reason to mirror the `gettimeofday` naming if it doesn't do the same thing. Also, this does not actually return the _time as ticks_. We should rename this to `libtock_gettime`.

@bradjc originally suggested gettimeasticks, I am happy with whatever. gettime is a little misleading since it does not return the time. Maybe `libtock_alarm_get_time_since_boot()`?

> * `libtock_alarm_at` and `libtock_alarm_in` are very similarly named. It is not obvious why these functions differ. `libtock_alarm_in` could be `libtock_alarm_ms` to match `delay_ms`. It may also make sense to have `libtock_alarm_at` be `libtock_alarm_ticks`.

yeah, I was renaming it from libtock_timer_in to libtock_alarm_in to better match the new naming scheme, but I see how it is confusing that it takes different measurement of time as input, in addition to being a relative time instead of an absolute time.

I like `libtock_alarm_at()`, but would be fine to change `libtock_alarm_in()` to `libtock_alarm_in_ms()`. Thoughts?

bradjc:
> I like `libtock_alarm_at()`, but would be fine to change `libtock_alarm_in()` to `libtock_alarm_in_ms()`. Thoughts?

I came to the same idea, so I agree.

> > * `libtock_gettimeasticks`: No reason to mirror the `gettimeofday` naming if it doesn't do the same thing. Also, this does not actually return the _time as ticks_. We should rename this to `libtock_gettime`.
> 
> @bradjc originally suggested gettimeasticks, I am happy with whatever. gettime is a little misleading since it does not return the time. Maybe `libtock_alarm_get_time_since_boot()`?

Does anyone actually know what this function does or is supposed to do? I think that is why it is hard to name. It seems like `libtock_alarm_get_a_time()` might be appropriate. Is there any guarantee the alarm counter hasn't paused for some time? Or that it won't wrap around? Or that it was started at any particular time? Or that it is incrementing at all?

libtock/services/alarm.h
hudson-ayers: Fixed in new PR.

hudson-ayers:
I renamed libtock_alarm_in --> libtock_alarm_in_ms in my new PR. I stuck with libtock_alarm_gettimeasticks because I don't know what is right and the bikeshedding is out of scope for this PR, especially since the function is barely used anyway.

Closing this PR in favor of the new/updated design in https://github.com/tock/libtock-c/pull/413

PR #412, opened 2024-05-02
lora: Build lora examples for RISC-V

alistair23:


bradjc:
Can we move to using the TockLibrary.mk file so that we get all of the relocation flags, etc.?

alistair23:
We can. I'd like to get this going first so we have a known working state for RISC-V and ARM then switch to `TockLibrary.mk`

ppannuto:
I pushed a few small robustness / correctness changes.

There's something broken in whatever makefile cmake generates (if it were correct, we could just specify `make RadioLib/libRadioLib.a` instead of the let-the-build-fail hack), but in lieu of fixing others' broken build systems, I just put an explicit check in that the build got far enough to make the file we care about.

---

In related news, for reasons that it's too late in the evening to debug on my machine, the arm build doesn't work?

```
...
[  1%] Building CXX object RadioLib/CMakeFiles/RadioLib.dir/src/ArduinoHal.cpp.o
In file included from /Users/ppannuto/code/helena-project/libtock-c/wip/lora/RadioLib/src/TypeDef.h:6,
                 from /Users/ppannuto/code/helena-project/libtock-c/wip/lora/RadioLib/src/ArduinoHal.h:2,
                 from /Users/ppannuto/code/helena-project/libtock-c/wip/lora/RadioLib/src/ArduinoHal.cpp:1:
/Users/ppannuto/code/helena-project/libtock-c/wip/lora/RadioLib/src/BuildOpt.h:127:12: fatal error: stdio.h: No such file or directory
  127 |   #include <stdio.h>
      |            ^~~~~~~~~
compilation terminated.
```

Could be a local toolchain thing somehow, but nothing else fails to build on my machine at the moment, so it feels likely to somehow be specific to the build invocation here.

Nothing special about my arm gcc install, just standard brew:
```
$ arm-none-eabi-gcc --version
arm-none-eabi-gcc (GCC) 13.2.0
Copyright (C) 2023 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

$ brew info arm-none-eabi-gcc
==> arm-none-eabi-gcc: stable 13.2.0 (bottled)
GNU compiler collection for arm-none-eabi
https://gcc.gnu.org
Installed
/usr/local/Cellar/arm-none-eabi-gcc/13.2.0 (567 files, 181.8MB) *
  Poured from bottle using the formulae.brew.sh API on 2024-04-17 at 15:45:31
From: https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/a/arm-none-eabi-gcc.rb
License: GPL-3.0-or-later with GCC-exception-3.1
==> Dependencies
Required: arm-none-eabi-binutils ✔, gmp ✔, isl ✔, libmpc ✔, mpfr ✔, zstd ✔
==> Analytics
install: 494 (30 days), 1,917 (90 days), 6,508 (365 days)
install-on-request: 471 (30 days), 1,786 (90 days), 6,169 (365 days)
build-error: 5 (30 days)
```

Can try to dig into that more another day if the arm build failure isn't reproducible for others.

alistair23:
> Could be a local toolchain thing somehow, but nothing else fails to build on my machine at the moment, so it feels likely to somehow be specific to the build invocation here.

Strange. It works as part of the CI: https://github.com/jgromes/RadioLib/blob/master/.github/workflows/main.yml#L216

alistair23:
RadioLib uses CMake. Any idea how to use `TockLibrary.mk` with that? @bradjc

bradjc:
Should just boil down to listing where the sources are and setting any needed `#define`s. We did it for openthread (which also uses cmake upstream): https://github.com/tock/libtock-c/blob/master/libopenthread/libopenthread-mtd/Makefile

PR #414, opened 2024-05-06
Resolve gettimeasticks imprecision

tyler-potyondy:
The prior implementation faced a trade off of imprecise time conversion or overflowing values. To prevent overflowing values, microseconds were calculated as `tv->tv_usec = (remainder * 1000) / (frequency / 1000);`. Scaling $\frac{frequency}{ 1000}$ losses the frequency's precision and introduces error in the conversion. 

In trying to fix this, I attempted to first convert to milliseconds, followed by an additional remainder calculation (using uint32). However, this also faced an issue of the intermediate calculation overflowing for frequencies greater than 4MHz. 

This updates `gettimeasticks` to use a uint64_t for calculations to avoid overflows and minimize loss of precision. If there are strong objections to introducing a uint64 here, we can also only perform the uint64 operation for frequencies that would overflow the uint32.

### DIFF
--- a/libtock/services/alarm.c
+++ b/libtock/services/alarm.c
@@ -171,18 +171,12 @@ void libtock_alarm_repeating_cancel(libtock_alarm_repeating_t* repeating) {
 int libtock_alarm_gettimeasticks(struct timeval *tv, __attribute__ ((unused)) void *tzvp)
 {
   uint32_t frequency, now, seconds, remainder;
+  const uint32_t microsecond_scaler = 1000000;
 
   libtock_alarm_command_get_frequency(&frequency);
   libtock_alarm_command_read(&now);
 
-  // Confirm frequency assumptions. Since remainder is
-  // ticks % frequency, remainder <= frequency. This assertion
-  // guards against overflowing the us calculation. At worst,
-  // remainder == frequency and scaling remainder to convert
-  // to us will overflow. We confirm the greatest possible
-  // remainder value (frequency) scaled by 1e6 will fit within
-  // a uint64 and guard against division by zero.
-  assert(frequency > 0 && (frequency * 1e6) <= UINT64_MAX);
+  assert(frequency > 0);
 
   // Obtain seconds and remainder due to integer divison
   seconds   = now / frequency;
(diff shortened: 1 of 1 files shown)


libtock/alarm_timer.c
```diff
@@ -215,33 +215,37 @@ int yield_for_with_timeout(bool* cond, uint32_t ms) {
...
-  // Confirm frequency assumption
-  assert(frequency > 0);
+  // Confirm frequency assumptions. Since remainder is
+  // ticks % frequency, remainder <= frequency. This assertion
+  // guards against overflowing the us calculation. At worst,
+  // remainder == frequency and scaling remainder to convert
+  // to us will overflow. We confirm the greatest possible
+  // remainder value (frequency) scaled by 1e6 will fit within
+  // a uint64 and guard against division by zero.
+  assert(frequency > 0 && (frequency * 1e6) <= UINT64_MAX);
```
alistair23, line 228: `frequency` is still a 32-bit value, so won't it always be less then `UINT64_MAX`?

libtock/alarm_timer.c
```diff
@@ -215,33 +215,37 @@ int yield_for_with_timeout(bool* cond, uint32_t ms) {
...
-  // of precision as the 3 least significant digits do not encode data. The only case of a lose
-  // in precision is for the frequency 32.768KHz. In this case, the loss of precision introduces ~1us
-  // of error.
-  tv->tv_sec  = seconds;
-  tv->tv_usec = (remainder * 1000) / (frequency / 1000);
+  // (ticks) * (1e6 us / s) * (s / ticks) = us
+  // Because remainder is by definition less than
+  // frequency, we must be sure to first scale remainder
+  // or else floor(remainder / frequency) = 0.
+  uint64_t us = remainder * 1e6 / frequency;
```
alistair23, line 238: I think these need to be cast to `uint64_t` to ensure they are treated as 64-bits

libtock/alarm_timer.c
ppannuto:
```suggestion
  assert(frequency > 0ull && (frequency * 1e6ull) <= UINT64_MAX);
```

I believe Alistair is correct here, though you can [make the constants 64-bit types](https://stackoverflow.com/questions/2835746/how-do-you-specify-a-64-bit-unsigned-int-const-0x8000000000000000-in-vs2008), and [standard type promotion](https://c0x.shape-of-code.com/6.3.1.8.html) should do the right thing from there.

libtock/alarm_timer.c
ppannuto: The plain `1e6` here and above [probably isn't doing what you want either, and is generating a double](https://stackoverflow.com/a/40465456) (incidentally, everything being promoted to doubles might be why this passed your tests, but isn't really what you were trying to do).

I think that somewhat verbose, explicit casts are warranted given the complexities and subtleties here, and the silent fungibility of C types.

ppannuto (changes requested):

libtock/alarm_timer.c
tyler-potyondy: I think we can remove the overflow check given this will never overflow (so long as the convention remains of frequency being a `uint32_t`

ppannuto:
```
#if __has_builtin(__builtin_types_compatible_p)
static_assert(__builtin_types_compatible_p(uint32_t, typeof(frequency))
#endif
```

You can always [enforce that assumption](https://stackoverflow.com/a/55804363) :)

ppannuto (approved):
Reasoned through this a few times, I believe this is all correct.

PR #415, opened 2024-05-08
hotp: fix to new libtock and kv

bradjc:
Several updates needed after testing all of the apps.

### DIFF
--- a/examples/tutorials/hotp/hotp_milestone_one/main.c
+++ b/examples/tutorials/hotp/hotp_milestone_one/main.c
@@ -55,7 +55,7 @@ static void program_new_secret(hotp_key_t* hotp_key) {
     libtocksync_console_read((uint8_t*) &c, 1, &number_read);
 
     // break on enter
-    if (c == '\n') {
+    if (c == '\n' || c == '\r') {
       break;
     }
 
@@ -100,7 +100,7 @@ int main(void) {
       "* Hold Button 1 to enter a new HOTP secret for that slot.\r\n");
 
   // Initialize buttons
-  if (initialize_buttons() != RETURNCODE_SUCCESS) {
+  if (initialize_buttons(false) != RETURNCODE_SUCCESS) {
     printf("ERROR initializing buttons\r\n");
     return 1;
   }
@@ -111,6 +111,7 @@ int main(void) {
   // Main loop. Waits for button presses
   while (true) {
     // Yield until a button is pressed
+    button_pressed = false;
     yield_for(&button_pressed);
     int btn_num = pressed_btn_num;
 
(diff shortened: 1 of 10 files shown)


examples/tutorials/hotp/hotp_milestone_one/main.c
```diff
@@ -55,7 +55,7 @@ static void program_new_secret(hotp_key_t* hotp_key) {
     libtocksync_console_read((uint8_t*) &c, 1, &number_read);
 
     // break on enter
-    if (c == '\n') {
```
ppannuto, line 58: I suspect/fear this fixes mac but breaks windows. I'm assuming the issue is that [legacy mac tools send just `\r`](https://blog.codinghorror.com/the-great-newline-schism/)? Now, though, with a DOS endpoint that sends `\r\n` this will break twice when it should only break once [previously the `\r` was just silently consumed IIUC].

examples/tutorials/hotp/hotp_milestone_two/step1.c
```diff
@@ -25,7 +25,7 @@ void program_new_secret(hotp_key_t* hotp_key) {
     libtocksync_console_read((uint8_t*) &c, 1, &number_read);
 
     // break on enter
-    if (c == '\n') {
```
ppannuto, line 28: (same)

examples/tutorials/hotp/hotp_oracle_complete/main.c
```diff
@@ -158,7 +158,7 @@ static void program_new_secret(int slot_num) {
     libtocksync_console_read((uint8_t*) &c, 1, &number_read);
 
     // break on enter
-    if (c == '\n') {
```
ppannuto, line 161: (same)

ppannuto (changes requested):
I'm so sorry — this is an annoying problem. Might just be worth a comment or an `#ifdef DOS_LINE_ENDINGS` or something given the context of when/where this will be used.

examples/tutorials/hotp/hotp_milestone_one/main.c
bradjc: This doesn't have to do with my tockloader not sending `\n` (I can run another console recv app and verify I'm getting both `\r` and `\n`).

The issue is that they are both sent, and sent at the same time, and the loop to the next console recv is too slow to catch the `\n` (since it only reads one byte).

ppannuto: I don't quite understand why being slow should matter, wouldn't the next character just be buffered and read the next time through the loop?

That said, this is a bugfix in an open tutorial, so I'm inclined to merge this for now and revisit the timing question another day.

ppannuto (approved):

PR #416, opened 2024-05-08
Add library headers to formatting

bradjc:
We are currently only formatting source files. This adds headers and does the format pass.


libtock-sync/services/unit_test.h
```diff
@@ -86,12 +86,12 @@ extern "C" {
...
  */
 typedef struct unit_test_fun_s {
-    bool (*fun)(void);
-    char name[24];
+  bool (*fun)(void);
+  char name[24];
 } unit_test_fun;
 
 #define TEST(NAME) { \
-    test_##NAME,     \
```
ppannuto, line 94: I _assume_ this doesn't change anything behaviorally, but C-macro magic like this always makes me want to look twice

libtock/tock.h
```diff
@@ -22,57 +22,57 @@ typedef void (subscribe_upcall)(int, int, int, void*);
...
+  TOCK_SYSCALL_FAILURE =   0,
+  TOCK_SYSCALL_FAILURE_U32 =   1,
+  TOCK_SYSCALL_FAILURE_U32_U32 =   2,
+  TOCK_SYSCALL_FAILURE_U64 =   3,
+  TOCK_SYSCALL_SUCCESS = 128,
+  TOCK_SYSCALL_SUCCESS_U32 = 129,
+  TOCK_SYSCALL_SUCCESS_U32_U32 = 130,
+  TOCK_SYSCALL_SUCCESS_U64 = 131,
   TOCK_SYSCALL_SUCCESS_U32_U32_U32 = 132,
-  TOCK_SYSCALL_SUCCESS_U32_U64     = 133
```
ppannuto, line 34: :(, this one (and below) feels like a regression

is there some kind of escape hatch comment from uncrustify we could apply here?

ppannuto (changes requested):
I always get such satisfaction out of more automation of arbitrary things.

---

Two comments; they aren't really the blocker here as much as wanting to get the all clear from @tyler-potyondy that there aren't any material pending changes needed from tutorial stuff that the formatting patch would cause needless conflict with.

libtock-sync/services/unit_test.h
bradjc: Added `sp_pp_concat                    = remove`

ppannuto (dismissed):
🎉 , awesome; marking as blocked until after next week's tutorial.

ppannuto (approved):

PR #417, opened 2024-05-08
Add uncrustify options (2024 edition)

bradjc:
Unfortunately uncrustify does not work like rustfmt, as it does not have preferred style that it defaults to, and then allows customization. Instead, it does nothing by default, and allows you to tell it what it to do. That means a lot of hunting for options to try to get to a consistent style. This is one step closer to having more consistent formatting.


ppannuto (dismissed):
If anyone else is motivated to look, `?w=1` on the url (ignore whitespace) is your friend once you've got a sense of what the changes are doing. Only 9 files have non-simple space changes, and those are almost all just `{`'s moving up a line.

---

Also marking this one as blocked until after next week's tutorial.

bradjc:
Ok this is ready again.

ppannuto (approved):

PR #421, opened 2024-05-09
libtock/ipc: add custom IPC service callback type with buffer pointer

lschuermann:
This changes the IPC service callback function type to provide the IPC buffer pointer as an actual pointer type. The fact that this is currently an integer is pretty weird, and appears to be an artifact of the implementation sharing the subscribe system call wrapper.

I'm not sure whether what I'm doing here works, or is legal. In fact, there's a chance that it's relying on undefined behavior. However, conveying pointers as `int` is also not exactly elegant. A proper fix may be to duplicate the `subscribe` system call wrapper implementation and declare it with a different type that accepts this upcall function signature.

### DIFF
--- a/examples/rot13_service/main.c
+++ b/examples/rot13_service/main.c
@@ -6,8 +6,8 @@ struct rot13_buf {
   char buf[31];
 };
 
-static void rot13_callback(int pid, int len, int buf, __attribute__ ((unused)) void* ud) {
-  struct rot13_buf *rb = (struct rot13_buf*)buf;
+static void rot13_callback(int pid, int len, void *buf, __attribute__ ((unused)) void* ud) {
+  struct rot13_buf *rb = buf;
   int length = rb->length;
   if (length > len - 1) {
     length = len - 1;
--- a/examples/services/ble-env-sense/main.c
+++ b/examples/services/ble-env-sense/main.c
@@ -94,7 +94,7 @@ typedef struct {
 } sensor_update_t;
 
 
-static void ipc_callback(int pid, int len, int buf, __attribute__ ((unused)) void* ud) {
+static void ipc_callback(int pid, int len, void *buf, __attribute__ ((unused)) void* ud) {
   if (len < (int) sizeof(sensor_update_t)) {
     printf("Error! IPC message too short.\n");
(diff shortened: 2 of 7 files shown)


ppannuto:
Shoving integers into pointers is a time-honored tradition in C. [From the example I'm most familiar with (GLib)](https://docs.gtk.org/glib/conversion-macros.html), however:

> Warning: You may not store pointers in integers. This is not portable in any way, shape or form. These macros only allow storing integers in pointers, and only preserve 32 bits of the integer; values outside the range of a 32-bit integer will be mangled.

If I understand correctly, the material part of the question boils down to:

```c
typedef void (ipc_service_upcall)(int, int, void*, void*);
typedef void (subscribe_upcall)(int, int, int, void*);

subscribe_return_t sval = subscribe(IPC_DRIVER_NUM, svc_id, *((subscribe_upcall*) &callback), ud);
                                                            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Which is functionally stuffing a `void*` into and `int` (though hidden behind some typedefs).

For the current, specific use cases of Tock, I believe our suite of hardware platforms and pointer models is sufficiently limited that this will work. I'm less sure whether all this means we should, or shouldn't, merge this change [which, really isn't changing any functionality; we're already doing the bad thing IIUC].

bradjc:
Do we need this now? Can you make these changes at least consistent with the guide?

lschuermann:
I don't know whether we need to change it, but if we do, I can make it consistent with the guide.

bradjc (changes requested):
Namespacing and fn typdef format.

lschuermann:
Let's postpone this change until after the tutorial.

ppannuto:
@lschuermann Is this something you still want to merge?

lschuermann:
> @lschuermann Is this something you still want to merge?

Yes and no. I think this is an improvement worth pursuing if we were to stick with the old IPC interface, but with a new IPC implementation around the corner, we might as well avoid an interface-breaking change with little tangible benefit. Let's hope we'll get a new, improved, message-passing IPC soon and then this will be irrelevant. :smile:

PR #422, opened 2024-05-10
libtock-sync: add missing static

bradjc:
These were missed in the original PR, and it seems that newer compilers don't care but older versions (circa 2020) do.


brghena (approved):

bradjc:
Report from Tyler is this fixed the issue.

PR #425, opened 2024-05-12
IEEE802154 libtock-sync sending timeout

tyler-potyondy:
This updates the libtock-sync send raw to timeout after 100ms. This prevents a failing or delayed 15.4 transmission from blocking the application. This PR also adds separate objects for maintaining state across raw/standard sending.

This fixes a last minute bug found in the Thread tutorial, but I am not convinced this is necessarily the correct approach long term.

### DIFF
--- a/libtock-sync/net/ieee802154.c
+++ b/libtock-sync/net/ieee802154.c
@@ -2,6 +2,8 @@
 
 #include "ieee802154.h"
 
+
+#include <stdio.h>
 struct ieee802154_receive_data {
   bool fired;
   int pan;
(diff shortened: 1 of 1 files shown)


lschuermann:
I think this is fine for now. The separation into raw / non-raw makes sense to me.

The timeout in the callback is more tricky. What can happen here is that the timeout fires, but a callback for the send operation will still be scheduled. A subsequent call to send might then receive this stale callback and assume that its packet has been completely transmitted, and proceed to unshare buffers, etc.

We should revisit this and make sure that we, e.g., associate callbacks with a particular send operation instead. Being able to have a timeout on a send call seems important.

lschuermann (approved):
Approving modulo the comment above.

PR #428, opened 2024-05-13
library: lvgl: Build library instead of shipping pre-built binaries

bradjc:
Same as with Lua. Build the library on demand. Also, use the Makefile.setup infrastructure rather than checking for the submodule in every app.


lvgl/Makefile.setup
```diff
@@ -0,0 +1,7 @@
+TOCK_USERLAND_BASE_DIR ?= ..
+
+$(TOCK_USERLAND_BASE_DIR)/lvgl/lvgl/lvgl.h:
+	git submodule init $(TOCK_USERLAND_BASE_DIR)/lvgl/lvgl
+	git submodule update $(TOCK_USERLAND_BASE_DIR)/lvgl/lvgl
+
+all: | $(TOCK_USERLAND_BASE_DIR)/lvgl/lvgl/lvgl.h
```
ppannuto, line 7: This won't actually update if the submodule pointer is updated.

I think it might make sense to do something more like

```suggestion
.PHONY: all
all:
	git submodule init $(TOCK_USERLAND_BASE_DIR)/lvgl/lvgl
	git submodule update $(TOCK_USERLAND_BASE_DIR)/lvgl/lvgl
```

(a quick test on my machine confirms that both `init` and `update` generate no output and return success if they are no-op's / already done)

ppannuto (changes requested):

lvgl/Makefile.setup
bradjc: But the deeper problem is this file won't run at all once you have the `.a` files, so this only matters if you have a fresh checkout or if a new library is added.

alevy (approved):

PR #434, opened 2024-05-28
alarm: create `ms_to_ticks` helper function

Samir-Rashid:
https://github.com/tock/libtock-c/pull/395 has [stalled](https://github.com/tock/libtock-c/pull/395#issuecomment-2127624790) as the PR has grown beyond original scope. This PR splits out the modifications to arithmetic logic and should get merged before #395.

---
This mega-size comment was originally in the code explaining the logic of the code change.
```c
static uint32_t ms_to_ticks(uint32_t ms) {
  uint32_t frequency;
  libtock_alarm_command_get_frequency(&frequency);

  // Note the following code is mathematically equivalent to the following
  // more readable conversion.
  // ```
  //   uint32_t seconds         = ms / 1000;
  //   uint32_t leftover_millis = ms % 1000;
  //   uint32_t millihertz      = frequency / 1000; // ticks per millisecond
  //   return (seconds * frequency) + (leftover_millis * millihertz);
  // ```
  // We can check the logical correctness of this conversion by doing dimensional
  // analysis of the units.
  // For the first half, `seconds` (sec) * `frequency` (ticks per second) = a quantity in ticks.
  // And for the second half, `leftover_millis` (milliseconds) * `millihertz` (ticks per millisecond)
  // = a quantity in ticks.
  // Thus, we know the calculation is logically correct as the units cancel out to return ticks.
  //
  // Although these are logically equivalent, this human readable implementation is
  // faulty. I will prove this by (1.) showing an upper bound on the error of this conversion,
  // and then (2.) explaining why this commented human readable implementation is
  // equivalent to the true implementation.
  //
  // Part 1:
  // The splitting of the input milliseconds has no error.
  uint32_t seconds         = ms / 1000;
  uint32_t leftover_millis = ms % 1000;
  // In other words, `seconds` + `leftover_millis` is always == `ms`.

  // The first half of the output converts the seconds component to ticks with no error.
  // This multiplication has no error because the board frequency is given in Hertz (ticks per second).
  uint64_t ticks = (uint64_t) seconds * frequency;

  // Now that we have converted the full seconds part of `ms` to ticks, we must
  // convert the `leftover_millis`.
  // The only error introduced is by the division step. Division loses
  // all fractional components, so you lose 3 decimal significant figures by doing the division.
  // This does not matter as the remainder lost to integer division is a fraction of a
  // millisecond. It is especially important that this conversion is lower than the
  // true value of the conversion. In summary, the error in `ms_to_ticks` is from losing
  // significant figures, but this loss is less than 1 millisecond of error.
  //
  // Note that the division happens after the multiplication. Let us take a look at a
  // concrete example of a worst case conversion.
  // For a board with an oscillator frequency of 32,768Hz (NRF52840dk), observe the
  // error introduced by arithmetic in the "human readable" implementation of this
  // same line of the code:
  //   uint32_t millihertz      = frequency / 1000; // ticks per millisecond
  //   ticks += leftover_millis * millihertz;
  //
  //   // The worst case is when the remainder `leftover_millis` is `frequency` - 1
  //   ticks += 32767 * millihertz;
  // But what is the value of millihertz here? It is `32768 / 1000` = `32`! This is
  // 2.4% error, has only 2 significant figures, and, worst of all, the division causes the
  // denominator to be smaller than it should be. Dividing by a smaller number causes the
  // entire value of `ticks` to be an overestimate. This is a critical correctness issue.
  // Let's trace an example execution of calculating overflows in `timer_in`. The code uses
  // this line to calculate how many overflows are needed.
  //     const uint32_t max_ticks_in_ms = ticks_to_ms(MAX_TICKS);
  // If `ticks_to_ms` is an overestimate, then a timer set for slightly longer than the
  // true value of `MAX_VALUE` in ms will cause the check for `length of timer` <
  // `max_ticks_in_ms` to be true. This incorrect check will set an overflowed timer and fire
  // too early.
  uint32_t milliseconds_per_second = 1000;
  ticks += ((uint64_t) leftover_millis * frequency) / milliseconds_per_second;

  // Part 2:
  // The human readable and actual implementation are logically equivalent.
  // The only difference is the location of the division by 1000.
  //   leftover_millis * millihertz = leftover_millis * (frequency / 1000)
  //   = leftover_millis * frequency / 1000 = (leftover_millis * frequency) / 1000
  // Thus, by a series of equivalences, these are doing the same calculation.

  return ticks;
}
```

### DIFF
--- a/libtock/services/alarm.c
+++ b/libtock/services/alarm.c
@@ -9,8 +9,8 @@ static int within_range(uint32_t a, uint32_t b, uint32_t c) {
 
 /** \brief Convert milliseconds to clock ticks
  *
- * WARNING: This function will silently overflow if the output
- * number of ticks is larger than `UINT32_MAX`.
+ * WARNING: This function will assert if the output
+ * number of ticks overflows `UINT32_MAX`.
  *
  * This conversion is accurate to within 1 millisecond of a true
  * fractional conversion.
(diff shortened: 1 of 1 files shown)


libtock/services/alarm.c
```diff
@@ -7,6 +7,95 @@ static int within_range(uint32_t a, uint32_t b, uint32_t c) {
...
+  libtock_alarm_command_get_frequency(&frequency);
+
+  // Note the following code is mathematically equivalent to the following
+  // more readable conversion.
+  // ```
+  //   uint32_t seconds         = ms / 1000;
+  //   uint32_t leftover_millis = ms % 1000;
+  //   uint32_t millihertz      = frequency / 1000; // ticks per millisecond
+  //   return (seconds * frequency) + (leftover_millis * millihertz);
+  // ```
```
tyler-potyondy, line 33: I think that with the exception of this comment, the other lengthy comments/proof in this function should be moved to the PR description.

You can leave a pointer to this PR for those interested in the more in depth discussion you provide here. Having the lengthy description is helpful for future work, but I think more verbose than necessary in the code base. Leaving this for posterity in the PR that is pointed to from the comment seems to offer a good middle ground.

libtock/services/alarm.c
```diff
@@ -7,6 +7,95 @@ static int within_range(uint32_t a, uint32_t b, uint32_t c) {
   return (b - a) < (b - c);
 }
 
+/** \brief Convert milliseconds to clock ticks
+ *
+ * WARNING: This function will silently overflow if the output
+ * number of ticks is larger than `UINT32_MAX`.
```
tyler-potyondy, line 13: Perhaps it would be better to pass a `uint32_t` reference that this function updates and instead returns a returncode (returning an error on an overflow)? This is not an optimal fix, but is a step in the right direction. Silent overflows have made debugging other parts of the libtock alarm infrastructure challenging and something I think we should avoid when possible.

tyler-potyondy (changes requested):

libtock/services/alarm.c
tyler-potyondy: One solution is to check if the calculated ticks value is greater than `UINT32_MAX` prior to casting from `uint64_t` to `uint32_t` and return the relevant `returncode_t` accordingly.

libtock/services/alarm.c
Samir-Rashid: Good idea. I've changed it, and I am pasting the content here temporarily and will move it to the PR description later.

---

```c
static uint32_t ms_to_ticks(uint32_t ms) {
  uint32_t frequency;
  libtock_alarm_command_get_frequency(&frequency);

  // Note the following code is mathematically equivalent to the following
  // more readable conversion.
  // ```
  //   uint32_t seconds         = ms / 1000;
  //   uint32_t leftover_millis = ms % 1000;
  //   uint32_t millihertz      = frequency / 1000; // ticks per millisecond
  //   return (seconds * frequency) + (leftover_millis * millihertz);
  // ```
  // We can check the logical correctness of this conversion by doing dimensional
  // analysis of the units.
  // For the first half, `seconds` (sec) * `frequency` (ticks per second) = a quantity in ticks.
  // And for the second half, `leftover_millis` (milliseconds) * `millihertz` (ticks per millisecond)
  // = a quantity in ticks.
  // Thus, we know the calculation is logically correct as the units cancel out to return ticks.
  //
  // Although these are logically equivalent, this human readable implementation is
  // faulty. I will prove this by (1.) showing an upper bound on the error of this conversion,
  // and then (2.) explaining why this commented human readable implementation is
  // equivalent to the true implementation.
  //
  // Part 1:
  // The splitting of the input milliseconds has no error.
  uint32_t seconds         = ms / 1000;
  uint32_t leftover_millis = ms % 1000;
  // In other words, `seconds` + `leftover_millis` is always == `ms`.

  // The first half of the output converts the seconds component to ticks with no error.
  // This multiplication has no error because the board frequency is given in Hertz (ticks per second).
  uint64_t ticks = (uint64_t) seconds * frequency;

  // Now that we have converted the full seconds part of `ms` to ticks, we must
  // convert the `leftover_millis`.
  // The only error introduced is by the division step. Division loses
  // all fractional components, so you lose 3 decimal significant figures by doing the division.
  // This does not matter as the remainder lost to integer division is a fraction of a
  // millisecond. It is especially important that this conversion is lower than the
  // true value of the conversion. In summary, the error in `ms_to_ticks` is from losing
  // significant figures, but this loss is less than 1 millisecond of error.
  //
  // Note that the division happens after the multiplication. Let us take a look at a
  // concrete example of a worst case conversion.
  // For a board with an oscillator frequency of 32,768Hz (NRF52840dk), observe the
  // error introduced by arithmetic in the "human readable" implementation of this
  // same line of the code:
  //   uint32_t millihertz      = frequency / 1000; // ticks per millisecond
  //   ticks += leftover_millis * millihertz;
  //
  //   // The worst case is when the remainder `leftover_millis` is `frequency` - 1
  //   ticks += 32767 * millihertz;
  // But what is the value of millihertz here? It is `32768 / 1000` = `32`! This is
  // 2.4% error, has only 2 significant figures, and, worst of all, the division causes the
  // denominator to be smaller than it should be. Dividing by a smaller number causes the
  // entire value of `ticks` to be an overestimate. This is a critical correctness issue.
  // Let's trace an example execution of calculating overflows in `timer_in`. The code uses
  // this line to calculate how many overflows are needed.
  //     const uint32_t max_ticks_in_ms = ticks_to_ms(MAX_TICKS);
  // If `ticks_to_ms` is an overestimate, then a timer set for slightly longer than the
  // true value of `MAX_VALUE` in ms will cause the check for `length of timer` <
  // `max_ticks_in_ms` to be true. This incorrect check will set an overflowed timer and fire
  // too early.
  uint32_t milliseconds_per_second = 1000;
  ticks += ((uint64_t) leftover_millis * frequency) / milliseconds_per_second;

  // Part 2:
  // The human readable and actual implementation are logically equivalent.
  // The only difference is the location of the division by 1000.
  //   leftover_millis * millihertz = leftover_millis * (frequency / 1000)
  //   = leftover_millis * frequency / 1000 = (leftover_millis * frequency) / 1000
  // Thus, by a series of equivalences, these are doing the same calculation.

  return ticks;
}
```

libtock/services/alarm.c
Samir-Rashid: I'm not fully sold on the need for returncodes for internal helper functions. What would the code do if it does detect an overflow? This is certainly a step in the right direction, so I have added an assert.

tyler-potyondy: I think having the assertion here is enough. I think we can now remove the comment warning of silently overflowing if ticks is greater than 2**32.

tyler-potyondy (changes requested):

tyler-potyondy:
Once you update the comment, rebase on master, and format to pass CI (run `make` in libtock-c directory) this should be good to go!

libtock/services/alarm.c
bradjc: Who is supposed to do the check to avoid the assert? The application?

tyler-potyondy: If I understand correctly, this is only a helper method and is not exposed to applications. The user of this method will be `timer_in` which with #395 will add the logic needed to avoid the assert.

For this internal method, it seems reasonable to leave this assert since we should be avoiding this case.

bradjc: Got it

bradjc (dismissed):

bradjc (approved):

tyler-potyondy (approved):

ppannuto (approved):

libtock/services/alarm.c
```diff
@@ -7,6 +7,35 @@ static int within_range(uint32_t a, uint32_t b, uint32_t c) {
...
+  // View the justification here https://github.com/tock/libtock-c/pull/434
+  uint32_t frequency;
+  libtock_alarm_command_get_frequency(&frequency);
+
+  uint32_t seconds                 = ms / 1000;
+  uint32_t leftover_millis         = ms % 1000;
+  uint32_t milliseconds_per_second = 1000;
+
+  uint64_t ticks = (uint64_t) seconds * frequency;
+  ticks += ((uint64_t) leftover_millis * frequency) / milliseconds_per_second;
```
lschuermann, line 33: I'm just catching up to speed, but given that we're in 64-bit arithmetic now, why is this still dealing with seconds and fractional seconds individually? Wouldn't this suffice now?

```
    ms = (ticks * milliseconds_per_second) / freq
    ticks = (ms * freq) / milliseconds_per_second
```

PR #438, opened 2024-06-05
15.4: Add new raw 15.4 commands; try to turn on radio

bradjc:
This supports adding a raw 15.4 interface for userspace. The changes are:

- Add radio on/off commands. We now have full control of the radio so we can do these.
- Add trying to turn the radio on when checking if the radio is up. Before we could assume the kernel was managing the stack. To not have two divergent stacks in userspace, this speculatively tries to turn on the radio. On the kernel-managed stack this will fail, but we don't care. On the raw stack this may be necessary, otherwise the radio isn't on.
- Use 64-bit interfaces rather than sharing a cfg buffer to set/get 64 bit addresses.

### DIFF
--- a/libtock/net/ieee802154.c
+++ b/libtock/net/ieee802154.c
@@ -17,11 +17,6 @@ int libtock_ieee802154_down(void) {
 }
 
 returncode_t libtock_ieee802154_is_up(bool* status) {
-  // For the raw implementation of the 15.4 driver we need to turn on the radio.
-  // If the kernel is managing the 15.4 stack, we likely do not need to do this.
-  // In that case, this will error, but we ignore the error.
-  libtock_ieee802154_command_radio_on();
-
   returncode_t ret = libtock_ieee802154_command_status();
   if (ret == RETURNCODE_EOFF) {
     *status = false;
--- a/libtock/net/ieee802154.c
+++ b/libtock/net/ieee802154.c
@@ -17,6 +17,11 @@ int libtock_ieee802154_down(void) {
 }
 
 returncode_t libtock_ieee802154_is_up(bool* status) {
+  // For the raw implementation of the 15.4 driver we need to turn on the radio.
+  // If the kernel is managing the 15.4 stack, we likely do not need to do this.
+  // In that case, this will error, but we ignore the error.
+  libtock_ieee802154_command_radio_on();
+
   returncode_t ret = libtock_ieee802154_command_status();
   if (ret == RETURNCODE_EOFF) {
     *status = false;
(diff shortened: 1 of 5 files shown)


libtock/net/ieee802154.c
```diff
@@ -17,6 +17,11 @@ int libtock_ieee802154_down(void) {
 }
 
 returncode_t libtock_ieee802154_is_up(bool* status) {
+  // For the raw implementation of the 15.4 driver we need to turn on the radio.
+  // If the kernel is managing the 15.4 stack, we likely do not need to do this.
+  // In that case, this will error, but we ignore the error.
+  libtock_ieee802154_command_radio_on();
+
```
tyler-potyondy, line 24: It seems to me that it may make more sense to not turn the radio on here. This function, as I see it, is just a status getter of the radio. The app can then turn the radio on using the exposed `libtock_ieee802154_radio_on` method if the app desires the radio to be turned on.

tyler-potyondy:
Only comment/feedback is above regarding the is_up behavior. Otherwise looks good to go.

libtock/net/ieee802154.c
bradjc: Yeah I put this in the wrong place. I meant to put it in `libtocksync_ieee802154_up()`.

bradjc: Moved.

tyler-potyondy (approved):

libtock/net/syscalls/ieee802154_syscalls.c
```diff
@@ -162,3 +162,25 @@ returncode_t libtock_ieee802154_command_send_raw(void) {
...
+
+returncode_t libtock_ieee802154_command_radio_on(void) {
+  syscall_return_t cval = command(DRIVER_NUM_IEEE802154, COMMAND_RADIO_ON, 0, 0);
+  return tock_command_return_novalue_to_returncode(cval);
+}
+
+returncode_t libtock_ieee802154_command_radio_off(void) {
+  syscall_return_t cval = command(DRIVER_NUM_IEEE802154, COMMAND_RADIO_OFF, 0, 0);
+  return tock_command_return_novalue_to_returncode(cval);
+}
```
tyler-potyondy, line 186: Thinking more about this, I do not think we want this to be `tock_command_return_novalue_to_returncode` since we are using the same 15.4 driver for both raw/standard 15.4 stack in the kernel. Since these syscalls are not implemented in the kernel, this should return the returncode here since the standard 15.4 stack will return an error here and this is useful information.

PR #440, opened 2024-06-06
U8G2: Add GIFs to readmes

bradjc:
This makes it a little easier to know what the screen should be doing if things are working.


ppannuto (approved):
This is a really nice touch of polish. Screens are fun :)

PR #442, opened 2024-06-17
Support the latest newlib, libc++ and GCC

alistair23:
@bradjc you will need to update the pre-built versions to support GCC 14

### DIFF
--- a/Configuration.mk
+++ b/Configuration.mk
@@ -253,8 +253,10 @@ else ifeq ($(CC_rv32_version_major),12)
   NEWLIB_VERSION_rv32 := 4.3.0.20230120
 else ifeq ($(CC_rv32_version_major),13)
   NEWLIB_VERSION_rv32 := 4.3.0.20230120
+else ifeq ($(CC_rv32_version_major),14)
+  NEWLIB_VERSION_rv32 := 4.4.0.20231231
 else
-  NEWLIB_VERSION_rv32 := 4.3.0.20230120
+  NEWLIB_VERSION_rv32 := 4.4.0.20231231
 endif
 NEWLIB_VERSION_rv32i    := $(NEWLIB_VERSION_rv32)
 NEWLIB_VERSION_rv32imc  := $(NEWLIB_VERSION_rv32)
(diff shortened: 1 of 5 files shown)


newlib/newlib-4.4.0.20231231.patch
lschuermann, line 1: What's up with the empty patch file here?

lschuermann (commented):
Thanks for updating this! @bradjc do you want to build the toolchains or I should I try and get my hands dirty on Docker?

lschuermann:
Skimming over the scripts, it seems like you actually want to copy the files in `docker/`, instead of creating a `.patch` file.

newlib/newlib-4.4.0.20231231.patch
alistair23: The build system requires a patch file to exist. So this avoids failures. I figured it can just be removed when support for the old newlib versions is removed

Configuration.mk
```diff
@@ -393,8 +397,10 @@ else ifeq ($(CC_cortex-m_version_major),12)
   NEWLIB_VERSION_cortex-m := 4.3.0.20230120
 else ifeq ($(CC_cortex-m_version_major),13)
   NEWLIB_VERSION_cortex-m := 4.3.0.20230120
+else ifeq ($(CC_cortex-m_version_major),13)
```
bradjc, line 400:
```suggestion
else ifeq ($(CC_cortex-m_version_major),14)
```

bradjc:
Yes this needs docker files.

Also, is this worth our time?

alistair23:
Docker files are added.

The current outdated libc++ and newlib fail to build with GCC 14, so we do need an update. This generally will probably need to be updated with each newlib and GCC release

libc++/docker/docker-libc++-14.1.0/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+RUN apt install -y gcc-arm-none-eabi gcc-riscv64-unknown-elf
+
+# Install needed tools
+RUN apt install -y git build-essential wget rsync zip texinfo
+
+# Install needed library
+RUN apt install -y libmpc-dev file
+
+# Clone the libtock-c source so we can use the build scripts
+RUN git clone https://github.com/alistair23/libtock-c
```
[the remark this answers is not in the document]
alistair23: Then it won't build. It's a chicken and egg problem, you need https://github.com/tock/libtock-c/pull/442/commits/1607c6bcb299477eb69968f53c2e478ed240b97a merged first

alistair23:
Any progress on this?

alistair23:
Gentle ping on this, we need to keep on top of updating these. GCC 14 is rolling out more and more

ppannuto:
#470 is my best solution to this — hopefully that is sufficient as-is. If not, we can make updates there?

ppannuto:
Closing in favor of #470.

PR #444, opened 2024-06-19
shell.nix: update deps, add jlink

Samir-Rashid:
## Overview

With the 24.05 nixpkgs release, we should be good to add jlink back. I have added necessary deps uncrustify and unzip. Decompressing newlib fails (silently) and tries to reinstall every time if `unzip` is unavailable.

### Testing Strategy

Built all examples successfully.

CC: @lschuermann

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -47,23 +47,15 @@ in
       python3Full
       tockloader
       pkgsCross.riscv32-embedded.buildPackages.gcc
-    ];
+      uncrustify
+      unzip
+    ] ++ (lib.optionals withUnfreePkgs [
+      segger-jlink
+      tockloader.nrf-command-line-tools
+    ]);
 
-    # Unfortunately, `segger-jlink` has been removed from Nixpkgs due to its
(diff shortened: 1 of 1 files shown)


shell.nix
```diff
@@ -47,23 +47,15 @@ in
...
-    # shellHook = ''
-    #   # TODO: This should be patched into the rpath of the respective libraries!
-    #   export LD_LIBRARY_PATH=${pkgs.libusb}/lib:${pkgs.segger-jlink}/lib:$LD_LIBRARY_PATH
-    # '';
-    #
-    # [1]: https://github.com/NixOS/nixpkgs/pull/255185
+      uncrustify
+      unzip
+    ] ++ (lib.optionals withUnfreePkgs [
+      segger-jlink
```
alevy, line 53: Ugh... it still requires `export NIXPKGS_ALLOW_INSECURE=1` because JLink relies on some ancient version of Qt. Fine...

alevy (approved):

PR #447, opened 2024-06-20
OpenThread Fixes and Updates

tyler-potyondy:
This PR refactors the libopenthread platform to embrace an async design. The prior implementation included some potentially sync/yielding code within callbacks. Now, these callbacks (radio, alarm, etc) set flags that are then polled during the "libopenthread work" of the the openthread main loop. 

Additionally, the 15.4 raw driver merged recently and corresponding changes caused OpenThread to no longer transmit/receive. This was due to OpenThread requesting to sleep the radio (i.e. turn it off), but failed to power the radio on again before requesting to tx/rx. This is now fixed in this PR.

The openthread_hello and tutorial openthread apps were updated respectively.


bradjc:
Looks good.

alevy:
+1 to @bradjc's comment and overall approval otherwise.

alevy (approved):

PR #448, opened 2024-06-21
U8G2: Add increment/decrement example app

bradjc:
This could be useful for choosing a temperature, say.


alevy (approved):

alevy:
Can't test with no screen, but looks good to me

ppannuto (approved):

PR #449, opened 2024-06-21
U8G2: Add tool to create .gif files from U8G2 apps

bradjc:
This is a simple tool that compiles u8g2 apps for the native machine and generates image files (rather than writing to a display) when the app runs. Those images can be then be merged into a .gif.

The difficulty here is that it is hard to emulate a tock app. So, this only really works for very simple apps (say that only use a timer). But it is nice to have images to show what an app does.


alevy (commented):
This seems a bit overcomplicated and and specific to be in this repo to me.

bradjc:
Specific to libtock? This needs to implement a libtock shim to work with libtock apps.

alevy (approved):
Fine :)

PR #452, opened 2024-06-28
use default shasum for os x if present

### LINKED ISSUE #450: lib: newlib and libc++ scripts assume sha256sum
We use SHA256 to verify the download, but we assume the user has a `sha256sum` binary. This isn't the case on mac by default. We should handle this case somehow.

ppannuto:
Fixes #450.

### DIFF
--- a/lib/fetch-libc++.sh
+++ b/lib/fetch-libc++.sh
@@ -19,6 +19,12 @@ MIRRORS=(\
   "https://alpha.mirror.svc.schuermann.io/files/tock"\
 )
 
+if test -x /usr/bin/shasum; then
+  CHECK_SHA_CMD="shasum -a 256 -c"
+else
+  CHECK_SHA_CMD="sha256sum -c"
+fi
+
 let FOUND=0
 
 # Try from each mirror until we successfully download a .zip file.
@@ -26,7 +32,8 @@ for MIRROR in ${MIRRORS[@]}; do
   URL=$MIRROR/$ZIP_FILE
   echo "Fetching libc++ from ${MIRROR}..."
   echo "  Fetching ${URL}..."
-  wget -O $ZIP_FILE  "$URL" && (echo "$GCC_SHA $ZIP_FILE" | sha256sum -c)
+  # Note: There must be two space characters for `shasum` (sha256sum doesn't care)
+  wget -O $ZIP_FILE  "$URL" && (echo "$GCC_SHA  $ZIP_FILE" | $CHECK_SHA_CMD)
   if [ $? -ne 0 ]; then
     echo "  WARNING: Fetching libc++ from mirror $MIRROR failed!" >&2
   else
--- a/lib/fetch-newlib.sh
+++ b/lib/fetch-newlib.sh
@@ -17,6 +17,12 @@ MIRRORS=(\
   "https://alpha.mirror.svc.schuermann.io/files/tock"\
 )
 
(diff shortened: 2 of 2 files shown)


lib/fetch-libc++.sh
```diff
@@ -19,14 +19,21 @@ MIRRORS=(\
...
+
 let FOUND=0
 
 # Try from each mirror until we successfully download a .zip file.
 for MIRROR in ${MIRRORS[@]}; do
   URL=$MIRROR/$ZIP_FILE
   echo "Fetching libc++ from ${MIRROR}..."
   echo "  Fetching ${URL}..."
-  wget -O $ZIP_FILE  "$URL" && (echo "$GCC_SHA $ZIP_FILE" | sha256sum -c)
+  # Note: There must be two space characters for `shasum` (sha256sum doesn't care)
```
alevy, line 35: Two space characters where? what is this referring to?

lib/fetch-libc++.sh
```diff
@@ -19,14 +19,21 @@ MIRRORS=(\
   "https://alpha.mirror.svc.schuermann.io/files/tock"\
 )
 
+if test -x /usr/bin/shasum; then
```
alevy, line 22: This _at least_ needs a comment. Why this particular path? Do we always prefer `shasum` if it is present? Or only if `sha256sum` _isn't_ present?

Is this the `shasum` that comes with Perl? If so, Perl is a pre-requisite for most Linux distributions in practice, does it make sense to just use `shasum`?

lib/fetch-newlib.sh
```diff
@@ -17,14 +17,21 @@ MIRRORS=(\
   "https://alpha.mirror.svc.schuermann.io/files/tock"\
 )
 
+if test -x /usr/bin/shasum; then
```
alevy, line 20: That this is duplicated is a bit of a red flag.

alevy (changes requested):
I think this is hacks on hacks that is going to be impossible to reliably test and maintain.

lib/fetch-libc++.sh
bradjc: Between the correct sha and the file name:

```
❯ echo "1e6f33cd5f44acefebc1ef677db4be33829a5228  README.md" | shasum -a 256 -c
README.md: OK
❯ echo "1e6f33cd5f44acefebc1ef677db4be33829a5228 README.md" | shasum -a 256 -c
shasum: standard input: no properly formatted SHA checksum lines found
```

lib/fetch-libc++.sh
bradjc: This is what happened in the kernel: https://github.com/tock/tock/pull/1669. Should we have done something differently? I think we just want this to work on as many systems as possible.

lschuermann (approved):
Solution looks good to me, and this does solve a real problem.

It's disappointing that this is still an issue, and (apart from all the other reasons why this is bad) we probably don't want to require a host C/Rust toolchain to build our own sha256sum tool in this repo.

ppannuto:
> I think this is hacks on hacks that is going to be impossible to reliably test and maintain.

For better or worse, it's just propagating over the existing approach (hack? :) ) from the kernel repo, which has been unchanged for years now. Think this hits least-bad pragmatic approach for the current ecosystem of developer machines writ large.

alevy (approved):

PR #453, opened 2024-06-28
Fix nix shell: withUnfreePackages = false

alevy:
#444 attempted to re-add segger-jlink only under the condition that `withUnfreePkgs` is true, but actually includes it implicitly in `LD_LIBRARY_PATH`.

This PR wraps the addition of that package in the same conditional.

Tested by entering the nix-shell with both `withUnfreePkgs` enabled and without.


Samir-Rashid (approved):
Works with and without unfree

lschuermann (approved):

PR #455, opened 2024-07-07
libtock-sync: alarm: Support empty delays

alistair23:
If libtocksync_alarm_delay_ms() is provided a delay of 0ms we currenty schedule an alarm which is a high overhead for a empty delay.

Instead let's just call a yield_no_wait() and return success if we don't actually need to delay.

### DIFF
--- a/libtock-sync/services/alarm.c
+++ b/libtock-sync/services/alarm.c
@@ -17,6 +17,11 @@ int libtocksync_alarm_delay_ms(uint32_t ms) {
   libtock_alarm_t alarm;
   int rc;
 
+  if (ms == 0) {
+    yield_no_wait();
+    return RETURNCODE_SUCCESS;
+  }
+
   if ((rc = libtock_alarm_in_ms(ms, delay_cb, NULL, &alarm)) != RETURNCODE_SUCCESS) {
     return rc;
   }


bradjc:
Why call `yield_no_wait()` at all? Why not just return?

lschuermann:
> Why call `yield_no_wait()` at all? Why not just return?

I'm going to guess that users may expect a delay to always cause upcalls to run. Which I think is a reasonable expectation.

However, with that premise, I'm not sure I buy this optimization, when it still runs a full system call. Sure, we get rid of a bit of overhead, but how common is this edge case?

bradjc:
We need to reckon with what libtock-sync guarantees, and I think the expectation we want to move towards (and why we added yield-nowait) is that unrelated events _don't_ run. I know that delay_ms doesn't promise that today, but I think the ideal is that it does block everything.

alistair23:
> I'm going to guess that users may expect a delay to always cause upcalls to run. Which I think is a reasonable expectation.

Exactly!

Basically FreeRTOS tasks (and lots of stuff is built around FreeRTOS) use `vTaskDelay(0)`. So this is adapting to support that type of usecase

alistair23:
`libtocksync_alarm_delay_ms()` today already allows anything else to run though, so this isn't really any different. We are just not setting an alarm

brghena:
So, to be clear, a delay of 0 ms does _work_ today with the code as it exists before this PR, right?

If so, then I don't see a good reason for the optimization here. Applications can call `yield_no_wait()` directly, and don't have to `libtocksync_alarm_delay_ms(0)`.

I can see a scenario where some higher-level library is deciding on delay times, and might incidentally decide to delay 0 ms. But in my mind, that means this optimization should go in that higher-level library. But maybe there's an important use-case I'm missing?

I personally think the argument that matters here is whether we should have the optimization or not. If yes, then per today's semantics a `yield_no_wait()` makes sense. And a future change to make `delay_ms()` actually delay and not yield for anything could fixup this optimization too.

alistair23:
> So, to be clear, a delay of 0 ms does _work_ today with the code as it exists before this PR, right?

It doesn't work for me, but I suspect that's a bug in Tock.

> If so, then I don't see a good reason for the optimization here. Applications can call `yield_no_wait()` directly, and don't have to `libtocksync_alarm_delay_ms(0)`.

Ha! The high level application said the exact opposite: https://github.com/jgromes/RadioLib/pull/1149#issuecomment-2211799640

One issue that other applications might not have a `yield_no_wait()` HAL to call, but they do have `delay(0)`

alistair23:
I'm also happy to drop the `yield_no_wait()` from this PR, but that feels a bit sneaky to me as the caller probably expects a yield to be called

lschuermann:
> It doesn't work for me, but I suspect that's a bug in Tock.

I agree, this seems to be the real fix then.

ppannuto:
I'm curious whether this is really a bug in Tock or a bug somewhere in the libtock-c virtual alarm stack?

In either case, this seems like papering over the real problem, e.g., putting this check here won't fix the case where someone chooses to call `libtock_alarm_in_ms(0, ...)` for a user of the async libtock interface (which is all that the `delay_ms(0` translates too under the hood.

I agree there's a bug somewhere in the timer stack, but this doesn't look like the right place to fix it (at minimum the `==0` check needs to move to the lower-layer async impl).

ppannuto (changes requested):
This isn't the right layer of the timer stack for this check

PR #432, opened 2024-05-16
lora: Support RISC-V build, update to latest libtock-c and move to examples

alistair23:

PR #456, opened 2024-07-14
lora: Initial commit of LoRaWAN example

alistair23:
This builds on top of https://github.com/tock/libtock-c/pull/432 and adds a LoRaWAN example where a Tock system can send data to [TTN](https://www.thethingsnetwork.org/).

### DIFF
--- a/examples/build_all.sh
+++ b/examples/build_all.sh
@@ -2,6 +2,8 @@
 
 NUM_JOBS=$(nproc 2>/dev/null || sysctl -n hw.ncpu 2>/dev/null || 4)
 
+export TOCK_BUILDALL=1
+
 set -e
 set -u
 set -o pipefail
--- a/examples/build_all.sh
+++ b/examples/build_all.sh
@@ -2,6 +2,27 @@
 
 NUM_JOBS=$(nproc 2>/dev/null || sysctl -n hw.ncpu 2>/dev/null || 4)
 
+# Sentinel flag to indicate to apps that a `build all` is running.
+#
+# Generally, apps should not be sensitive to this flag. Build all is
+# usually only invoked by a user when they're looking for local, CI-like
+# behavior. However, a local repo checkout is not quite the same as a
+# pristine CI environment, and certain test apps may need slightly
+# different behavior for a local "build all" versus CI.
+#
+# For example: Some apps, e.g. LoRa-based ones, require radio
+# configuration that end users *must* set themselves [e.g. to select
+# operating region]. For users of these apps, we *do not* want them to
+# build successfully 'out of the box' without explicit user
+# intervention, but we do still want them to be compile-tested in CI. An
+# app may ship with a `radioConfig_example.h` but include (and
+# `.gitignore`) `radioConfig.h`.  CI can unconditionally swap in the
+# `_example` version, but a local checkout should use a real config file
+# it is available. At the same time, apps which are not used in a local
+# checkout should not throw compile errors for a local "build all", and
+# should thus swap in the `_example` if a real config is not present. An
+# app build which is sensitive to `TOCK_BUILDALL` can capture this
+# subtlety.
 export TOCK_BUILDALL=1
 
 set -e
--- a/examples/lora/sensor-lorawan/CayenneLPP.cc
+++ b/examples/lora/sensor-lorawan/CayenneLPP.cc
@@ -1,0 +1,217 @@
+/* SPDX-License-Identifier: GPL-3.0 */
+/*
+ * The Cayenne Low Power Payload (LPP) provides a convenient and
+ * easy way to send data over LPWAN networks such as LoRaWAN. The Cayenne LPP
+ * is compliant with the payload size restriction, which can be lowered down to
+ * 11 bytes, and allows the device to send multiple sensor data at one time.
+ *
+ * CayenneLPP is also support by TTN:
+ *     https://www.thethingsindustries.com/docs/integrations/payload-formatters/cayenne/
+ *
+ * This code is direct from: https://github.com/myDevicesIoT/CayenneLPP
+ *
+ * GPL-3.0 license, see `LICENSE.GPL`
+ * See https://github.com/myDevicesIoT/CayenneLPP for license details
+ */
+
+#include "CayenneLPP.h"
+
+CayenneLPP::CayenneLPP(uint8_t size) : maxsize(size) {
+  buffer = (uint8_t*) malloc(size);
+  cursor = 0;
+}
+
+CayenneLPP::~CayenneLPP(void) {
+  free(buffer);
+}
+
+void CayenneLPP::reset(void) {
+  cursor = 0;
+}
+
+uint8_t CayenneLPP::getSize(void) {
+  return cursor;
+}
+
+uint8_t* CayenneLPP::getBuffer(void) {
+  return buffer;
+}
+
+uint8_t CayenneLPP::copy(uint8_t* dst) {
+  memcpy(dst, buffer, cursor);
+  return cursor;
+}
+
+uint8_t CayenneLPP::addDigitalInput(uint8_t channel, uint8_t value) {
+  if ((cursor + LPP_DIGITAL_INPUT_SIZE) > maxsize) {
+    return 0;
+  }
+  buffer[cursor++] = channel;
+  buffer[cursor++] = LPP_DIGITAL_INPUT;
+  buffer[cursor++] = value;
+
+  return cursor;
+}
+
+uint8_t CayenneLPP::addDigitalOutput(uint8_t channel, uint8_t value) {
+  if ((cursor + LPP_DIGITAL_OUTPUT_SIZE) > maxsize) {
+    return 0;
+  }
+  buffer[cursor++] = channel;
+  buffer[cursor++] = LPP_DIGITAL_OUTPUT;
+  buffer[cursor++] = value;
+
+  return cursor;
+}
+
+uint8_t CayenneLPP::addAnalogInput(uint8_t channel, float value) {
+  if ((cursor + LPP_ANALOG_INPUT_SIZE) > maxsize) {
+    return 0;
+  }
+
+  int16_t val = value * 100;
+  buffer[cursor++] = channel;
+  buffer[cursor++] = LPP_ANALOG_INPUT;
+  buffer[cursor++] = val >> 8;
+  buffer[cursor++] = val;
+
+  return cursor;
+}
+
+uint8_t CayenneLPP::addAnalogOutput(uint8_t channel, float value) {
+  if ((cursor + LPP_ANALOG_OUTPUT_SIZE) > maxsize) {
+    return 0;
+  }
+  int16_t val = value * 100;
+  buffer[cursor++] = channel;
+  buffer[cursor++] = LPP_ANALOG_OUTPUT;
+  buffer[cursor++] = val >> 8;
+  buffer[cursor++] = val;
+
+  return cursor;
+}
+
+uint8_t CayenneLPP::addLuminosity(uint8_t channel, uint16_t lux) {
+  if ((cursor + LPP_LUMINOSITY_SIZE) > maxsize) {
+    return 0;
+  }
+  buffer[cursor++] = channel;
+  buffer[cursor++] = LPP_LUMINOSITY;
+  buffer[cursor++] = lux >> 8;
+  buffer[cursor++] = lux;
+
+  return cursor;
+}
+
+uint8_t CayenneLPP::addPresence(uint8_t channel, uint8_t value) {
+  if ((cursor + LPP_PRESENCE_SIZE) > maxsize) {
+    return 0;
+  }
+  buffer[cursor++] = channel;
+  buffer[cursor++] = LPP_PRESENCE;
+  buffer[cursor++] = value;
+
+  return cursor;
+}
+
+uint8_t CayenneLPP::addTemperature(uint8_t channel, float celsius) {
+  if ((cursor + LPP_TEMPERATURE_SIZE) > maxsize) {
+    return 0;
+  }
+  int16_t val = celsius * 10;
+  buffer[cursor++] = channel;
+  buffer[cursor++] = LPP_TEMPERATURE;
+  buffer[cursor++] = val >> 8;
+  buffer[cursor++] = val;
+
+  return cursor;
+}
+
+uint8_t CayenneLPP::addRelativeHumidity(uint8_t channel, float rh) {
+  if ((cursor + LPP_RELATIVE_HUMIDITY_SIZE) > maxsize) {
+    return 0;
+  }
+  buffer[cursor++] = channel;
+  buffer[cursor++] = LPP_RELATIVE_HUMIDITY;
+  buffer[cursor++] = rh * 2;
+
+  return cursor;
+}
+
+uint8_t CayenneLPP::addAccelerometer(uint8_t channel, float x, float y, float z) {
+  if ((cursor + LPP_ACCELEROMETER_SIZE) > maxsize) {
+    return 0;
+  }
+  int16_t vx = x * 1000;
+  int16_t vy = y * 1000;
+  int16_t vz = z * 1000;
+
+  buffer[cursor++] = channel;
+  buffer[cursor++] = LPP_ACCELEROMETER;
+  buffer[cursor++] = vx >> 8;
+  buffer[cursor++] = vx;
+  buffer[cursor++] = vy >> 8;
+  buffer[cursor++] = vy;
+  buffer[cursor++] = vz >> 8;
+  buffer[cursor++] = vz;
+
+  return cursor;
+}
+
+uint8_t CayenneLPP::addBarometricPressure(uint8_t channel, float hpa) {
+  if ((cursor + LPP_BAROMETRIC_PRESSURE_SIZE) > maxsize) {
+    return 0;
+  }
+  int16_t val = hpa * 10;
+
+  buffer[cursor++] = channel;
+  buffer[cursor++] = LPP_BAROMETRIC_PRESSURE;
+  buffer[cursor++] = val >> 8;
+  buffer[cursor++] = val;
+
+  return cursor;
+}
+
+uint8_t CayenneLPP::addGyrometer(uint8_t channel, float x, float y, float z) {
+  if ((cursor + LPP_GYROMETER_SIZE) > maxsize) {
+    return 0;
+  }
+  int16_t vx = x * 100;
+  int16_t vy = y * 100;
+  int16_t vz = z * 100;
+
+  buffer[cursor++] = channel;
+  buffer[cursor++] = LPP_GYROMETER;
+  buffer[cursor++] = vx >> 8;
+  buffer[cursor++] = vx;
+  buffer[cursor++] = vy >> 8;
+  buffer[cursor++] = vy;
+  buffer[cursor++] = vz >> 8;
+  buffer[cursor++] = vz;
+
+  return cursor;
+}
+
+uint8_t CayenneLPP::addGPS(uint8_t channel, float latitude, float longitude, float meters) {
+  if ((cursor + LPP_GPS_SIZE) > maxsize) {
+    return 0;
+  }
+  int32_t lat = latitude * 10000;
+  int32_t lon = longitude * 10000;
+  int32_t alt = meters * 100;
+
+  buffer[cursor++] = channel;
+  buffer[cursor++] = LPP_GPS;
+
+  buffer[cursor++] = lat >> 16;
+  buffer[cursor++] = lat >> 8;
+  buffer[cursor++] = lat;
+  buffer[cursor++] = lon >> 16;
+  buffer[cursor++] = lon >> 8;
+  buffer[cursor++] = lon;
+  buffer[cursor++] = alt >> 16;
+  buffer[cursor++] = alt >> 8;
+  buffer[cursor++] = alt;
+
+  return cursor;
+}
--- a/examples/lora/sensor-lorawan/Makefile
+++ b/examples/lora/sensor-lorawan/Makefile
@@ -10,8 +10,14 @@ CXX_SRCS := $(wildcard *.cc)
 
 override CPPFLAGS += -DRADIOLIB_CLOCK_DRIFT_MS=9
 
-ifneq ($(CI),)
+# If we are building for a testing configuration (e.g. CI) then it's okay to
+# use the example config header. However, if someone is doing local testing,
+# especially of a different thing, we don't want to accidentally overwrite
+# the build obeject here with the example config when a real one exists.
+ifneq ($(TOCK_BUILDALL),)
+ifeq ($(wildcard radioConfig.h),)
 override CPPFLAGS += "-DRADIO_CONFIG_CI=radioConfig_example.h"
+endif
 endif
 
 # Use the libtock-c Make system
--- a/examples/lora/sensor-lorawan/Makefile
+++ b/examples/lora/sensor-lorawan/Makefile
@@ -1,0 +1,20 @@
+# Makefile for user application
+
+# Specify this directory relative to the current application.
+TOCK_USERLAND_BASE_DIR = ../../..
+
+STACK_SIZE = 4096
+
+# Which files to compile.
+CXX_SRCS := $(wildcard *.cc)
+
+override CPPFLAGS += -DRADIOLIB_CLOCK_DRIFT_MS=9
+
+ifneq ($(CI),)
+override CPPFLAGS += "-DRADIO_CONFIG_CI=radioConfig_example.h"
+endif
+
+# Use the libtock-c Make system
+EXTERN_LIBS += $(TOCK_USERLAND_BASE_DIR)/RadioLib
+
+include $(TOCK_USERLAND_BASE_DIR)/AppMakefile.mk
--- a/examples/lora/sensor-lorawan/Makefile
+++ b/examples/lora/sensor-lorawan/Makefile
@@ -24,3 +24,12 @@ endif
 EXTERN_LIBS += $(TOCK_USERLAND_BASE_DIR)/RadioLib
 
 include $(TOCK_USERLAND_BASE_DIR)/AppMakefile.mk
+
+# Protect from the (unlikely) case where the app happens to have been
+# built from an unrelated `build all` event, but now the user is trying
+# to flash the app with an invalid configuration.
+flash: radioConfig.h
+
+program: radioConfig.h
+
+install: radioConfig.h
--- a/examples/lora/sensor-lorawan/README.md
+++ b/examples/lora/sensor-lorawan/README.md
@@ -1,0 +1,30 @@
+Sensor LoRaWAN Transmitter
+==========================
+
+This example builds an application to transmit sensor data via LoRaWAN.
+
+See https://github.com/jgromes/RadioLib/blob/master/examples/LoRaWAN/LoRaWAN_Starter/notes.md
+for notes on setting up the LoRaWAN device.
+
+The most important part is creating a radioConfig.h file with the secrets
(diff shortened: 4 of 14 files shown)


alistair23:
Any ideas?

```
2024-07-14T10:37:45.7525104Z riscv64-unknown-elf-g++ -Wctor-dtor-privacy  -Wdelete-non-virtual-dtor  -Wold-style-cast  -Woverloaded-virtual  -Wsign-promo  -Wstrict-null-sentinel  -Wsuggest-final-methods  -Wsuggest-final-types  -Wsuggest-override  -Wuseless-cast  -Wzero-as-null-pointer-constant  -isystem ../../..//libradio/RadioLib/src -isystem ../../..//libradio/RadioLib/examples/NonArduino/Tock -I../../..// -DRADIOLIB_CLOCK_DRIFT_MS=9 -frecord-gcc-switches -gdwarf-2 -Os -fdata-sections -ffunction-sections -fstack-usage -D_FORTIFY_SOURCE=2 -Wall -Wextra -I../../../ -Wdate-time  -Wfloat-equal  -Wformat-nonliteral  -Wformat-security  -Wformat-y2k  -Winit-self  -Wmissing-declarations  -Wmissing-field-initializers  -Wmissing-format-attribute  -Wmissing-noreturn  -Wmultichar  -Wpointer-arith  -Wredundant-decls  -Wshadow  -Wunused-macros  -Wunused-parameter  -Wwrite-strings  -Wstack-usage=4096 -Wlogical-op  -Wtrampolines  -isystem ../../..//lib/libtock-newlib-4.2.0.20211231/riscv/riscv64-unknown-elf/include -isystem ../../..//lib/libtock-libc++-10.5.0/riscv/riscv64-unknown-elf/include/c++/10.5.0 -isystem ../../..//lib/libtock-libc++-10.5.0/riscv/riscv64-unknown-elf/include/c++/10.5.0/riscv64-unknown-elf -march=rv32imac -mabi=ilp32 -mcmodel=medlow -c -o build/rv32imac/CayenneLPP.o CayenneLPP.cc
2024-07-14T10:37:45.7594123Z In file included from CayenneLPP.h:20,
2024-07-14T10:37:45.7594817Z                  from CayenneLPP.cc:16:
2024-07-14T10:37:45.7596462Z ../../..//lib/libtock-libc++-10.5.0/riscv/riscv64-unknown-elf/include/c++/10.5.0/cstdlib:75:15: fatal error: stdlib.h: No such file or directory
2024-07-14T10:37:45.7597996Z    75 | #include_next <stdlib.h>
2024-07-14T10:37:45.7598516Z       |               ^~~~~~~~~~
2024-07-14T10:37:45.7599024Z compilation terminated.
2024-07-14T10:37:45.7608222Z make: *** [../../..//AppMakefile.mk:340: build/rv32imac/CayenneLPP.o] Error 1
2024-07-14T10:37:45.7635538Z ##[error]Process completed with exit code 2.
2024-07-14T10:37:45.7718454Z Post job cleanup.
```

examples/lora/sensor-lorawan/CayenneLPP.cc
```diff
@@ -0,0 +1,219 @@
...
+ * is compliant with the payload size restriction, which can be lowered down to
+ * 11 bytes, and allows the device to send multiple sensor data at one time.
+ *
+ * CayenneLPP is also support by TTN:
+ *     https://www.thethingsindustries.com/docs/integrations/payload-formatters/cayenne/
+ *
+ * This code is direct from: https://github.com/myDevicesIoT/CayenneLPP
+ *
+ * GPL-3.0 license
+ * See https://github.com/myDevicesIoT/CayenneLPP for license details
```
ppannuto, line 13: I am so not a lawyer, but I don't think we have any other GPL in this repo, so I'm a bit wary (mostly out of ignorance) of whether this can cause any complications for us / others.

Right this second not passing judgement one way or another, just elevating attention to this so we can talk/think about it.

examples/lora/sensor-lorawan/main.cc
```diff
@@ -0,0 +1,112 @@
...
+
+/* These need to be updated to use values from your LoRaWAN server */
+uint64_t joinEUI = 0x0000000000000000;
+uint64_t devEUI  = 0x0000000000000000;
+uint8_t nwkKey[] = {0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00};
+uint8_t appKey[] = {0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00};
+
+// regional choices: EU868, US915, AU915, AS923, IN865, KR920, CN780, CN500
+const LoRaWANBand_t* Region = &AU915;
+const uint8_t subBand       = 2;
```
ppannuto, line 29: The way I have seen this handled before, which I think might be good to do here, is something like a `#include "radioConfig.h"` to pick these up.

We commit to the repo a `radioConfig_example.h` or similar, and add `radioConfig.h` to `.gitignore` for this folder.

That does two important things:
 (1) It forces people to think about region explicitly before anything will work (bonus points for `const LoRaWANBand_t* Region = ...` or something else that literally won't compile in the example file)
 (2) It makes it easier for folks to keep a working local example without always having a diff lying around in their tree.

ppannuto (changes requested):
This is cool! TTN is great for having an 'out-of-box' working example. If I can find a minute, I'll try to do a similar for Helium at some point (which has much better penetration in North America than TTN).

examples/lora/sensor-lorawan/main.cc
alistair23: I like that idea, but that will fail to build with the CI won't it?

alistair23:
CI fixed!

examples/lora/sensor-lorawan/main.cc
ppannuto: Hrmm... that's an interesting wrinkle. Here's one idea:

In the app Makefile, add `radioConfig.h` as a dependency of `main.cc` (this will have the nice side benefit of make erroring for users for the missing required file)

Also in the Makefile, we can do:
```make
ifneq ($(CI),)
radioConfig.h: radioConfig_example.h
        cp radioConfig_example.h radioConfig.h
endif
```

---

Hmm..., that might have the negative effect of overwriting the local config if anyone does some local CI testing though...

maybe in main:

```c
#ifndef RADIO_CONFIG_H
#define RADIO_CONFIG_H radioConfig.h
#endif

#include "RADIO_CONFIG_H"
```
 and then in the Makefile:
```make
ifneq ($(CI),)
CPPFLAGS += "-DRADIO_CONFIG_H=radioConfigExample.h"
endif
```

---

(forgive minor syntax errors in conceptual code here please, but hopefully idea is clear)

ppannuto:
Marking as blocked until we can resolve the GPL question.

@tock/core-wg This is something we need to reason through, either on a core team call, or out of band.

tl;dr:
 - Is it reasonable to have GPL'd code in the main `tock/libtock-c` repo?
 - Does it change the discussion if GPL code is restricted to only be allowed in individual `examples` folders?
     - Or even perhaps a `examples/license-restricted/...` subfolder?

alevy:
> Is it reasonable to have GPL'd code in the main tock/libtock-c repo?

I don't think anyone is reasonably getting in any trouble here as long as they either don't depend on this library for a distributed program, or license any program that does depend on it under GPL, but the question is reasonable and understanding the answer is probably important.

My interpretation is that including it in this manner is fine. Anyone who doesn't use it in their program won't actually link against it, "components" in the libtock-c repo are ultimately licensed individually anyway (primarily under MIT/Apache), and the affects for an actual program compiled/distributed from this repo depend on what is actually distributed and what it depends on.

I'd be more concerned if we made any core parts of userland depend on GPL code.

If C had a good package management story, we would anyway just be saying that programs should depend on this other library, distributed elsewhere to use it, which would have the same implications and I don't think the question would come up. In this case, we'd just be hosting a copy of that library in the same repo. I don't see a significant difference.

Basically my vote is that this is fine. It's only  actually a dependency if it's used for a particular application (in which case that application as a whole I believe would need to be distributed under GPL).

>Does it change the discussion if GPL code is restricted to only be allowed in individual examples folders?

I think it's cleaner if the GPL code lives exclusively in a single example, less clean but perfectly good (and more reasonable) if it is only ever _used_/_linked_ from individual example applications under the examples folder, and a red flag if we start linking against it from dependencies of applications (don't see that happening here).


> Or even perhaps a examples/license-restricted/... subfolder?

I think this would be overkill.

TockLibrary.mk
```diff
@@ -29,10 +29,6 @@ $(call check_defined, LIBNAME)
 $(call check_defined, $(LIBNAME)_DIR)
 $(call check_defined, $(LIBNAME)_SRCS)
 
-ifeq ($(strip $($(LIBNAME)_SRCS)),)
-  $(error Library "$(LIBNAME)" has no SRCS?)
-endif
-
```
bradjc, line 35: Do we need to remove this?

examples/lora/sensor-lorawan/Makefile
```diff
@@ -0,0 +1,36 @@
...
+CXX_SRCS := $(wildcard *.cc)
+
+# Include the core RadioLib headers
+override CPPFLAGS += -isystem $(TOCK_USERLAND_BASE_DIR)/libradio/RadioLib/src
+
+# Include the Tock specific headers
+override CPPFLAGS += -isystem $(TOCK_USERLAND_BASE_DIR)/libradio/RadioLib/examples/NonArduino/Tock
+
+# Include the base of libtock-c to fix the libtock/ includes from RadioLib
+override CPPFLAGS += -I$(TOCK_USERLAND_BASE_DIR)/ -DRADIOLIB_CLOCK_DRIFT_MS=9
```
bradjc, line 18: Do we need `-I$(TOCK_USERLAND_BASE_DIR)`? I think that should already be happening.

examples/lora/sensor-lorawan/CayenneLPP.cc
kupiakos: At least two things should be changed:

- The GPL license should be included in full somewhere in the repo, like in a `LICENSE.GPL` file.
- An `SPDX-License-Identifier` tag near the top of the file so this file doesn't accidentally get redistributed in a non-GPL-compliant manner

examples/lora/sensor-lorawan/Makefile
bradjc, line 18: I _think_ these should go in `Makefile.app`. I assume every app using radiolib would want these?

examples/lora/sensor-lorawan/Makefile
```diff
@@ -0,0 +1,36 @@
...
+endif
+
+# Use the libtock-c Make system
+LIBS_cortex-m0 += $(TOCK_USERLAND_BASE_DIR)/libradio/RadioLib/build/cortex-m0/RadioLib.a
+LIBS_cortex-m3 += $(TOCK_USERLAND_BASE_DIR)/libradio/RadioLib/build/cortex-m3/RadioLib.a
+LIBS_cortex-m4 += $(TOCK_USERLAND_BASE_DIR)/libradio/RadioLib/build/cortex-m4/RadioLib.a
+LIBS_cortex-m7 += $(TOCK_USERLAND_BASE_DIR)/libradio/RadioLib/build/cortex-m7/RadioLib.a
+
+LIBS_rv32imc += $(TOCK_USERLAND_BASE_DIR)/libradio/RadioLib/build/rv32imc/RadioLib.a
+LIBS_rv32imac += $(TOCK_USERLAND_BASE_DIR)/libradio/RadioLib/build/rv32imac/RadioLib.a
```
bradjc, line 31: I think this should happen automatically.

examples/lora/sensor-lorawan/Makefile
```diff
@@ -0,0 +1,36 @@
...
+
+EXTERN_LIBS := $(TOCK_USERLAND_BASE_DIR)/libradio/RadioLib
+include $(TOCK_USERLAND_BASE_DIR)/libradio/Makefile
```
bradjc, line 34: This `include` should happen as a result of setting `EXTERN_LIBS`.

TockLibrary.mk
alistair23: Yep https://github.com/tock/libtock-c/pull/432#discussion_r1645043185

examples/lora/sensor-lorawan/Makefile
alistair23: It doesn't get pulled in correctly as we need to use `libradio/RadioLib` for the `EXTERN_LIBS`

alevy:
OK, looking into this a bit more deeply, two things:

1. I think the GPL thing is a non-issue the way that this PR uses it. Perhaps worth sticking a notice in the relevant example apps.
2. This PR makes apps/libraries that are highly hardware specific (the specific lora chip configured for a _specific_ board that uses particular pins) in a way that is really unnecessary. Let me explain...

RadioLib is implemented in a way that allows for pluggable hardware implementations via the [`PhysicalLayer`](https://jgromes.github.io/RadioLib/class_physical_layer.html) class, which the `SX1262` class instantiated in the examples is a subclass of.

A trivially portable version of this would implement an instance of `PhysicalLayer` that interacts with a system call driver---this would be akin to the way we use OpenThread on top of an abstract 802154 syscall driver, and would allow a LoraWAN application to run on any board + module configuration.

Now, I realize that implementing a driver for the SX1262 is non-trivial. It's ~1000 LoC in RadioLib. `lora-rs` has an implementation in slightly less, it does not seem particularly timing sensitive or subtle, but nonetheless there are a lot of details to get right.

Now, it's clearly possible to use Tock + libtock-c to build these kinds of applications (as the PR clearly demonstrates). It's perfectly fine to have applications out in the world that don't adhere to the particular layering choice that we advocate upstream, and _generally_ supporting exporting a SPI device (in this case) directly to userspace is good to support for that reason. 

However, I think the question is: do we want to have these kinds of examples upstream? No doubt we already have a few, but I'd argue those are basically legacy or tests and should be removed (or retained merely as tests).

I suggest there are at least two options:

1. No. These kinds of applications need to live elsewhere.
2. No, but there can be in-progress work that lives in upstream. E.g., an SX1262 implementation and Lora device system call driver is in the process of being implemented in the kernel, until that's ready, we have an example that uses the C implementation in userspace, but progress is being made to change that.

examples/lora/sensor-lorawan/Makefile
bradjc: The library makefile should adjust the paths so that all an app developer needs to do is EXTERN_LIBS. Our other libraries (eg openthread and u8g2) are good examples of how this can be done.

bradjc:
> However, I think the question is: do we want to have these kinds of examples upstream? No doubt we already have a few, but I'd argue those are basically legacy or tests and should be removed (or retained merely as tests).

I agree, but I'm also torn. It took a year implementing the lr1110 stack using the same SPI/GPIO HAL interface (#458). The examples in that would too only work for one radio chip. The radio lib [driver](https://github.com/jgromes/RadioLib/blob/master/src/modules/LR11x0/LR11x0.cpp) for that is massive (and is also very new and didn't exist when we started). Like OpenThread, I kind of feel like these are necessary evils to make progress on other networking goals.

alistair23:
> * This PR makes apps/libraries that are highly hardware specific (the specific lora chip configured for a _specific_ board that uses particular pins) in a way that is really unnecessary. Let me explain...

Have you read the entire PR? This is specifically not board specific (obviously you need a LoRa radio).

The GPIO pin mapping and SPI allocation is handled in the kernel. So this application can run on *any* board with a SX1261. `main.rs` in the kernel maps the GPIO pins and SPI bus, so that the application doesn't have to worry about it.

I agree that this example only supports the SX1261 module. It is easy to change the module in RadioLib to support any module that RadioLib supports, such as the LR1110 for example.

It obviously would be great to have the module support in the kernel. But that's yet another thing that needs to be re-written in a Tock way. So until someone goes ahead and does that I think this is a good start and works well. This implementation doesn't stop anyone from writing an implementation in the kernel and migrating to that in the future.

It's probably also worth noting that the https://github.com/tock/libtock-c/pull/458 implementation only works on a single board. So maybe that needs more scrutiny?

alevy:
@alistair23 from the code:

```c
  // now we can create the radio module
  // pinout corresponds to the SparkFun LoRa Thing Plus - expLoRaBLE
  // NSS pin:   0
  // DIO1 pin:  2
  // NRST pin:  4
  // BUSY pin:  1
```

> Have you read the entire PR? This is specifically not board specific (obviously you need a LoRa radio).

I have read the entire PR. Though I'm realizing now, that I had not read the RadioLib Tock adaptation library, which depends on libtock-c, which is a convoluted enough dependency that I admit I don't really understand now what happens with those pins that are passed in.

> It's probably also worth noting that the https://github.com/tock/libtock-c/pull/458 implementation only works on a single board. So maybe that needs more scrutiny?

I agree, and I have not yet really looked at that PR, though one important difference is that #458 adds tests only, while this is moving examples from `wip` to regular end-to-end examples.

alistair23:
> I have read the entire PR. Though I'm realizing now, that I had not read the RadioLib Tock adaptation library, which depends on libtock-c, which is a convoluted enough dependency that I admit I don't really understand now what happens with those pins that are passed in.

Fair. The comment is trying to describe that this is what I tested with.

I did also test with a SX1261 externally attached via SPI, which also works. That does require changes to `main.rs` in the kernel though for the different SPI bus and GPIO pins.

The code in Tock where the GPIO lines are mapped looks like this

```rust
    let sx1262_gpio = components::gpio::GpioComponent::new(
        board_kernel,
        LORA_GPIO_DRIVER_NUM,
        components::gpio_component_helper!(
            apollo3::gpio::GpioPin,
            0 => &peripherals.gpio_port[36], // H6 - SX1262 Slave Select
            1 => &peripherals.gpio_port[39], // J8 - SX1262 Radio Busy Indicator
            2 => &peripherals.gpio_port[40], // J9 - SX1262 Multipurpose digital I/O (DIO1)
            3 => &peripherals.gpio_port[47], // H9 - SX1262 Multipurpose digital I/O (DIO3)
            4 => &peripherals.gpio_port[44], // J7 - SX1262 Reset
        ),
    )
    .finalize(components::gpio_component_static!(apollo3::gpio::GpioPin));
```

> I agree, and I have not yet really looked at that PR, though one important difference is that #458 adds tests only, while this is moving examples from `wip` to regular end-to-end examples.

https://github.com/tock/libtock-c/pull/432 is actually the PR that moves LoRa to examples. The PR backlog is so large that PRs are starting to build on each other :)

For clarity, this was originally in examples and [moved out](https://github.com/tock/libtock-c/pull/401#issuecomment-2057681156) because it didn't support RISC-V (which it does now)

ppannuto:
@alistair23 Once #432 is in, can you rebase this? I think my disposition is similar to Brad, i.e., networking stuff is hard and a working example has real value, even if it's a bit less hardware agnostic / purely layered than some idea thing might be, but I'll look at that more in depth early next week.

alistair23:
Rebased, this is just a single commit and can be merged once https://github.com/tock/libtock-c/pull/432 is

ppannuto:
Okay! #432 is in, but this needs one more quick rebase now it looks like.

alistair23:
> Okay! #432 is in, but this needs one more quick rebase now it looks like.

Done

examples/lora/sensor-lorawan/main.cc
```diff
@@ -0,0 +1,107 @@
...
+#define MAX_SIZE 10
+
+// the entry point for the program
+int main(void) {
+  CayenneLPP Payload(MAX_SIZE);
+
+  printf("[SX1261] Initialising Radio ... \n");
+
+  // create a new instance of the HAL class
+  TockHal* hal = new TockHal();
```
ppannuto, line 38:
```suggestion
  TockRadioLibHal* hal = new TockRadioLibHal();
```

Or I suppose since this is C++, even better would be to wrap everything in a `namespace RadioLib` (or `RadioLibHal` or `TockRadioLib::{Hal, Module}`, etc — I might like this last option best but don't feel strongly). The same issue applies for the "`Module`" below — these are generic terms that make composition really hard.

[I know this comes from upstream, so the comment doesn't apply to this PR directly per se, but IIUC, you're also driving the Tock interface in RadioLib, so hopefully this is an easy enough change to push there and propagate here via submodule hash bump]

examples/lora/sensor-lorawan/README.md
```diff
@@ -0,0 +1,16 @@
...
+for notes on setting up the LoRaWAN device.
+
+The most important part is creating a radioConfig.h file with the secrets
+from your LoRaWAN server and any country specific settings. There is an
+existing radioConfig_example.h which can be used as a useful starting point.
+
+This has been tested against The Things Network. Before changing settings
+make sure you consider regulatory duty cycles and TTN's Fair Usage Policy,
+for example don't drop the delay in the loop, otherwise you will break the
+TTN Fair Usage Policy.
```
ppannuto, line 16: Add "example expected output" section

ppannuto (changes requested):

examples/lora/sensor-lorawan/Makefile
```diff
@@ -0,0 +1,20 @@
...
+CXX_SRCS := $(wildcard *.cc)
+
+override CPPFLAGS += -DRADIOLIB_CLOCK_DRIFT_MS=9
+
+ifneq ($(CI),)
+override CPPFLAGS += "-DRADIO_CONFIG_CI=radioConfig_example.h"
+endif
+
+# Use the libtock-c Make system
+EXTERN_LIBS += $(TOCK_USERLAND_BASE_DIR)/RadioLib
```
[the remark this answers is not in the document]
alistair23: There is a `Makefile.setup` at: `RadioLib/Makefile.setup`, I think that's all that is required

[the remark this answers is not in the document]
ppannuto: My error, sorry; my local checkout of your branch hadn't fetched the update post-rebase when I was testing yesterday.

ppannuto:
@alistair23 FYI—I just pushed a commit to your branch here slightly changing the radio config override logic (so that a local run of `examples/build_all.sh` still works but also doesn't clobber a real config).

ppannuto:
Crap. And with apologies, just force-pushed a quick fix to my fix—forgot the `export`

alistair23:
Great! This shoudl be ready then.

I'll look at addressing https://github.com/tock/libtock-c/pull/456#discussion_r1819519494 in upstream

examples/build_all.sh
```diff
@@ -2,6 +2,8 @@
 
 NUM_JOBS=$(nproc 2>/dev/null || sysctl -n hw.ncpu 2>/dev/null || 4)
 
+export TOCK_BUILDALL=1
```
alevy, line 5: What is the meaning of this argument in a way that is meaningful to someone reading this script, without the context of the specific narrow use in the sensor-lorawan example Makefile, such that they could evaluate, e.g.: whether to have it on; whether it's being used correctly; whether it's ok to remove it because they're coming back to this in a year and it's unclear whether/why it was ever used; etc...

alistair23: @ppannuto You added this can you help explain here

ppannuto: Added an explanation and example.

alevy (approved):

alevy:
@alistair23 needs a rebase. I would do it, but wouldn't be able to test if rebasing somehow broke anything in this PR

alistair23:
Rebased!

alevy (approved):

ppannuto:
Re https://github.com/tock/libtock-c/pull/456#discussion_r1819519494, I updated everything that I think needs doing, but I can't actually build anything to test because their cmake doesn't correctly separate platform detection from compiler feature selection.* However, hopefully @alistair23  can pull / clone / test https://github.com/ppannuto/RadioLib/tree/tock-hal-updates and submit a RadioLib PR?

> *Unimportant details that mostly validate my continued avoidance of cmake:
<details>

```
[-bash] Wed 06 Nov 14:29 [[tock-hal-updates] ~/code/ppannuto/RadioLib/examples/NonArduino/Tock]
$ LIBTOCK_C_DIRECTORY="$(pwd)/libtock-c" ./build.sh
...
-- The C compiler identification is AppleClang 16.0.0.16000026
-- The CXX compiler identification is AppleClang 16.0.0.16000026
...
[  6%] Building CXX object RadioLib/CMakeFiles/RadioLib.dir/src/modules/CC1101/CC1101.cpp.o
arm-none-eabi-g++: error: unrecognized command-line option '-mmacosx-version-min=14.6'
arm-none-eabi-g++: error: unrecognized command-line option '-mmacosx-version-min=14.6'
arm-none-eabi-g++: error: unrecognized command-line option '-mmacosx-version-min=14.6'
make[2]: *** [RadioLib/CMakeFiles/RadioLib.dir/src/modules/CC1101/CC1101.cpp.o] Error 1
make[2]: *** Waiting for unfinished jobs....
make[2]: *** [RadioLib/CMakeFiles/RadioLib.dir/src/hal/Arduino/ArduinoHal.cpp.o] Error 1
make[2]: *** [RadioLib/CMakeFiles/RadioLib.dir/src/Hal.cpp.o] Error 1
arm-none-eabi-g++: error: unrecognized command-line option '-mmacosx-version-min=14.6'
make[2]: *** [RadioLib/CMakeFiles/RadioLib.dir/src/Module.cpp.o] Error 1
make[1]: *** [RadioLib/CMakeFiles/RadioLib.dir/all] Error 2
make: *** [all] Error 2
```
</details>

alevy:
@ppannuto I can't tell if you're still blocking this PR or just suggesting a future improvement to RadioLib.

ppannuto:
Blocking.

There's a bunch of stuff in the radiolib tock header that would pollute the global namespace in a challenging way for anything to co-exist with radiolib being included in your app.

It's a bit weird that the Tock HAL file is in the RadioLib repo and not in the libtock-c repo IMHO, but that seems to be the norm for the RadioLib project, so I'm letting that be for the moment. But it does put us in an odd situation of having comments on this PR that will ultimately be resolved by a version bump of a submodule.

alistair23:
PR opened: https://github.com/jgromes/RadioLib/pull/1313

It breaks their CI though, as there is a backwards incompatible name change.

I have updated this PR to use the new names and not break the CI. So once this is merged I can get https://github.com/jgromes/RadioLib/pull/1313 merged and fix everything up.

Let's try not to block this for too long

alistair23:
Ping! It'd be good to get this in so then I can get https://github.com/jgromes/RadioLib/pull/1313 merged

jgromes:
@ppannuto hey there! Just my two cents on one of the points raised:

> It's a bit weird that the Tock HAL file is in the RadioLib repo and not in the libtock-c repo IMHO, but that seems to be the norm for the RadioLib project

It probably makes more sense to move the Tock HAL file to the tock project itself. The HAL files present in RadioLib started out as examples that turned out to be more easy to maintain as part of the project source. However, unlike the other HALs which are pretty generic (e.g. for Raspberry Pi), the Tock HAL is very specific. So I'm not opposed to moving it, in fact I would support it ;)

alevy (dismissed):

alevy:
@ppannuto can we maybe move forward with this and kind of iterate our way past the circular dependency?

ppannuto:
I pushed an update that copies over my updated RadlioLib Tock HAL to the libtock-c repo, and has the application use our HAL to build.

@alistair23 This is compile-tested only; sorry, I don't have this hardware on-hand.

alevy (approved):

alevy:
OK, I think we just want @alistair23 to confirm this still works.

alistair23:
> OK, I think we just want @alistair23 to confirm this still works.

Still works!

ppannuto (approved):

PR #458, opened 2024-07-22
library: lr1110: Add library for LR1110 support

ethanzhouyc:
Add lr1110 library and two related example applications. This library uses [Seeed Studio's code](https://github.com/Seeed-Studio/Seeed_Wio_WM1110_Dev_Board) as user space, built upon Tock, to provide support for the LR1110 LoRa chip on Wio-WM1110 development board. Applications could use the library to collect sensor data, join TTN, perform Wi-Fi scans, and send data to TTN for geolocation.


alistair23 (changes requested):
Cool! Soon Tock will have so much LoRa support!

bradjc:
I'm working on putting the finishing touches on this pr

YanchengZhou:
@bradjc Maybe we should keep wifi_helpers.c in the src_changed folder for now. Changes in [these lines](https://github.com/tock/libtock-c/pull/458/commits/4ad88a703a3e4e7f3a557342a804be8ef7600e1c#diff-cf252f0bdf9d6e5d1d3d4110031fc95840d8bc8b46206319fc878ec1c5956242R161) make sure that the app does not return error when the board sometimes scans too many Wi-Fi access points.

YanchengZhou:
Great to see the library been further cleaned up!

bradjc:
> @bradjc Maybe we should keep wifi_helpers.c in the src_changed folder for now. Changes in [these lines](https://github.com/tock/libtock-c/pull/458/commits/4ad88a703a3e4e7f3a557342a804be8ef7600e1c#diff-cf252f0bdf9d6e5d1d3d4110031fc95840d8bc8b46206319fc878ec1c5956242R161) make sure that the app does not return error when the board sometimes scans too many Wi-Fi access points.

Good point. I think I was thinking we should change `WIFI_MAX_RESULTS`, but that isn't straightforward easy, and could run into the same issue.

bradjc:
I wasn't able to actually get this to work, and I _think_ I've traced the issue back to changes to how the alarm_read function works. I forgot to compile the library when I tested so I still need to verify this, but this is my working diff:

```diff
diff --git a/lr1110/lr1110/src_changed/smtc_hal_rtc.c b/lr1110/lr1110/src_changed/smtc_hal_rtc.c
index 39cc844e..15cb80ac 100644
--- a/lr1110/lr1110/src_changed/smtc_hal_rtc.c
+++ b/lr1110/lr1110/src_changed/smtc_hal_rtc.c
@@ -10,7 +10,7 @@
 // static uint32_t rtc_2_tick_diff = 0;

 // static void rtc_2_handler( nrf_drv_rtc_int_type_t int_type )
-// {
+// {
 //     if( int_type == NRF_DRV_RTC_INT_OVERFLOW )
 //     {
 //         rtc_2_tick_diff += RTC_2_MAX_TICKS;
@@ -43,17 +43,47 @@ uint32_t hal_rtc_get_time_ms( void )
     return timeInms / 10;
 }

+static uint32_t ticks_to_100us(uint32_t ticks) {
+  // `ticks_to_ms`'s conversion will be accurate to within the range
+  // 0 to 1 milliseconds less than the exact conversion
+  // (true millisecond conversion - [0,1) milliseconds).
+
+  uint32_t frequency;
+  libtock_alarm_command_get_frequency(&frequency);
+
+  uint32_t seconds = (ticks / frequency);
+  uint32_t hundredus_per_second = 10000;
+
+  // Calculate the conversion of full seconds to ticks.
+  uint32_t hundredus = seconds * hundredus_per_second;
+
+  // To get conversion accuracy within 1 millisecond, the conversion
+  // must also convert partial seconds.
+  uint32_t leftover_ticks = ticks % frequency;
+
+  // This calculation is mathematically equivalent to doing:
+  //
+  //   `leftover_ticks` / (`frequency` / `hundredus_per_second`)
+  //
+  // This is taking the remaining unconverted ticks (leftover_ticks)
+  // and dividing by the number of ticks per millisecond
+  // (`frequency` (ticks per second) / `1000` hundredus per second)
+  // The division is done this way because of the same argument in
+  // `ms_to_ticks`.
+  hundredus += (leftover_ticks * hundredus_per_second) / frequency;
+
+  return hundredus;
+}
+
 uint32_t hal_rtc_get_time_100us( void )
 {
     // double temp = nrf_drv_rtc_counter_get( &rtc_2 ) + rtc_2_tick_diff;
     // return temp * RTC_2_PER_TICK * 10;
-
+
     uint32_t time = 0;
     uint32_t frequency = 0;
     libtock_alarm_command_read(&time);
-    libtock_alarm_command_get_frequency(&frequency);
-    uint32_t timeIn100us = (time * 10000) / frequency;
-    return timeIn100us;
+    return ticks_to_100us(time);
 }

 // uint32_t hal_rtc_get_max_ticks( void )
@@ -64,7 +94,7 @@ uint32_t hal_rtc_get_time_100us( void )
 void hal_rtc_wakeup_timer_set_ms( const int32_t milliseconds )
 {
     libtocksync_alarm_delay_ms(milliseconds);
-
+
     // uint32_t temp = 0;
     // float temp_f = milliseconds;
     // temp_f = temp_f / RTC_2_PER_TICK + 1;
@@ -82,5 +112,5 @@ void hal_rtc_wakeup_timer_set_ms( const int32_t milliseconds )

 // void hal_rtc_wakeup_timer_stop( void )
 // {
-
+
 // }
```

alistair23:
> > @bradjc Maybe we should keep wifi_helpers.c in the src_changed folder for now. Changes in [these lines](https://github.com/tock/libtock-c/pull/458/commits/4ad88a703a3e4e7f3a557342a804be8ef7600e1c#diff-cf252f0bdf9d6e5d1d3d4110031fc95840d8bc8b46206319fc878ec1c5956242R161) make sure that the app does not return error when the board sometimes scans too many Wi-Fi access points.
> 
> Good point. I think I was thinking we should change `WIFI_MAX_RESULTS`, but that isn't straightforward easy, and could run into the same issue.

Can you upstream the fix?

bradjc:
I need to work on the build error of course but otherwise I think I have this PR to a state where it is reasonably mergeable.

examples/tests/lr1110/lorawan/main_lorawan.c
```diff
@@ -0,0 +1,460 @@
...
+ * --- PRIVATE CONSTANTS -------------------------------------------------------
+ */
+
+#define LORAWAN_REGION      SMTC_MODEM_REGION_US_915
+#define LORAWAN_CLASS       SMTC_MODEM_CLASS_A
+
+// wm1110dev parameters
+#define LORAWAN_DEVICE_EUI "70B3D57ED00650D9"
+#define LORAWAN_JOIN_EUI   "901AB1F40E1BCC81"
+#define LORAWAN_APP_KEY    "3356A7047ECF1F2F78C72AE9B1635BC1"
```
alistair23, line 82: These are secrets, you will need to allow users to set then and reset yours.

See https://github.com/tock/libtock-c/pull/456 for how this should be handled

examples/tests/lr1110/lorawan/main_lorawan.c
```diff
@@ -0,0 +1,460 @@
...
+  } else {
+    printf("Humidity sensor not available.\n");
+  }
+
+  app_data_buffer[app_data_size++] = (uint8_t)((temp >> 8) & 0xFF);
+  app_data_buffer[app_data_size++] = (uint8_t)(temp & 0xFF);
+  app_data_buffer[app_data_size++] = (uint8_t)((humi >> 8) & 0xFF);
+  app_data_buffer[app_data_size++] = (uint8_t)(humi & 0xFF);
+
+  send_frame(app_data_buffer, app_data_size, LORAWAN_CONFIRMED_MSG_ON);
```
alistair23, line 358: This is how I started out as well, the problem is TTN can't easily decode it.

You should look at using the Cayenne Low Power Payload (LPP) encoder so then TTN will display the actual results

examples/tests/lr1110/wifi_scan/README.md
```diff
@@ -0,0 +1,179 @@
...
+Information messages are displayed on the serial console, starting with the
+DevEUI, AppEUI/JoinEUI and PIN that you might need in order to register your
+device with the LoRa Cloud Device Join service.
+
+### LoRaWAN Network Server / Application Server
+
+This application needs an Application Server to run in order to perform the Wi-Fi
+solving.
+Two possibilities :
+* A Node-Red application server is provided in folder *apps/geolocation_application_server*.
```
alistair23, line 106: Is it?

examples/tests/lr1110/wifi_scan/main_geolocation_wifi.c
```diff
@@ -0,0 +1,512 @@
...
+ */
+
+
+#define LORAWAN_REGION      SMTC_MODEM_REGION_US_915
+#define LORAWAN_CLASS       SMTC_MODEM_CLASS_A
+
+// wm1110dev parameters
+#define LORAWAN_DEVICE_EUI  "70B3D57ED00650D9"
+#define LORAWAN_JOIN_EUI    "901AB1F40E1BCC81"
+#define LORAWAN_APP_KEY     "3356A7047ECF1F2F78C72AE9B1635BC1"
```
alistair23, line 91: Same here

examples/tests/lr1110/wifi_scan/main_geolocation_wifi.c
```diff
@@ -0,0 +1,512 @@
...
+ */
+
+static void on_modem_reset(uint16_t reset_count) {
+  printf("on_modem_reset %i\n", reset_count);
+
+  /* Basic LoRaWAN configuration */
+  wifi_apps_modem_common_configure_lorawan_params( );
+
+  // We use TTN so only enable bank 2 channels.
+  region_us_915_the_things_network_init();
```
alistair23, line 306: What if a user is in a different region?

examples/tests/lr1110/wifi_scan/main_geolocation_wifi.c
```diff
@@ -0,0 +1,512 @@
...
+
+  // ASSERT_SMTC_MODEM_RC( smtc_modem_alarm_start_timer( WIFI_SCAN_PERIOD ) );
+  // printf( "smtc_modem_alarm_start_timer: %d s\n\n", WIFI_SCAN_PERIOD );
+
+  // mw_return_code_t wifi_rc;
+  // wifi_rc = wifi_mw_scan_start( 5 );
+  // if( wifi_rc != MW_RC_OK )
+  // {
+  //     HAL_DBG_TRACE_ERROR( "Failed to start WiFi scan\n" );
+  // }
```
alistair23, line 353: Still some commented out code that should be removed

lr1110/Makefile
```diff
@@ -0,0 +1,139 @@
...
+CHANGED_CSRCS := $(wildcard $(LR1110_DIR)/src_changed/*.c)
+
+# include changed headers
+override CFLAGS += -I$(LR1110_DIR)/inc_changed
+
+# -DREGION_US_915: Defines the LoRaWAN region to US 915 MHz band
+# -DRP2_103: Specifies the use of Regional Parameters version RP2-1.0.3
+# -DTASK_EXTENDED_2: The LoRa Basics Modem extended uplink ID to be used for Wi-Fi uplinks
+# -DLR11XX, -DLR11XX_TRANSCEIVER: Let the code select methods for LR11XX radio board
+override CFLAGS += -DREGION_US_915 -DRP2_103 -DTASK_EXTENDED_2 -DLR11XX -DLR11XX_TRANSCEIVER
```
alistair23, line 25: What if this doesn't match what's in the `main*.c` macros?

lr1110/Makefile.app
```diff
@@ -0,0 +1,52 @@
+LR1110_DIR = $(TOCK_USERLAND_BASE_DIR)/lr1110/lr1110
+SEEED_DIR = $(LR1110_DIR)/seeed
+
+# include changed headers and parameters
+override CFLAGS += -I$(LR1110_DIR)/inc_changed -DREGION_US_915 -DRP2_103 -DTASK_EXTENDED_2 -DLR11XX -DLR11XX_TRANSCEIVER
```
alistair23, line 5: Same here. I think the region setting needs a better solution

lr1110/lr1110/us_915_ttn.c
```diff
@@ -0,0 +1,32 @@
+#include "lr1_stack_mac_layer.h"
+#include "lr1mac_utilities.h"
+#include "lorawan_api.h"
```
alistair23, line 3: Again, what about other regions?

lr1110/lr1110/src_changed/smtc_hal_mcu.c
```diff
@@ -0,0 +1,68 @@
...
+}
+
+void __attribute__( ( optimize( "O0" ) ) ) hal_mcu_wait_us( const int32_t microseconds )
+{
+    // Work @64MHz
+    uint32_t dummy = 0;
+    const uint32_t nb_nop = microseconds * 1000 / 171;
+    for( uint32_t i = 0; i < nb_nop; i++ )
+    {
+        //__NOP( );
```
alistair23, line 38: ?

lr1110/wifi_helpers.patch
```diff
@@ -0,0 +1,13 @@
+--- wifi_helpers.c	2024-08-01 15:07:38
++++ wifi_helpers.c	2024-08-01 15:08:24
+@@ -158,8 +158,8 @@
+     max_nb_results = sizeof( wifi_results_mac_addr ) / sizeof( wifi_results_mac_addr[0] );
+     if( nb_results > max_nb_results )
+     {
+-        MW_DBG_TRACE_ERROR( "Wi-Fi scan result size exceeds %u (%u)\n", max_nb_results, nb_results );
+-        return false;
++        // Just use the number we received that can fit in our results array.
++        nb_results = max_nb_results;
```
alistair23, line 10: Why change this?

lr1110/radio_planner.patch
```diff
@@ -0,0 +1,41 @@
...
++        // For some reason we get a spurious RX_TIMEOUT interrupt sometimes.
++        // Handling it causes the stack to break, but clearing it causes the
++        // stack to work.
++        // re-get the irq status if RX_TIMEOUT happens (current workaround)
++        if( rp_task_type == RP_TASK_TYPE_WIFI_RSSI) {
++            if(rp->status[rp->radio_task_id] == RP_STATUS_RX_TIMEOUT) {
++                rp_irq_get_status( rp, rp->radio_task_id );
++                // return;
++            }
++        }
```
alistair23, line 38: This should be upstreamed, not just hacked around

alistair23 (changes requested):

examples/tests/lr1110/lorawan/main_lorawan.c
```diff
@@ -0,0 +1,460 @@
...
+
+// wm1110dev parameters
+#define LORAWAN_DEVICE_EUI "70B3D57ED00650D9"
+#define LORAWAN_JOIN_EUI   "901AB1F40E1BCC81"
+#define LORAWAN_APP_KEY    "3356A7047ECF1F2F78C72AE9B1635BC1"
+
+/*!
+ * @brief Defines the application data transmission duty cycle. 60s (changed to 3s), value in [s].
+ */
+#define APP_TX_DUTYCYCLE 3
```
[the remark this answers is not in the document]
ethanzhouyc: I remembered it is how often the app makes an attempt to transmit data to TTN, and if the transmission succeeds, the duty cycle will stop. I agree that we should make it a larger number, but if it's too large, like 60, it might fail the app.

examples/tests/lr1110/lorawan/main_lorawan.c
ethanzhouyc: I wrote local scripts to decode data from TTN, and LLP encoder sounds much more convenient

examples/tests/lr1110/wifi_scan/README.md
ethanzhouyc: No. This file is copied from [seeed's repo](https://github.com/ethanzhouyc/Seeed_Wio_WM1110_Dev_Board/blob/master/apps/examples/geolocation_wifi/README.md) just as other files, and we should write our own version

ethanzhouyc: I suggest deleting this file and add a link to the original README on top of the file

lr1110/Makefile.app
ethanzhouyc: Ideally we should prompt users to input their region code, but we discussed to agree that this approach will be sufficient for the current stage.

lr1110/lr1110/us_915_ttn.c
ethanzhouyc: It will be hard to test in other regions too

lr1110/wifi_helpers.patch
ethanzhouyc: The app returns an error if more than 5 results are detected, which sometimes occurs when there are many Wi-Fi access points nearby. This update retains all original functionality while preventing potential errors.

lr1110/radio_planner.patch
ethanzhouyc: It's the only workaround we could find. We spent lots of time debugging this and it's hard to resolve

lr1110/Makefile.app
alistair23: Who discussed?

This isn't sufficient as you don't actually allow people to change it. It's set here and in the code and completely undocumented

lr1110/wifi_helpers.patch
alistair23: That should go upstrem then

lr1110/radio_planner.patch
alistair23: Is there an issue opened against the original library?

examples/tests/lr1110/wifi_scan/README.md
alistair23: Sounds good

examples/tests/lr1110/lorawan/main_lorawan.c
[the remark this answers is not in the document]
alistair23: It's way too low then, this needs to be at least 60 minutes

examples/tests/lr1110/lorawan/main_lorawan.c
```diff
@@ -81,9 +83,10 @@
 #define LORAWAN_CLASS       SMTC_MODEM_CLASS_A
 
 // wm1110dev parameters
-#define LORAWAN_DEVICE_EUI "70B3D57ED00650D9"
-#define LORAWAN_JOIN_EUI   "901AB1F40E1BCC81"
-#define LORAWAN_APP_KEY    "3356A7047ECF1F2F78C72AE9B1635BC1"
```
alistair23: Just an FYI, the key is still here. So make sure you revoke it from TTN

examples/tests/lr1110/lorawan/main_lorawan.c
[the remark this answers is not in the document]
ethanzhouyc: Since this is only for a test app, the impact should be minimal here.

lr1110/wifi_helpers.patch
ethanzhouyc: Seeed Studio is not going to upstream this. It is caused by integrating their code with Tock OS.

lr1110/radio_planner.patch
ethanzhouyc: No, their code will work fine on their OS stack, but we are integrating their code into Tock OS here.

lr1110/Makefile.app
ethanzhouyc: We will think about that a bit more

examples/tests/lr1110/lorawan/main_lorawan.c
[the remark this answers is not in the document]
alistair23: It's not minimal impact, people will be banned from TTN if they run the test app and get caught.

Not to mention it's possible a crime in some countries

lr1110/wifi_helpers.patch
alistair23: Then that's a bug that should be fixed, not just hacked around

bradjc:
Merged in #490

PR #464, opened 2024-08-20
configuration: add target for VeeR EL2

wsipak:
This allows building Tock applications for VeeR EL2  (see related PR: https://github.com/tock/tock/pull/4118)

### DIFF
--- a/Configuration.mk
+++ b/Configuration.mk
@@ -66,6 +66,8 @@ OPENTITAN_TOCK_TARGETS := rv32imc|rv32imc.0x20030080.0x10005000|0x20030080|0x100
 ARTY_E21_TOCK_TARGETS := rv32imac|rv32imac.0x40430080.0x80004000|0x40430080|0x80004000\
                          rv32imac|rv32imac.0x40440080.0x80007000|0x40440080|0x80007000
 
+VEER_EL2_TOCK_TARGETS := rv32imc|rv32imc.0x20300080.0x20602000|0x20300080|0x20602000
+
 # Include the RISC-V targets.
 #  rv32imac|rv32imac.0x20040080.0x80002800 # RISC-V for HiFive1b
 #  rv32imac|rv32imac.0x403B0080.0x3FCC0000 # RISC-V for ESP32-C3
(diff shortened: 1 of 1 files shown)


lschuermann (approved):

alistair23 (approved):

ppannuto (dismissed):
In the abstract this looks good, but I assume that build error is real?

wsipak:
From what I see here: https://github.com/tock/libtock-c/blob/master/AppMakefile.mk#L273
The base address I've used (`0x80030080`) is problematic. The generated linker script (`*userland_debug.ld`) has a negative value specified for the origin of FLASH:
```
FLASH (rx) : ORIGIN = -87, LENGTH = PROG_LENGTH
```

I am able to change the base address, so I'll just change the linkerscript in Tock and the addresses in libtock-c, then respond again here.

lschuermann:
> Oh I see, this is because we have a long standing assumption that using address 0x80000000 as a fake flash address (for PIC apps) won't conflict with anything. But if app flash does start nearish 0x80000000 then this error happens.

That seems ... problematic. Any immediate / obvious solutions, or is this a good thing to discuss synchronously on the call Friday?

ppannuto:
It's not problematic for ARM, which defines the suites of addresses which can be used for memory-like things. 
- (That said `0x80... — 0x9f...` is reserved for SRAM-like structures, "Memory with Write-Through cache attribute." — we probably should change the sentinel value to be something more definitely not a memory, e.g. `0xE0000000-`, which is the start of the PPB space, and holds ARM-defined [and immutable address] system configuration registers)

Is there any similar address region defined by the RISC-V spec, which could never for any valid core have memory?

wsipak:
I've adjusted the memory map in the PR to Tock: https://github.com/tock/tock/pull/4118
And also created a related PR to tockloader: https://github.com/tock/tockloader/pull/117

I consider these changes in-sync and ready.

I'm not sure if the CI here failed because of the changes actually, from the log:
```
2024-08-28T14:00:13.8161250Z 218350K .......... .......... .......... .......... .........  WARNING: Fetching newlib from mirror https://alpha.mirror.svc.schuermann.io/files/tock failed!
2024-08-28T14:00:13.8162450Z . 74% 92.0M 1s
2024-08-28T14:00:13.8163036Z ERROR: Unable to find tock-newlib
```

Could you please restart the CI or explain how I should handle this?

ppannuto:
(n.b., re-running CI job now assuming that was a transient server error)

ppannuto (approved):

lschuermann:
> @lschuermann ^ Looks like something's broken with the newlib files?

Yeah, it's weird that it fails -- it should try both UVA and my server as a mirror, and we can add more if necessary. @wsipak I can't find the logs of this failed workflow but I'd be curious as to whether it tried to fetch something from `cs.virgina.edu` first and that also failed. In this case, it's more likely to be a local (transient) network issue at the CI host.

lschuermann (approved):

PR #470, opened 2024-10-23
 Support the latest newlib, libc++ and GCC (replaces #442)

ppannuto:
#442 has an unfortunate chicken-and-egg problem wherein a commit new enough to checkout by hash in the dockerfile will only exist in the main libtock-c repo if the pull request comes from a branch within the libtock-c repo.

Hence this PR, which is based on the external branch, but is now coming from an internal branch. This isn't the most ergonomic or friendly solution, but this is also a rare enough issue [hopefully] that a little occasional human labor seems the path of least resistance.

I added commits to update the repo path, remove the unneeded patch file, and update the hash to check out.

---

~~TODO: Technically, the Makefile change to only patch if there's a patchfile is untested as newlib is still building on my laptop, but I doubt there will be any issue~~ Hubris will get you every time... https://github.com/tock/libtock-c/pull/470/commits/ceda51b15c125599ee90bb61ce61ee97e0103cfa

### DIFF
--- a/newlib/Makefile
+++ b/newlib/Makefile
@@ -50,8 +50,8 @@ rebuild-newlib: newlib-$(NEWLIB_VERSION)
 	@mkdir -p libtock-newlib-$(NEWLIB_VERSION)/riscv
 	@cp -r newlib-arm-$(NEWLIB_VERSION)-install/arm-none-eabi libtock-newlib-$(NEWLIB_VERSION)/arm
 	@cp -r newlib-riscv-$(NEWLIB_VERSION)-install/$(TOOLCHAIN_rv32i) libtock-newlib-$(NEWLIB_VERSION)/riscv
-	@cd libtock-newlib-$(NEWLIB_VERSION)/arm/arm-none-eabi; patch -p1 < ../../../newlib-$(NEWLIB_VERSION).patch
-	@cd libtock-newlib-$(NEWLIB_VERSION)/riscv/$(TOOLCHAIN_rv32i); patch -p1 < ../../../newlib-$(NEWLIB_VERSION).patch
+	@cd libtock-newlib-$(NEWLIB_VERSION)/arm/arm-none-eabi; [ -f ../../../newlib-$(NEWLIB_VERSION).patch ] && patch -p1 < ../../../newlib-$(NEWLIB_VERSION).patch || true
+	@cd libtock-newlib-$(NEWLIB_VERSION)/riscv/$(TOOLCHAIN_rv32i); [ -f ../../../newlib-$(NEWLIB_VERSION).patch ] && patch -p1 < ../../../newlib-$(NEWLIB_VERSION).patch || true
 	@mv newlib-arm-$(NEWLIB_VERSION)-out/build-arm.log newlib-riscv-$(NEWLIB_VERSION)-out/build-riscv.log libtock-newlib-$(NEWLIB_VERSION)
 	@zip -r libtock-newlib-$(NEWLIB_VERSION).zip libtock-newlib-$(NEWLIB_VERSION)
 	@echo ""
--- a/Configuration.mk
+++ b/Configuration.mk
@@ -256,8 +256,10 @@ else ifeq ($(CC_rv32_version_major),12)
   NEWLIB_VERSION_rv32 := 4.3.0.20230120
 else ifeq ($(CC_rv32_version_major),13)
   NEWLIB_VERSION_rv32 := 4.3.0.20230120
+else ifeq ($(CC_rv32_version_major),14)
+  NEWLIB_VERSION_rv32 := 4.4.0.20231231
 else
-  NEWLIB_VERSION_rv32 := 4.3.0.20230120
+  NEWLIB_VERSION_rv32 := 4.4.0.20231231
 endif
 NEWLIB_VERSION_rv32i    := $(NEWLIB_VERSION_rv32)
 NEWLIB_VERSION_rv32imc  := $(NEWLIB_VERSION_rv32)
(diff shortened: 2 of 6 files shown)


alistair23 (approved):

alevy (dismissed):

alevy:
@ppannuto (and @bradjc & @lschuermann) does this require updating our binary repositories of pre-compiled libraries?

lschuermann:
@alevy Yes, I guess so. @bradjc I have never built these toolchains, do you have this infrastructure already set up? We can then add these new artifacts to the [mirrorcheck DB](https://github.com/tock/mirrorcheck/blob/main/urls.json) which will ensure that we have this replicated everywhere.

ppannuto:
@lschuermann re build infra: that's what the dockerization is for; it should just be a matter of running each of the new `docker-create.sh`'s.

That said, I'm building the toolchains currently and can send you built artifacts for the mirror when they finish.

alevy (dismissed):

libc++/Makefile
bradjc, line 1: Why these changes?

newlib/Makefile
bradjc, line 1: Same here for these `@` changes. Patch changes make sense.

bradjc (changes requested):
Assuming it is supposed to be `libtock-libc++-14.1.0.zip` I have added these to the uva mirror.

newlib/Makefile
ppannuto: The reason is in the title of the commit: ` show execution of commands that can fail `

When running the docker script, it was opaque which command failed if the build failed without this. (i.e., I had to make this change to debug why a build failed).

bradjc: That makes sense for debugging but now that it is debugged why would this fail? If this is failing then our docker approach doesn't work.

ppannuto: I mean, I agree, they _shouldn't_ fail [and don't currently], but if, e.g., something changes when we add the next newlib version I'd rather not build everything twice just to turn on a few prints.

More generally, it aligns with the rest of the build system, i.e., print commands that do something [noting that for the more common / frequent builds with giant commands, we by default _shorten_ (but do not hide) the prints for things like `CC foo.c`].

bradjc:
I guess we can vote, but I feel that these are not commands to build our project but are frozen steps to reproduce a dependency. The changes are also not related to the PR title.

lschuermann (approved):
I'm indifferent w.r.t. the `@` changes. I agree they maybe shouldn't be part of this PR, but they are in a separate commit, and I don't have strong feelings about the actual changes either way.

Rest looks good.

alevy:
I'm similarly agnostic regarding the `@`, though am sympathetic to @bradjc's kvetching over mixing this into a basically unrelated PR. I won't block either way, but I suggest @ppannuto just remove that commit (and maybe make it `VERBOSE=1` option in a different PR?) in the name of peaceable coexistance and harmony

ppannuto:
Updated. Now with 100% fewer controversial `@` changes.

lschuermann (approved):

alevy (approved):

ppannuto:
@alevy Why did this need a rebase? It should have merged cleanly, no?

The last commit ( https://github.com/tock/libtock-c/pull/470/commits/54b93c4018dfca66ed80efff4306d54bf4d6e5c1 ) needs to update to point to 2e6baf26e9d8cd4f819baf1b0fab4e9c5e3e742e now...

alevy:
> @alevy Why did this need a rebase? It should have merged cleanly, no?

Idk, but it wouldn't have merged cleanly, GitHub wouldn't let me hit the merge button.

alevy:
@bradjc I think your comments have been addressed. Re-review (or just dismiss your old review)?

ppannuto:
image
I do think the new github UX is confusing, but I think the "out of date" is just a warning that clicking Merge will create a merge commit [while rebase would allow just fast forward]. I don't think you _have_ to rebase—at least I didn't have to last week when I merged one that looked similar.

alevy:
Ahhh.... ok

ppannuto:
Undid the rebase with force push from my laptop still at old head. Should be good to go.

alevy (approved):

alevy:
@bradjc ?

lschuermann (approved):

lschuermann:
Let's give this one more day for `last-call`.

alistair23:
This seems to break builds

```
Fetching newlib from https://www.cs.virginia.edu/~bjc8c/archive/tock...
  Fetching https://www.cs.virginia.edu/~bjc8c/archive/tock/libtock-newlib-4.4.0.20231231.zip...
--2024-11-12 12:58:09--  https://www.cs.virginia.edu/~bjc8c/archive/tock/libtock-newlib-4.4.0.20231231.zip
Loaded CA certificate '/etc/ssl/certs/ca-certificates.crt'
Resolving www.cs.virginia.edu (www.cs.virginia.edu)... 128.143.67.8
Connecting to www.cs.virginia.edu (www.cs.virginia.edu)|128.143.67.8|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 355123001 (339M) [application/zip]
Saving to: 'libtock-newlib-4.4.0.20231231.zip'

libtock-newlib-4.4.0.20231231.zip      100%[============================================================================>] 338.67M  11.6MB/s    in 35s     

2024-11-12 12:58:45 (9.68 MB/s) - 'libtock-newlib-4.4.0.20231231.zip' saved [355123001/355123001]

sha256sum: 'standard input': no properly formatted checksum lines found
  WARNING: Fetching newlib from mirror https://www.cs.virginia.edu/~bjc8c/archive/tock failed!
```

ppannuto:
https://github.com/tock/libtock-c/pull/476

PR #474, opened 2024-11-08
15.4: Fix CFG buffer length allowed to kernel (tx)

tyler-potyondy:
In the libtock sync/async rewrite, the CFG buffer length allowed to the kernel was mistakenly changed from 11 to 27. This reverts the mistake.

This resolves https://github.com/tock/tock/issues/4227.

### DIFF
--- a/libtock/net/ieee802154.c
+++ b/libtock/net/ieee802154.c
@@ -351,7 +351,7 @@ returncode_t libtock_ieee802154_send(uint32_t                              addr,
   }
 
   // Allow CFG buffer to the kernel
-  returncode_t ret = libtock_ieee802154_set_readwrite_allow_cfg((void*) BUF_CFG, 27);
+  returncode_t ret = libtock_ieee802154_set_readwrite_allow_cfg((void*) BUF_CFG, 11);
   if (ret != RETURNCODE_SUCCESS) return ret;
 
   // Allow payload buffer to the kernel


libtock/net/ieee802154.c
```diff
@@ -351,7 +351,7 @@ returncode_t libtock_ieee802154_send(uint32_t                              addr,
   }
 
   // Allow CFG buffer to the kernel
-  returncode_t ret = libtock_ieee802154_set_readwrite_allow_cfg((void*) BUF_CFG, 27);
+  returncode_t ret = libtock_ieee802154_set_readwrite_allow_cfg((void*) BUF_CFG, 11);
```
ppannuto, line 354: I think line 11 (`BUF_CFG` declaration) needs to change too?

Should the value `11` be changed to a `BUF_CFG_LEN` macro or similar to help with consistency? (And the macro definition can have comment / explainer for the length if there is something sensible to say?)

ppannuto (dismissed):

libtock/net/ieee802154.c
tyler-potyondy: > I think line 11 (BUF_CFG declaration) needs to change too?
This is confusing and an artifact of how it has always been :)

The BUF_CFG given to the kernel is length 10, 11, 16, or 27 (depending on what it is used for). BUF_CFG is used for passing keys, getting keys, and also mac address info.

tyler-potyondy: I think this current approach is not good and easily can lead to bugs when trying to make changes (as seen here). 

Perhaps updating the length field to all be constants would help with this?

ppannuto: Oof, yeah, maybe something like 

    #define CFG_GETKEY_LEN 10
    #define CFG_SETKEY_LEN 11
    #define CFG_LEN_MAX MAX(CFG_GETKEY_LEN, CFG_SETKEY_LEN, ...)
    uint8_t BUF_CFG[CFG_LEN_MAX];

it's a bit wordy, but hopefully a lot clearer?

alevy: Does it need to be the precise length or can it be larger than the kernel needs?

Presumably, `BUF_CFG` is only ever used for one thing at a time (if not, it should probably be several different buffers, no?). Can we always just allow its max length to the kernel? (`sizeof(BUF_CFG)`) then it's just a matter of sizing it to be as large as the maximum needed.

Alternatively, might be an appropriate place for dynamic allocation :)

alevy (approved):
I'm inclined to accept these without requiring fixing the more general human source of the bug, though I agree the fix is as brittle as the original.

lschuermann (approved):

PR #475, opened 2024-11-12
examples: sensors: Support moisture readings

alistair23:
Support the new moisture sensor and add a environment sensor example application

### DIFF
--- a/examples/sensors/main.c
+++ b/examples/sensors/main.c
@@ -3,6 +3,7 @@
 
 #include <libtock-sync/sensors/ambient_light.h>
 #include <libtock-sync/sensors/humidity.h>
+#include <libtock-sync/sensors/moisture.h>
 #include <libtock-sync/sensors/ninedof.h>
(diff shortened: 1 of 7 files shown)


ppannuto (changes requested):
Looks good to me, only blocking comment is the 'example output'.

---

Any reason to have a new app here instead of adding to the existing sensors app ( https://github.com/tock/libtock-c/tree/master/examples/sensors ) ?

alistair23:
> Any reason to have a new app here instead of adding to the existing sensors app ( https://github.com/tock/libtock-c/tree/master/examples/sensors ) ?

Nope! I just didn't know that existed. I've switched to using that instead

ppannuto (dismissed):
lgtm; the 'example output' for sensors is a bit awkward, as we don't have any board with _all_ of the sensors I don't think, so I think leaving that unchanged here is fine

ppannuto (approved):

PR #477, opened 2024-11-14
lib/fetch-{libc++,newlib}: exit on error

lschuermann:
This helps diagnose issues when one or more of the tools that the script requires are not installed.

In my case, I was working on a system that did not have `unzip` installed. This error message was hidden in between a lot of other log messages, and the error was masked by a generic "can't find this header" message.

This also removes the `let` bindings. The bash `let` commands returns exit code 1 when assigning the value `0` to a variable, which is ... 100% what you want, just great behavior, and definitely did not cost me another 10 minutes to debug.


lschuermann:
Seems like we were relying on failures being silently ignored in CI:

```
2024-11-14T20:33:32.5126662Z 293100K .......... .......... ...                             100%  195M=8.0s
2024-11-14T20:33:32.5127237Z 
2024-11-14T20:33:32.5128455Z 2024-11-14 20:33:32 (35.9 MB/s) - ‘libtock-newlib-4.2.0.20211231.zip’ saved [300158324/300158324]
2024-11-14T20:33:32.5129275Z 
2024-11-14T20:33:33.2008185Z libtock-newlib-4.2.0.20211231.zip: OK
2024-11-14T20:33:33.2016379Z Unpacking libtock-newlib-4.2.0.20211231.zip...
2024-11-14T20:33:33.7019460Z libtock-newlib-4.2.0.20211231.zip: OK
2024-11-14T20:33:33.7028612Z Unpacking libtock-newlib-4.2.0.20211231.zip...
2024-11-14T20:33:33.7049070Z replace libtock-newlib-4.2.0.20211231/arm/arm-none-eabi/include/argz.h? [y]es, [n]o, [A]ll, [N]one, [r]ename:  NULL
2024-11-14T20:33:33.7050216Z (EOF or read error, treating as "[N]one" ...)
2024-11-14T20:33:33.7196157Z replace libtock-newlib-4.2.0.20211231/arm/arm-none-eabi/lib/thumb/v7ve+simd/softfp/redboot.specs? [y]es, [n]o, [A]ll, [N]one, [r]ename:  NULL
2024-11-14T20:33:33.7197691Z (EOF or read error, treating as "[N]one" ...)
2024-11-14T20:33:37.7497198Z make[1]: *** [../Precompiled.mk:61: ../lib/libtock-newlib-4.2.0.20211231/arm/arm-none-eabi/lib/thumb/v6-m/nofp/libc.a] Error 1
2024-11-14T20:33:37.7498747Z make[1]: Leaving directory '/home/runner/work/libtock-c/libtock-c/lvgl'
2024-11-14T20:33:37.7502329Z make: *** [../../AppMakefile.mk:97: ../../lvgl/build/cortex-m0/lvgl.a] Error 2
```

I'll push a commit that changes the invocation of `unzip` to select the (currently inferred) default of not overwriting existing files.

ppannuto (approved):
I don't think these fetch scripts should be getting called in a way that causes this to be an issue, so I have a mild fear this is papering over something else, but that is an issue for another day.

ppannuto:
Oof. Or... it's not an error for another day. Looks like the CI error is competing unzip's trampling each other, likely because the first unzip was only partially complete when the second one first checked if it needed to do anything :/

lschuermann:
@ppannuto Can you perhaps validate that this still works on macOS?

ppannuto:
Out of the box experience unfortunately is a lot of this:
```
./fetch-newlib.sh: line 44: flock: command not found
Could not acquire non-blocking lock on libtock-newlib-4.4.0.20231231.zip, waiting for lock to be released...
```

I feel like the underlying problem here is really an error in the make rules somewhere, make should only be trying to satisfy the `$(TOCK_NEWLIB_TARGETS)` once, will try to dig into this...

ppannuto (approved):
Thanks for battling the bash arcana! This worked well on Mac locally too stressing it a few different ways.

Someday we should probably de-dup these scripts; they kinda do the same thing.. but that day is not today!

ppannuto:
@bradjc I know you were testing this some too — will leave for you to merge

PR #484, opened 2025-01-02
kv: Support garbage collection from userspace

alistair23:


examples/tests/kv/kv_unit_tests/Makefile
```diff
@@ -6,7 +6,7 @@ TOCK_USERLAND_BASE_DIR = ../../../..
 # Which files to compile.
 C_SRCS := $(wildcard *.c)
 
-ELF2TAB_ARGS += --write_id 17767 --read_ids 17767 --access_ids 17767
+ELF2TAB_ARGS += --write_id 1153320202 --read_ids 1153320202 --access_ids 1153320202
```
lschuermann, line 9: Just curious, why these changes? If there's meaning in the magic numbers, might be worth documenting?

lschuermann (approved):

examples/tests/kv/kv_unit_tests/Makefile
alistair23: As Tock no longer uses `write_id`s these numbers need to match the actual short ID generated in the kernel. Which generally is a SHA of the name. So this is updating these to match the numbers the kernel uses.

It's a pain to work with and very confusing unfortunately

lschuermann: Gotcha. Mind adding a comment for this? I agree that this is confusing, esp. to someone uninitiated to the whole AppId subsystem, so I think a comment might help a lot!

lschuermann (approved):

alistair23:
Ping!

alistair23:
Ping!

alistair23:
Ping! This has been approved

alistair23:
Ping! @lschuermann

bradjc (approved):

PR #487, opened 2025-01-17
RadioLib/RadioLib: Bump to latest commit

alistair23:


ppannuto (approved):
This is a small, contained enough corner that it's not worth delaying over, but in the future if you can toss a sentence or bullet or two or something in the PR body of what feature/bug/whatever is motivating the update that'd be appreciated.

PR #490, opened 2025-01-29
Add LR1110 library for WM1110 LoRa radio

bradjc:
This is a followup to #458 with the commits cleaned up.

We've verified that with https://github.com/tock/tock/pull/4320 this is working and we can join TTN with the board and send sensor readings.

We have three patches to core files in the library that we found we needed when trying to get this working. We don't have the bandwidth to investigate those further right now, and while there has been some activity to the upstream repo, there haven't been any source changes in two years.

This isn't a perfect port, but it is self contained and we want to build on this.


brghena (approved):
This looks good to me. Lots of high-quality, self-contained stuff

bradjc:
> Can you clarify the relationship between this and #458? Does this PR replace that PR? (It looks to me like it does)

Yes this replaces it with a cleaned up commit history.

bradjc:
Ok the CI passes!

brghena (approved):

PR #492, opened 2025-02-16
libtock: add util/streaming_process_slice

lschuermann:
This adds userspace helpers implementing the "streaming process slice" contract as implemented in the Tock kernel in [1]. Instead of working with two separate buffers, these helpers present an interface similar to a regular read-write allow based on a single buffer. Internally, this buffer is then subdivided into an "application-owned" and "kernel-owned" part which are swapped atomically. The caller is responsible for keeping a `streaming_process_slice_state_t`, and is provided with ephemeral references to the kernel-written payloads on each swap operation. The full, original buffer can be reclaimed by deinitializing the `streaming_process_slice_state_t` struct.

Currently, these helpers are quite minimal and do not expose the ability to set any optional flags (such as `halt`). They are tested to work against against the interface implemented in [1] (with the fix of [2] included) as part of the LwIP Ethernet tap-driver userspace app.

[1]: https://github.com/tock/tock/pull/4208
[2]: https://github.com/tock/tock/pull/4343

### DIFF
--- a/libtock/util/streaming_process_slice.c
+++ b/libtock/util/streaming_process_slice.c
@@ -1,0 +1,144 @@
+#include "streaming_process_slice.h"
+#include <string.h>
+
+static void streaming_process_slice_prepare_header(uint8_t* buf) {
+  buf[0] = 0;  // version[H]
+  buf[1] = 0;  // version[L]
+  buf[2] = 0;  // flags[H]
+  buf[3] = 0;  // flags[L]
+  buf[4] = 0;  // write offset
+  buf[5] = 0;  // write offset
+  buf[6] = 0;  // write offset
+  buf[7] = 0;  // write offset
+}
+
+returncode_t streaming_process_slice_init(
+  streaming_process_slice_state_t* state,
+  uint32_t                         driver,
+  uint32_t                         allow,
+  void*                            buffer,
+  size_t                           size) {
+  // Each slice's header is 8 bytes long, and we create two slices from this
+  // buffer. Thus ensure that the provided buffer is at least 16 bytes long:
+  if (size < 16) {
+    return RETURNCODE_ESIZE;
+  }
+
+  state->driver = driver;
+  state->allow  = allow;
+
+  // We split the buffer in half, an application and kernel side. These two
+  // buffers are then atomically swapped with each other.
+  //
+  // Initially, the first half of this buffer is designated as the application
+  // buffer.
+  state->app_buffer_ptr  = buffer;
+  state->app_buffer_size = size / 2;
+
+  // Write a streaming process slice header to the second half of this buffer,
+  // and allow it to be the kernel buffer. We currently only support version
+  // 0, and don't set the `halt` flag:
+  uint8_t* kernel_buffer_ptr = state->app_buffer_ptr + state->app_buffer_size;
+  size_t kernel_buffer_size  = size - state->app_buffer_size;
+  streaming_process_slice_prepare_header(kernel_buffer_ptr);
+
+  allow_rw_return_t allow_res =
+    allow_readwrite(driver, allow, kernel_buffer_ptr, kernel_buffer_size);
+  if (!allow_res.success) {
+    memset(state, 0, sizeof(streaming_process_slice_state_t));
+  }
+
+  return tock_status_to_returncode(allow_res.status);
+}
+
+returncode_t streaming_process_slice_get_and_swap(
+  streaming_process_slice_state_t* state,
+  uint8_t**                        buffer,
+  uint32_t*                        size,
+  bool*                            exceeded) {
+  uint8_t* ret_buffer;
+  uint32_t ret_size;
+  bool ret_exceeded;
+
+  // Prepare the current app buffer to be shared with the kernel (writing a
+  // zeroed-out header):
+  streaming_process_slice_prepare_header(state->app_buffer_ptr);
+
+  // Swap the current app buffer for the kernel buffer:
+  allow_rw_return_t allow_res =
+    allow_readwrite(state->driver, state->allow, state->app_buffer_ptr,
+                    state->app_buffer_size);
+
+  if (allow_res.success) {
+    // Record the new app buffer:
+    state->app_buffer_ptr  = allow_res.ptr;
+    state->app_buffer_size = allow_res.size;
+
+    // Return information about the received payload:
+    ret_buffer = state->app_buffer_ptr + 8;
+    memcpy(&ret_size, state->app_buffer_ptr + 4, sizeof(uint32_t));
+    ret_exceeded = (state->app_buffer_ptr[3] & 0x01) == 0x01;
+  } else {
+    // Allow was not successful, return safe dummy values instead:
+    ret_buffer   = NULL;
+    ret_size     = 0;
+    ret_exceeded = false;
+  }
+
+  // Write return values if provided with non-NULL pointers:
+  if (buffer != NULL) {
+    *buffer = ret_buffer;
+  }
+  if (size != NULL) {
+    *size = ret_size;
+  }
+  if (exceeded != NULL) {
+    *exceeded = ret_exceeded;
+  }
+
+  return tock_status_to_returncode(allow_res.status);
+}
+
+returncode_t streaming_process_slice_deinit(
+  streaming_process_slice_state_t* state,
+  uint8_t**                        buffer,
+  size_t*                          size) {
+  uint8_t* ret_buffer;
+  size_t ret_size;
+
+  // Unallow the buffer currently allowed to the kernel:
+  allow_rw_return_t unallow_res =
+    allow_readwrite(state->driver, state->allow, NULL, 0);
+
+  if (unallow_res.success) {
+    // Unallow failed, don't modify the state struct.
+    ret_buffer = NULL;
+    ret_size   = 0;
+  } else {
+    // The unallow worked, recreate the full, initial buffer from the app and
+    // kernel halves:
+    if ((void*)state->app_buffer_ptr < unallow_res.ptr) {
+      // App buffer is left half, kernel buffer is right half:
+      // `[ app_buffer ][ kernel_buffer ]`
+      ret_buffer = state->app_buffer_ptr;
+      ret_size   = state->app_buffer_size + unallow_res.size;
+    } else {
+      // App buffer is right half, kernel buffer is left half:
+      // `[ kernel_buffer ][ app_buffer ]`
+      ret_buffer = unallow_res.ptr;
+      ret_size   = unallow_res.size + state->app_buffer_size;
+    }
+
+    // Wipe the state struct:
+    memset(state, 0, sizeof(streaming_process_slice_state_t));
+  }
+
+  if (buffer != NULL) {
+    *buffer = ret_buffer;
+  }
+  if (size != NULL) {
+    *size = ret_size;
+  }
+
+  return tock_status_to_returncode(unallow_res.status);
+}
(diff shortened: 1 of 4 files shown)


libtock/util/streaming_process_slice.h
```diff
@@ -0,0 +1,85 @@
...
+// a signal that some data has been inserted into the kernel-owned
+// buffer (such as an upcall).
+//
+// This method abstracts this interface by consuming one buffer and
+// splitting it into two halves, owned by the application and kernel
+// respectively. It tracks all necessary state in the
+// `streaming_process_slice_state_t` object. For this to work, the
+// passed buffer must be able to hold at least two streaming process
+// slice headers (8 byte each), i.e., it must be at least 16 bytes
+// long.
```
brghena, line 29: Why not have the user provide two separate buffer pointers? To me, it would make sense for an application to implement two buffers of the same size (Header + Data) and pass both into this, rather than one buffer of double size (2x Header + Data).

libtock/util/streaming_process_slice.h
```diff
@@ -0,0 +1,85 @@
...
+// specified driver and read-write allow slot, this function converts
+// this error-status into a returncode using
+// `tock_status_to_returncode` and returns it to the caller. When this
+// method returns `RETURNCODE_SUCCESS`, the passed buffer is assumed
+// to be owned by this `streaming_process_slice_state_t` and must not
+// be used until after a successful call to
+// `streaming_process_slice_deinit`. When the buffer is of
+// insufficient size, it returns `RETURNCODE_ESIZE`.
+returncode_t streaming_process_slice_init(
+  streaming_process_slice_state_t* state,
```
brghena, line 41: The expectation here is that `state` is currently uninitialized and this function will initialize it? That should be documented above.

libtock/util/streaming_process_slice.h
```diff
@@ -0,0 +1,85 @@
...
+// must not be considered valid.
+returncode_t streaming_process_slice_get_and_swap(
+  streaming_process_slice_state_t* state,
+  uint8_t**                        buffer,
+  uint32_t*                        size,
+  bool*                            exceeded);
+
+// Deinitialize an initialized `streaming_process_slice_state_t`
+//
+// This function reconstructs the passed into `streaming_process_slice_init` and
```
brghena, line 74:
```suggestion
// This function reconstructs the buffer passed into `streaming_process_slice_init` and
```

libtock/util/streaming_process_slice.c
```diff
@@ -0,0 +1,144 @@
...
+    // Return information about the received payload:
+    ret_buffer = state->app_buffer_ptr + 8;
+    memcpy(&ret_size, state->app_buffer_ptr + 4, sizeof(uint32_t));
+    ret_exceeded = (state->app_buffer_ptr[3] & 0x01) == 0x01;
+  } else {
+    // Allow was not successful, return safe dummy values instead:
+    ret_buffer   = NULL;
+    ret_size     = 0;
+    ret_exceeded = false;
+  }
```
brghena, line 86: I think you can re-arrange those variable initializations and remove an else case.

```suggestion
  // Prepare the current app buffer to be shared with the kernel (writing a
  // zeroed-out header):
  streaming_process_slice_prepare_header(state->app_buffer_ptr);

  // Swap the current app buffer for the kernel buffer:
  allow_rw_return_t allow_res =
    allow_readwrite(state->driver, state->allow, state->app_buffer_ptr,
                    state->app_buffer_size);

  // Initialize to safe dummy values in case the allow was not successful
  uint8_t* ret_buffer = NULL;
  uint32_t ret_size = 0;
  bool ret_exceeded = false;
  if (allow_res.success) {
    // Record the new app buffer:
    state->app_buffer_ptr  = allow_res.ptr;
    state->app_buffer_size = allow_res.size;

    // Return information about the received payload:
    ret_buffer = state->app_buffer_ptr + 8;
    memcpy(&ret_size, state->app_buffer_ptr + 4, sizeof(uint32_t));
    ret_exceeded = (state->app_buffer_ptr[3] & 0x01) == 0x01;
  }
```

brghena (changes requested):
This makes sense to me. Pretty reasonably clean interface, with an excellently documented implementation. A couple of questions for you.

libtock/util/streaming_process_slice.c
brghena: Hmm, that diff makes it look far more complicated than it is.

libtock/util/streaming_process_slice.h
```diff
@@ -0,0 +1,85 @@
+#include "../tock.h"
+
+typedef struct {
+  uint8_t* app_buffer_ptr;
+  size_t app_buffer_size;
+  uint32_t driver;
+  uint32_t allow;
+} streaming_process_slice_state_t;
+
```
brghena, line 9: The header could provide a #define that specifies the expected Header size, so users can rely on that for their buffer-size math.

libtock/util/streaming_process_slice.h
lschuermann: Yeah, that's also an option. I pondered over this for a while, but didn't manage to come up with any good arguments for this arguably more complex interface. There's not really a good reason for why an app would want to allow two differently sized buffers (as they're both used for the same purpose), and I don't know whether there are any significant benefits to being able to allocate the two halves separately.

I'm open to changing this, but wouldn't want to do so when there's not a good reason. I quite liked the elegance of this behaving quite similar to a regular allow, apart from the need to carry the state struct in user code.

libtock/util/streaming_process_slice.h
lschuermann: Yes, will add.

libtock/util/streaming_process_slice.c
lschuermann: No, this makes sense to me, thanks!

libtock/util/streaming_process_slice.h
brghena: My argument would be that it's currently hidden that two buffers are happening, so it's easy to incorrectly calculate how much space the buffer should have to support that. For example, needing to reserve 2*header space.

In reality, there should be library interfaces that use this and provide something better anyways. This is sort of an internal, developer interface. So it's probably not a big deal to trust them to read and understand documentation.

lschuermann: Hm, to me there's not much of a difference to telling users of this interface that they need to reserve `2*HEADER_LEN` vs. `HEADER_LEN` in the buffer, or passing two 8-byte vs. one 16-byte buffer.

In line with your argument, if we expect a single buffer, than any higher level abstraction must also take a single buffer. If we were to take two, then the higher-level layer could still do the splitting internally...

I'm still torn on this. I think the current interface is neat, and I'm yet to be convinced that there will ever be a time where we'd want to pass two non-contiguous buffers.

brghena: Here's a confusion: let's say you set `#define HEADER_LEN (16)`. Somebody could reserve 2*HEADER_LEN still. Similarly if you set `#define HEADER_LEN (8)` somebody could only reserve 1*HEADER_LEN. But again, these are really end-user arguments, and internal developers can be expected to try harder. Our documentation can be clear on which is correct, and most users will probably be a member of the core team anyways. Maybe in the long term there will be others, but they can copy from our examples.

I do NOT have a good example of why the buffers would need to be non-contiguous or different in size.

So, I'd say leave this stream open and see if anyone else agrees with one side or the other. If there aren't a lot of people on the "two buffers" side, I'd say leave it as-is.

lschuermann: I switched it to using two buffers.

lschuermann: I split this into two commits, so we can refer back to the previous version for how I implemented the splitting (it's pretty trivial, but still).

libtock/util/streaming_process_slice.c
```diff
@@ -16,34 +16,31 @@ returncode_t streaming_process_slice_init(
...
-  // Write a streaming process slice header to the second half of this buffer,
-  // and allow it to be the kernel buffer. We currently only support version
-  // 0, and don't set the `halt` flag:
-  uint8_t* kernel_buffer_ptr = state->app_buffer_ptr + state->app_buffer_size;
-  size_t kernel_buffer_size  = size - state->app_buffer_size;
-  streaming_process_slice_prepare_header(kernel_buffer_ptr);
+  // Initially, `buffer_a` is used as the application-owned buffer.
+  state->app_buffer_ptr     = buffer_a;
+  state->app_buffer_size    = size_a;
+  state->current_app_buffer = false; // false -> buffer_a
```
brghena, line 34: We could rename this to `state->app_buffer_is_b` or something like that?

libtock/util/streaming_process_slice.c
```diff
@@ -101,43 +94,55 @@ returncode_t streaming_process_slice_get_and_swap(
...
-  uint8_t**                        buffer,
-  size_t*                          size) {
-  uint8_t* ret_buffer;
-  size_t ret_size;
+  uint8_t**                        buffer_a,
+  size_t*                          size_a,
+  uint8_t**                        buffer_b,
+  size_t*                          size_b) {
+  uint8_t* ret_buffer_a, * ret_buffer_b;
+  size_t ret_size_a, ret_size_b;
```
brghena, line 102: Please don't do multiple variable initializations on a single line. Especially with pointers. It's way too easy to make a mistake.

And is there any reason to not give these default values of NULL and 0 respectively? Then we could again remove the default case in the if statement.

libtock/util/streaming_process_slice.c
```diff
@@ -101,43 +94,55 @@ returncode_t streaming_process_slice_get_and_swap(
...
     allow_readwrite(state->driver, state->allow, NULL, 0);
 
   if (unallow_res.success) {
     // Unallow failed, don't modify the state struct.
-    ret_buffer = NULL;
-    ret_size   = 0;
+    ret_buffer_a = NULL;
+    ret_size_a   = 0;
+    ret_buffer_b = NULL;
+    ret_size_b   = 0;
```
brghena, line 113: Silly comment, but is this logic right? `unallow_res.success` being `true` means the unallow failed?

brghena (changes requested):
A couple of changes for you.

I do think this is cleaner, although again if no one else cares I'm fine going with your single-buffer implementation instead.

libtock/util/streaming_process_slice.c
lschuermann: Yikes!

libtock/util/streaming_process_slice.c
lschuermann: Good idea.

libtock/util/streaming_process_slice.c
lschuermann: Makes sense. Changed to default initializers!

libtock/util/streaming_process_slice.c
lschuermann: Fixed.

libtock/util/streaming_process_slice.c
```diff
@@ -0,0 +1,144 @@
...
+      ret_buffer_a = state->app_buffer_ptr;
+      ret_size_a   = state->app_buffer_size;
+      ret_buffer_b = unallow_res.ptr;
+      ret_size_b   = unallow_res.size;
+    } else {
+      ret_buffer_a = unallow_res.ptr;
+      ret_size_a   = unallow_res.size;
+      ret_buffer_b = state->app_buffer_ptr;
+      ret_size_b   = state->app_buffer_size;
+    }
```
brghena, line 124: I know this is really bikeshedding, but could you flip the logic here? I think the code makes a lot more sense that way.

brghena (dismissed):

brghena (approved):

bradjc (approved):

alexandruradovici (approved):

PR #493, opened 2025-02-16
Configuration.mk: add TOCK_TARGETS for QEMU rv32 virt board

lschuermann:

PR #494, opened 2025-02-16
Add LwIP Ethernet example using tap network driver

lschuermann:
This adds an example application running an LwIP-based TCP/IP network
stack and HTTP server based on the `EthernetTapDriver` kernel capsule,
and the streaming process slice abstractions.

It includes commits from the following other PRs, which I will remove
once they have been merged:

- #492: **libtock: add util/streaming_process_slice**
- ~#493: **Configuration.mk: add TOCK_TARGETS for QEMU rv32 virt board**~


lschuermann:
@bradjc @ppannuto I couldn't figure out how to exclude the `lwip/lwip` submodule from the formatting pass. I assume I'm missing something obvious, as it does work for all the other submodules...

```
Formatting ./lwip_ethernet_tap
../../lwip/Makefile:14: ../../lwip/lwip/src/Filelists.mk: No such file or directory
make: *** No rule to make target '../../lwip/lwip/src/Filelists.mk'.  Stop.
 ⤤ Failure formatting ./lwip_ethernet_tap
Rebuilding Verbose: ./lwip_ethernet_tap
```

examples/lwip_ethernet_tap/main.c
```diff
@@ -0,0 +1,438 @@
...
+
+// ########## Global buffers
+
+// Buffer to transmit frames from / to the kernel:
+static uint8_t tx_frame_buffer[ETHERNET_MTU];
+
+// This buffer is used for the RX_FRAMES streaming process slice.
+#define RX_FRAME_BUFFER_LEN \
+  (ETHERNET_MTU + TAP_DRIVER_FRAME_HEADER_LEN) * RX_FRAME_BUFFER_FRAMES
+static uint8_t rx_frames_streaming_buffer[RX_FRAME_BUFFER_LEN];
```
brghena, line 63: Does this account for the extra 8 bytes *2 that the StreamingProcessSlice needs? Otherwise this holds one packet less than you were expecting, right? Actually, two less, I think, since we split it in half.

examples/lwip_ethernet_tap/main.c
```diff
@@ -0,0 +1,438 @@
...
+}
+
+static void timeouts_alarm_callback(__attribute__((unused)) uint32_t now,
+                                    __attribute__((unused)) uint32_t scheduled,
+                                    void*                            opaque) {
+  bool* alarm_fired = opaque;
+  *alarm_fired = true;
+}
+
+int main(void) {
```
alistair23, line 346: Cool!

Can you add a README describing what this does and what is expected?

examples/lwip_ethernet_tap/main.c
lschuermann: That's correct, and it does not! Great point, will fix and add the `#define` as you recommended in #492.

bradjc:
@lschuermann https://github.com/tock/libtock-c/blob/14e590afe938f886859b4e38ce30a9ef8d6f43b7/u8g2/Makefile#L25-L26

bradjc:
Need a https://github.com/tock/libtock-c/blob/master/lvgl/Makefile.setup and can you add a readme to the `libtock/util` folder to explain what goes in it? I assume we want it to be only be utilities related to abstractions provided by the tock kernel.

lschuermann:
> @lschuermann
> 
> https://github.com/tock/libtock-c/blob/14e590afe938f886859b4e38ce30a9ef8d6f43b7/u8g2/Makefile#L25-L26

Still doesn't fix the formatting errors on the `lwip` submodule, unfortunately. Any hint on how to disable uncrustify for the submodule?

examples/lwip_ethernet_tap/main.c
```diff
@@ -0,0 +1,444 @@
...
+
+// TAP-driver frame reception upcall handler.
+//
+// We only use this as a mechanism for the kernel to inform the app that at
+// least one frame has been placed into the streaming process slice. Hence this
+// simply sets the `frame_received` flag passed in `data`:
+static void frame_rx_upcall(__attribute__((unused)) int   p0,
+                            __attribute__((unused)) int   p1,
+                            __attribute__((unused)) int   p2,
+                            __attribute__((unused)) void* data) {
```
brghena, line 111: Copy-paste error: `data` _is_ actually used in this function

examples/lwip_ethernet_tap/main.c
```diff
@@ -0,0 +1,444 @@
...
+  if (statuscode != TOCK_STATUSCODE_SUCCESS) {
+    printf("-> frame_tx_upcall error, statuscode: %04lx\r\n", statuscode);
+    return;
+  }
+
+  debug_print("-> frame_tx_upcall: len %d, id %lu\r\n", len, transmission_id);
+}
+
+// LwIP packet (Ethernet frame) transmission callback:
+static err_t tapif_output(__attribute__((unused)) struct netif* tapif,
```
brghena, line 145: `tapif` is used in this function

examples/lwip_ethernet_tap/main.c
brghena: It would be good to include an example of "it worked on this board at this Tock release"

lwip/Makefile
```diff
@@ -0,0 +1,30 @@
...
+# This Makefile's directory, library is actually located in the `lwip`
+# submodule under this directory
+$(LIBNAME)_DIR := $(TOCK_USERLAND_BASE_DIR)/$(LIBNAME)
+
+# Include the filelist makefile of lwip. The filelist requires a variable named
+# `LWIPDIR` to be in scope, which points to the `src` directory inside the git
+# submodule. This is not to be confused with `$(LIBNAME)_DIR` (expands to
+# `lwip_DIR`), which is the top-level Tock directory name:
+LWIPDIR := $($(LIBNAME)_DIR)/$(LIBNAME)/src
+include $($(LIBNAME)_DIR)/lwip/src/Filelists.mk
```
brghena, line 14: We discussed on the Network working group call moving this to include a vendored file, as the submodule might not be initialized yet by the time this line runs.

brghena (changes requested):
Some small things and reminders. But this is just about ready to go once you have time to update it.

ppannuto:
Okay, I believe all the make/build nonsense should be fixed. One thing I'm not 100% sure is okay is https://github.com/tock/libtock-c/pull/494/commits/3552552f12ab01225f3f3203f53a2f462269e4c7, but I don't really know what else to do here.

I think it's okay because newlib doesn't `undef` before trying to define `BYTE_ORDER`, so hopefully if there were ever a conflict due to include ordering, it'd show up as a redefinition warning?

brghena (commented):
Two more small nits

brghena (approved):
Looks good to me. This is a really great addition to userspace.

brghena:
This is ready to go. So if anyone still has opinions, please comment soon.

lschuermann:
@brghena Fixed a small indentation mismatch in the README.

brghena (approved):

PR #495, opened 2025-02-16
shell.nix: add specific uncrustify version (0.75.1)

lschuermann:
The libtock-c formatting scripts expect to use a specific version of `uncrustify` (v0.75.1, currently), and will attempt to build it if not present. Instead of using whatever version the user's Nixpkgs has, we build the specific `uncrustify` version required by `libtock-c`.


ppannuto (approved):
fwiw, rustfmt in the kernel does the same version pinning for the same reason — the format tools change defaults with versions, so you have to pin the version (and not use the system package manager version, be it nix or apt or whatever) to avoid churn

PR #497, opened 2025-03-12
tests: add an MPU test that circumvents libc

ppannuto:
**Untested.**

Inspired by the [recent MPU configuration bug](https://github.com/tock/tock/issues/4366), here is a MPU test that is 'bare metal', i.e. would not be saved by the implicit `sbrk` in Tock's libc startup. Note, that this specific test would not actually catch the issue — this currently provides coverage of everything that the app should access without fault. But the goal is to provide complete coverage of the 'good cases' and be easy to modify for specific test cases.


alevy (approved):

lschuermann:
@ppannuto How should we move forward on this? Are you still planning on doing some more testing, should someone else do that, or should we just merge?

lschuermann:
@tyler-potyondy @vrindisbacher Did you also test this on RISC-V?

lschuermann:
Cool, that's good to hear :sweat_smile:. I'll put it on my todo list to run this app on RISC-V, just to make sure we won't be surprised later if the test app itself somehow was incorrect.

lschuermann:
@vrindisbacher @ppannuto This seems to _almost_ work on RISC-V, apart from one error that I believe should also affect Cortex-M. Fix and discussion thread for that in #535.

PR #498, opened 2025-03-12
tests: Organize and label MPU unit tests

ppannuto:
We have lots of MPU tests, and it's not necessarily obvious why.

This aims to give a "memory map" to the array of tests, as well as to make more obvious where you might start for any interactive testing.


alistair23 (approved):

bradjc:
Why do we need another sub folder?

ppannuto:
We don't _need_ a subfolder, but I'll quote myself from the README updates for the rationale:

> MPU test applications are necessarily somewhat fine-grained, as each MPU
violation results in termination of the test process.
> 
> As it can be annoying to flash myriad apps during development, the tests in the
> top-level directory attempt to exercise all of the reads and writes of regions
> that _should_ work during normal operation (and allow use of a button to inject
> faults for interactive testing). The `unit/` folder holds all of the smaller
> tests which attempt to violate access rules at each of the boundaries.

ppannuto:
Yes we do?

`mpu_walk_region` is still in the root folder here, and #497 adds a second root-level test.

alevy (approved):

examples/tests/mpu/unit/mpu_ram_end/Makefile
lschuermann, line 1: Weird that this is the only `Makefile` in the diff that git doesn't seem to track as a rename? (at least using GitHub's diff settings / algorithm)

lschuermann (approved):

Issue #377, opened 2024-03-22
remove mention of RISCV flag support in readme


lauralt:
Hi,

I've been trying to build the `c_hello` application for a Cortex-M target with WSL. I've noticed in the Readme that there is a RISCV flag that can enable/disable the compilation for the riscv targets. From what it looks like from the scripts, RISCV flag is no longer supported. The README should be updated accordingly in that case.

Also, is there a way to compile an application only on Cortex-M (without commenting the riscv targets in Configuration.mk :D)? If not, why did you choose to build on both of them as the default, and I think, the only option.

lauralt:
It worked, thanks! I'll submit a PR in the following days.

lauralt:
@lschuermann  I see the flag was removed in https://github.com/tock/libtock-c/commit/4de0e701932a3ff8565681e4dd172055ce78d9f5, so I will close the issue.

Might be worth documenting also the `TOCK_TARGETS` option you mentioned in the README.

eddyp:
> @lschuermann I see the flag was removed in [4de0e70](https://github.com/tock/libtock-c/commit/4de0e701932a3ff8565681e4dd172055ce78d9f5), so I will close the issue.
> 
> Might be worth documenting also the `TOCK_TARGETS` option you mentioned in the README.

I agree documenting `TOCK_TARGETS` especially since one might be interested in an Arm Cortex M target, but being forced to also install a RISC-V toolchain makes little sense from that PoV. Also, for build speed reasons it makes sense to be able to selectively choose what is to be compiled and ran.

PR #501, opened 2025-04-09
If not targeting an architecture family don't run it's compiler

ratmice:
This is related to #377, when running

```
TOCK_TARGETS="cortex-m4" ./build_all.sh
```

One ends up getting a lot of messages if they don't have the riscv compiler installed,

```console
make: riscv64-unknown-elf-gcc: No such file or directory
```

This appears to be because `Configuration.mk` runs the compiler to try and extract the compiler version.
This patch tries to only run the compiler/extract the version if the `TOCK_TARGETS` contains the string `rv32`.

I tested two cases:
* With `TOCK_TARGETS="cortex-m4"`, ensuring it doesn't produce the error
* Without `TOCK_TARGETS` set, ensuring it still does because the default value contains the string.

### DIFF
--- a/Configuration.mk
+++ b/Configuration.mk
@@ -249,8 +249,10 @@ CC_rv32imac := $(CC_rv32)
 
 # Determine the version of the RISC-V compiler. This is used to select the
 # version of the libgcc library that is compatible.
-CC_rv32_version := $(shell $(TOOLCHAIN_rv32)$(CC_rv32) -dumpfullversion)
-CC_rv32_version_major := $(shell echo $(CC_rv32_version) | cut -f1 -d.)
+ifneq ($(findstring rv32i,$(TOCK_ARCH_FAMILIES)),)
+  CC_rv32_version := $(shell $(TOOLCHAIN_rv32)$(CC_rv32) -dumpfullversion)
+  CC_rv32_version_major := $(shell echo $(CC_rv32_version) | cut -f1 -d.)
+endif
 
 # Match compiler version to support libtock-newlib versions.
 ifeq ($(CC_rv32_version_major),10)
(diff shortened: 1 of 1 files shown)


ppannuto:
You can check `TOCK_ARCH_FAMILIES` instead of parsing the targets again yourself, e.g. out of the box 

    TOCK_ARCH_FAMILIES=cortex-m rv32i

It'd be reasonable to add that to the config print from the verbose build to improve discoverability as part of this PR probably.

---

Yes, `TOCK_BOARD` should be removed from the documentation. I think the last vestiges of it were removed from here around a year ago as tockloader improved. It's also in imix documentation, which looks like it has more than should be purged, e.g., [I don't think any of this paragraph is valid with modern tockloader](https://github.com/tock/tock/blob/master/boards/imix/README.md?plain=1#L87-L90).

ratmice:
I used `TOCK_ARCH_FAMILIES`, and added it to the verbose build output like you asked.

Yeah, I'll work on a patch for the nordic stuff that I can figure out, but I was really not sure what to do about imix especially the parts about the application address, so I'll probably just leave that alone and make a note of that in an issue.

Configuration.mk
```diff
@@ -686,6 +688,7 @@ ifneq ($(V),)
   $(info MAKEFLAGS=$(MAKEFLAGS))
   $(info PACKAGE_NAME=$(PACKAGE_NAME))
   $(info TOCK_ARCHS=$(TOCK_ARCHS))
+  $(info TOCK_ARCH_FAMILIES=$(TOCK_ARCH_FAMILIES))
```
ratmice, line 691: There is apparently one more place to do this check in this configuration dump.

ratmice:
I figured I should also do the same for the `cortex-m` family.

ppannuto (approved):

ppannuto:
This all looks good to me, thanks for the patches!

brghena (approved):

PR #503, opened 2025-05-09
ci: group output of builds separately

ppannuto:
Now instead of the world's longest (and slowest to load) ci text output, we get:
image

(Where you can click to expand any of those groups to see their contents)


alevy (approved):
@ppannuto is the King of bash and the earl of Makefiles. The King of bash and the earl of Makefiles is the real ruler of these lands.

PR #502, opened 2025-05-07
Update default compilation to C23 (from C11)

ppannuto:
This is basically a no-op at this point, as C23 didn't really break anything*.

The prior selection of C11 was mostly arbitrary. At the time the first
libtock-c build infrastructure was built, C17 hadn't been ratified yet,
but we needed some newer stuff than C99 provided, so we explicitly set
things to C11 (well, gnu11).

We have an impending need for the new C23 `#embed` compiler directive,
so this version bump lays the groundwork.

Future PRs may also look at other new C23 features; there are many that
are likely useful to libtock-c.

*There was one function signature it's more persnickety about, which
is the other commit in this PR.

PR #504, opened 2025-05-14
Lr1110 quickfixes

ppannuto:
After #502, the (newer than CI) compiler (on my machine) is a bit stricter / catches some things that result in hard errors.

These fixup the missing includes and other small warnings that were now promoted to errors.

The switch away from `unzip + mv + rmdir` is to (1) simplify, and (2) handle hidden directories correctly (since there is now a `.github/` folder in the upstream root).

### DIFF
--- a/lr1110/Makefile.setup
+++ b/lr1110/Makefile.setup
@@ -1,7 +1,7 @@
 TOCK_USERLAND_BASE_DIR ?= ..
 LR1110_DIR = $(TOCK_USERLAND_BASE_DIR)/lr1110/lr1110
 
-VERSION_HASH := 7c33c9ef136c841a7b44f530d2415789c86f48af
+VERSION_HASH := e34742993932c97731f513ffa3028d284b163d31
 
 SEEED_DIR := $(LR1110_DIR)/seeed
 ZIP_FILE := $(SEEED_DIR).zip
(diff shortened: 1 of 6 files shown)


lr1110/lr1110/src_changed/smtc_hal_mcu.c
```diff
@@ -57,7 +59,7 @@ void hal_hex_to_bin( char *input, uint8_t *dst, int len )
 {
 	char tmp[3];
 	uint16_t length = strlen( input );
-	tmp[2] = NULL;
+	tmp[2] = '\0';
```
brghena, line 62: Big fan of this change. I frequently tell my students 0, False, '\0', and NULL are separate things and should be used in semantically meaningful ways. /rant

lr1110/Makefile.setup
```diff
@@ -27,12 +27,15 @@ $(LR1110_DIR)/radio_planner.c: $(SEEED_DIR)/smtc_hal/smtc_modem_hal.c
...
+	mkdir -p $(SEEED_DIR)
+
+$(SEEED_DIR)/smtc_hal/smtc_modem_hal.c: | $(SEEED_DIR)
 	curl -L --output $(ZIP_FILE) "https://codeload.github.com/Seeed-Studio/Seeed_Wio_WM1110_Dev_Board/zip/$(VERSION_HASH)"
-	unzip -q $(ZIP_FILE) -d $(SEEED_DIR)
-	rm -f $(ZIP_FILE) # Remove zip file after unzip
-	mv $(EXTRACTED_DIR)/* $(SEEED_DIR)/
-	rmdir $(EXTRACTED_DIR)  # move unzipped code to SEEED_DIR, remove original directory
+	# Strip-components removes leading directory path (i.e., Seeed_Wio_WM1110_Dev_Board-$(HASH)/) during extraction
+	bsdtar --strip-components=1 -xf $(ZIP_FILE) --cd $(SEEED_DIR)
```
brghena, line 37: Are we sure this is the tool to go with? It's not installed on Ubuntu by default (although it is installable). I'm worried if it's available on other Linux distros. I'm not saying it needs to change, but we should consider it.

brghena (changes requested):

lr1110/Makefile.setup
ppannuto: Ugh, midnight google made that seem like an easy / robust solution, and it happened to already be on my machine so I was hopeful. No, I don't think this is worth installing a new tool for; I'll change this.

ppannuto: ... to wit, plain `tar` also supports that flag 🙃

bradjc (approved):

brghena (approved):

PR #510, opened 2025-06-12
run AUTOGEN_HEADER only when APPS_TO_EMBED is changed

viswajith-g:


examples/tutorials/dynamic-apps-and-policies/app_loader/Makefile
```diff
@@ -21,9 +21,19 @@ APPS_TO_EMBED := \
 	# ../../../blink \
 
 AUTOGEN_HEADER = loadable_binaries.h
+APPS_HASH := build/$(TOCK_TARGETS)/apps_to_embed
```
bradjc, line 24: We can't use `$(TOCK_TARGETS)` because that could be an array.

PR #514, opened 2025-06-16
Clean up unused task queue code

ZhekaS:
This PR removes code and API related to task queue which isn't in use anymore.

### DIFF
--- a/libtock-sync/kernel/read_only_state.c
+++ b/libtock-sync/kernel/read_only_state.c
@@ -1,18 +1,14 @@
 #include "read_only_state.h"
 
 returncode_t libtocksync_read_only_state_quick_yield(void* base, int* pending_tasks) {
-  if (yield_check_tasks()) {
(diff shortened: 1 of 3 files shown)


ppannuto:
@alevy I think you wrote the task queue stuff (long ago now), will let this sit for a few days (or until you hit merge) in case there's some context I'm unaware of.

lschuermann (approved):

bradjc (dismissed):
I want to wait for @alevy to review.

alevy (approved):
Yes, this is good. The task queue was originally part of the virtual time implementation in user space. @bradjc helpfully removed that implementation in de9dd0e5ecab6102c3f8bafa8e56fec6a456ff93 with no comment, as part of the split of libtock and libtock-sync. But I believe the current virtual timers are better overall, and if we ever need a queue, we can surely either re-implement one or revive this big beautiful building-block.

PR #521, opened 2025-06-24
Implement Yield-WaitFor, add guide doc for YWF, demonstration with temperature

bradjc:
This is a version of #345. The `yield_wait_for()` implementation is the same.

This codifies in `doc/guide.md` that all `libtock-sync` APIs must use Yield-WaitFor and provides a format for doing so.

I think this design is a reasonable compromise. The upside is `yield_wait_for()` use remains _only_ in libtock-sync. The downside is there is _some_ duplication, as the mapping from syscall args to meaningful variables is copied, and the high-level "how do I do a X operation" is copied to both libtock and libtock-sync. However, we can't re-use the libtock operations as those all set an upcall which we no longer want.

I've sketched out what this looks like with a very simple driver (temperature).

### DIFF
--- a/doc/guide.md
+++ b/doc/guide.md
@@ -236,7 +236,7 @@ they should have the last argument be a callback function pointer.
 returncode_t libtock_[name]_[desc](<arguments>, libtock_[name]_callback_[desc] cb);
 ```
 
-### Example
+### Example:
 
 
 For example, a library called "sensor" with a sensor read operation should look
@@ -294,7 +294,7 @@ application upcall will run until the synchronous API has finished.
 | Source File Name | `libtock-sync/[category]/syscalls/[name]_syscalls.c` |
 | Header File Name | `libtock-sync/[category]/syscalls/[name]_syscalls.h` |
 
-### Header File
+### Synchronous Syscall Header File
 
 The `[name]_syscalls.h` must be wrapped in `extern "C" { ... }` if the header
 file is used in a C++ app.
@@ -354,7 +354,7 @@ returncode_t libtocksync_[name]_yield_wait_for(int* value) {
 }
 ```
 
-### Synchronous APIs
+### Synchronous Operations
 
 The asynchronous operations should have matching synchronous versions using
 Yield-WaitFor internally.
@@ -365,7 +365,7 @@ Yield-WaitFor internally.
 | Source File Name | `libtock-sync/[category]/[name].c` |
 | Header File Name | `libtock-sync/[category]/[name].h` |
 
-### Header Files
+### Synchronous Operation Header Files
 
 The libtock-sync `[name].h` header file must look like:
 
@@ -386,7 +386,7 @@ extern "C" {
 #endif
 ```
 
-### Synchronous APIs
+### Example:
 
 For our sensor example:
 


doc/guide.md
```diff
@@ -337,17 +417,13 @@ operation:
 
 ```c
 returncode_t libtocksync_sensor_read(int* val) {
-  int err;
-  result.fired = false;
+  returncode_t err;
 
-  err = libtock_sensor_read(sensor_cb);
+  err = libtock_sensor_command_read(sensor_cb);
```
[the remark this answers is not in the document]
bradjc: Yes good catch.
```suggestion
  err = libtock_sensor_command_read();
```

alevy:
Was it resolved elsewhere that we definitely want _every_ sync call to be YieldForWait? That (not exactly, subtlety to follow) excludes patterns where you want, e.g., a periodic timer to fire and do some unobtrusive stuff while be sync waiting for some very log event (e.g. user input).

OK, so you _can_ still do that, you just would need to do `yield_for` yourself, the way it is currently used in the library. And also, of course, maybe if the application is concurrent at all, it should all be async.

alevy:
Without regard for that question though, this seems good to me.

alevy:
I'm not against this. It will basically result in two userlevel libraries that really ought not be mixed: the async and sync ones. The sync one will be able to dispense with things like printf buffering and virtual timers, but will only be capable of one outstanding operation at a time. The async can do as many things as you want at once, but will be async.

alevy:
I'd like to know how this might affect programs that we don't see that downstream users have written and deployed. If it is the case that C programs are almost entirely either one-at-a-time or async (or maybe with like just one sync-looking operation that wouldn't be a big problem to stub in the app itself), this seems maybe good.

bradjc:
> I'd like to know how this might affect programs that we don't see that downstream users have written and deployed. If it is the case that C programs are almost entirely either one-at-a-time or async (or maybe with like just one sync-looking operation that wouldn't be a big problem to stub in the app itself), this seems maybe good.

Even our upstream apps mix sync/async, in arguably reasonable ways. I think we typically use sync functions, but for events (eg buttons) we use callbacks.

Hmm...I wonder if with the switch to YWF we could have the sync header _not_ include the async header. That would make it more obvious if an app was including both versions for the same driver based on the includes. But, that would require either a change to how the libraries are implemented or duplicating content in the libraries, specifically for commands that do not start async operations (eg getting the number of buttons).

ppannuto (dismissed):

libtock/tock.h
```diff
@@ -127,6 +127,18 @@ typedef struct {
...
+// Return structure for a Yield-WaitFor syscall. The return value are the
+// arguments that would have been passed to the upcall the the Yield-WaitFor was
+// waiting on.
+typedef struct {
+  // Upcall argument 1.
+  uint32_t data0;
+  // Upcall argument 2.
+  uint32_t data1;
+  // Upcall argument 3.
+  uint32_t data2;
```
brghena, line 139: One nit question: the arguments are `uint32_t`, but above in `libtock/tock.h`, upcall function pointers are defined as `typedef void (subscribe_upcall)(int, int, int, void*)`. Is the difference meaningful? Do we care about it?

To me, it feels like maybe the `subscribe_upcall` signature is what's wrong.

doc/guide.md
```diff
@@ -277,7 +277,87 @@ returncode_t libtock_sensor_read(libtock_sensor_callback_reading cb) {
...
 
-Most system call interfaces will want to provide a synchronous API as well.
+Most system call interfaces will want to provide a synchronous API as well. This
+requires another file for syscalls.
+
+All synchronous APIs MUST use the
+[Yield-WaitFor](https://book.tockos.org/trd/trd104-syscalls.html#413-yield-waitfor)
+(`yield_wait_for()`) variant of the yield syscall. This ensures predictable
+behavior for `libtock-sync` users because Yield-WaitFor ensures that no other
+application upcall will run until the synchronous API has finished.
```
brghena, line 287: Do we intend to move all the other synchronous APIs to Yield-WaitFor in this same PR? Otherwise the docs and the code will be out-of-sync.

brghena (commented):
I can get behind this. Although we certainly shouldn't merge before discussing at the meeting.

brghena:
I realized the doc had three different headers titled "Synchronous APIs", so I cleaned up the header names throughout the doc and pushed a commit.

libtock/tock.h
bradjc: Agree that subscribe upcall is older and was created before we got more explicit about syscall types.

doc/guide.md
bradjc: I'm not really excited about another massive change that requires touching every driver. I think we need to make forward progress on YWF and we haven't been able to so far. I would like to open a tracking issue for all of the drivers and convert them over time.

brghena: That's totally reasonable. Maybe the doc should say that this is the ideal and synchronous drivers will be changed over to this over time?

ppannuto (approved):

ppannuto:
Discussed during core team call — consensus to merge this as baseline and continue to iterate with sync/async nuances as we gain more experience.

brghena:
There was still an outstanding discussion on this to update the docs. I guess that can happen in an additional PR, but you jumped too quickly here Pat.

PR #522, opened 2025-06-24
Uncrustify: Update and include new config options

bradjc:
This is another round of trying to use uncrustify to make our code consistent and preferably similar in structure to what rustfmt does. Some features are new so this updates uncrustify. That caused some changes as uncrustify fixed some bugs.

Each option/change is its own commit.

New changes:
- Force a newline at the end of files
- Use unix line endings
- Align "\" character in macros
- Force 1 or 2 blank lines between functions
- Remove newlines at the end of a function
- Sort `#include`s. This works pretty well!


lschuermann:
@bradjc I tacked on an update to the uncrustify dependency in `shell.nix`.

lschuermann:
> * Sort `#include`s. This works pretty well!

This breaks one of the `u8g2` demo apps:

```diff
leons@iron ~/p/t/l/e/t/u/led-menu (uncrustify-config-2025-06)> git diff master -- main.c
diff --git a/examples/tests/u8g2-demos/led-menu/main.c b/examples/tests/u8g2-demos/led-menu/main.c
index 1a498719..ccd3f1c3 100644
--- a/examples/tests/u8g2-demos/led-menu/main.c
+++ b/examples/tests/u8g2-demos/led-menu/main.c
@@ -6,11 +6,10 @@
 #include <libtock/interface/led.h>
 
 // These have to be included before mui.h
-#include <u8g2-tock.h>
-#include <u8g2.h>
-
 #include <mui.h>
 #include <mui_u8g2.h>
+#include <u8g2-tock.h>
+#include <u8g2.h>
 
 u8g2_t u8g2;
 mui_t ui;
```

```
leons@iron ~/p/t/l/e/t/u/led-menu (uncrustify-config-2025-06)> make
  CC        main.c
In file included from main.c:10:
../../../../u8g2/u8g2-c4f9cd9f8717661c46be16bfbcb0017d785db3c1/csrc/mui_u8g2.h:165:1: error: unknown type name 'u8g2_uint_t'
  165 | u8g2_uint_t mui_get_x(mui_t *ui);
      | ^~~~~~~~~~~
../../../../u8g2/u8g2-c4f9cd9f8717661c46be16bfbcb0017d785db3c1/csrc/mui_u8g2.h:166:1: error: unknown type name 'u8g2_uint_t'
  166 | u8g2_uint_t mui_get_y(mui_t *ui);
      | ^~~~~~~~~~~
../../../../u8g2/u8g2-c4f9cd9f8717661c46be16bfbcb0017d785db3c1/csrc/mui_u8g2.h:167:1: error: unknown type name 'u8g2_t'
  167 | u8g2_t *mui_get_U8g2(mui_t *ui);
      | ^~~~~~
../../../../u8g2/u8g2-c4f9cd9f8717661c46be16bfbcb0017d785db3c1/csrc/mui_u8g2.h:169:42: error: unknown type name 'u8g2_uint_t'  169 | void mui_u8g2_draw_button_utf(mui_t *ui, u8g2_uint_t flags, u8g2_uint_t width, u8g2_uint_t padding_h, u8g2_uint_t padding_v, const char *text);
      |                                          ^~~~~~~~~~~
[...]
```

ppannuto:
Ah, thought that merged upstream PR meant it was a non-issue already — guessing some other updates to the library are bringing unhappiness when bumping that submodule?

lschuermann:
> After this, the `libc` and `nrf` things end up mixed up :/. Now, `libtock` hangs out on its own, because we intentionally set an expectation of pointing the compiler include path below the `libtock/**.h` things (which allows uncrustify regex rules to work). But I don't think we can do the same for something like `#include <nrf/eddystone.h>` since any includes _inside_ `eddystone` would break.

@ppannuto I do agree with you on that, and I think this example is sufficient to sway me to say I'd rather not see include sorting be mandatory. At the very least, we should ensure that any stdlib includes come before other library includes (which, I guess, would mean exhaustively listing out stdlib header files in the `uncrustify.cfg`...).

bradjc:
> > After this, the `libc` and `nrf` things end up mixed up :/. Now, `libtock` hangs out on its own, because we intentionally set an expectation of pointing the compiler include path below the `libtock/**.h` things (which allows uncrustify regex rules to work). But I don't think we can do the same for something like `#include <nrf/eddystone.h>` since any includes _inside_ `eddystone` would break.
> 
> @ppannuto I do agree with you on that, and I think this example is sufficient to sway me to say I'd rather not see include sorting be mandatory. At the very least, we should ensure that any stdlib includes come before other library includes (which, I guess, would mean exhaustively listing out stdlib header files in the `uncrustify.cfg`...).

I guess it's a tradeoff between allowing very carefully addressed cases and bringing everything up to a general standard. I prefer to have the floor. For example I think having this consistency makes more of a difference:

image

bradjc:
Oh, this builds now!

I agree the sort includes makes the ble examples worse. However, that is because those includes are not namespaced, which is a mistake, and one that we shouldn't bother fixing because no one uses hail/imix at this point.

lschuermann:
> I agree the sort includes makes the ble examples worse. However, that is because those includes are not namespaced, which is a mistake, and one that we shouldn't bother fixing because no one uses hail/imix at this point.

Can we, to mitigate some of the worst cases, define a regex that groups all `<std*` includes together at least?

bradjc:
> Any reason why the last two include categories aren't flipped? I always used to group as standard library, then libtock, then anything else.

My instinct is reverse order by "distance", stdlib is farthest from the app, third party libraries are next, then our libtock library, then its own local resources.

PR #518, opened 2025-06-21
tutorials: root-of-trust: add Root of Trust Tutorial screen and encryption service

pqcfox:
This PR adds the applications for the Tock Root of Trust tutorial.

PR #523, opened 2025-06-24
tutorials: root-of-trust: modify HWRoT SRAM dump application to use Memop syscall for detecting app memory start

pqcfox:
Small addition to #518 (created so as to not invalidate reviews) to make the HWRoT SRAM dump application use the Memop syscall to detect its own application start.


pqcfox:
Merging into #518 for simplicity!

PR #525, opened 2025-06-25
Build system: handle library version updates and update u8g2 library

bradjc:
This PR adds a feature to `AppMakefile.md` to rebuild libraries if their version updates. The issue is that if a user already has compiled output for a library, if the version changes that compiled output is still there and make doesn't know to rebuild. This change allows libraries to specify a particular file as a dependency that make can track, and if the version updates that file will not exist and the library gets rebuilt.

This isn't necessary for completely new libraries as the pre-built files won't exist. This also doesn't affect libraries that only ever have one version.

As an example, I updated u8g2 to use this new feature.


alevy (approved):

brghena (approved):
Looks reasonable and it all compiled when I tried it

PR #538, opened 2025-07-16
doc: guide: update with more style changes

bradjc:
- Add unallow example in libtock-sync
- Add note about _types.h file

### DIFF
--- a/doc/guide.md
+++ b/doc/guide.md
@@ -410,3 +410,47 @@ returncode_t libtocksync_sensor_read(int* val) {
   return err;
 }
 ```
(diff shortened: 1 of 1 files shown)


ppannuto:
I updated the example here to use the `defer` pattern introduced in #541.

(Also, in support of `defer`, there was a bug in the `goto exit` approach written previously that this fixes 😄)

ppannuto (approved):

bradjc:
Merge this?

ppannuto (approved):

PR #540, opened 2025-07-16
CI: Don't pin ARM GCC version

ppannuto:
The removes the version pin from the arm gcc toolchain used in CI. We don't pin the RISC-V version, and I don't think we should pin ARM either (just use 'latest' implicitly instead). If new compilers have issues, it'd be nice to catch them in CI proactively. Warnings also get measurably better with each version, and I doubt we intended to have a compiler in CI that's 5 years out of date...

This PR also removes the toolchain install from the `ci-format` job, since the toolchains aren't used there anyway.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,11 +22,6 @@ jobs:
       - uses: actions/checkout@v2
         with:
           submodules: false # LVGL makefile manually installs the submodule
-      - uses: carlosperate/arm-none-eabi-gcc-action@v1
-        with:
-          release: '10.3-2021.07' # The arm-none-eabi-gcc release to use.
-      - name: setup-riscv-toolchain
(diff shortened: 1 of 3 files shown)


bradjc (dismissed):

lschuermann:
I don't like the idea of a compiler update silently breaking our code. A nicer model would be an updater bot that files PRs for new major revisions, or on a fixed schedule (like NixOS/nixpkgs has). But realistically I won't have time for that, so I'm happy to go with not pinning as the next best alternative. :smile:

ppannuto:
Yeah, it's both a safe bet that I originally pinned the compiler version for exactly that reason, and then have also never updated it... :)

However, as the subsequent commits illustrate, our pinning policy has been hiding warnings from new compilers, so I think 'latest' is indeed the lesser-evil here.

PR #541, opened 2025-07-16
Yield-WaitFor: Introduce `defer` and update HMAC as example usage

ppannuto:
This is a minimal example of what "cleanup attribute" like code would look like.

This probably isn't the exact best way to implement it (due to the "magic variable name" of `ret_{param}`), but it gives a sense of what it would look like.

### DIFF
--- a/libtock/defer.h
+++ b/libtock/defer.h
@@ -1,0 +1,18 @@
+#pragma once
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#define __DEFER__(F, V) \
+  auto void F(int*); \
+  [[gnu::cleanup(F)]] int V; \
+  auto void F(int*)
+
+#define defer __DEFER(__COUNTER__)
+#define __DEFER(N) __DEFER_(N)
+#define __DEFER_(N) __DEFER__(__DEFER_FUNCTION_ ## N, __DEFER_VARIABLE_ ## N)
+
+#ifdef __cplusplus
+}
+#endif
--- a/libtock/defer.h
+++ b/libtock/defer.h
@@ -5,9 +5,9 @@ extern "C" {
 #endif
 
 #define __DEFER__(F, V) \
-  auto void F(int*); \
+  auto void F(int* V ## SENTINEL); \
   [[gnu::cleanup(F)]] int V; \
-  auto void F(int*)
+  auto void F(__attribute__ ((unused)) int* V ## SENTINEL)
 
 #define defer __DEFER(__COUNTER__)
 #define __DEFER(N) __DEFER_(N)
--- a/libtock/defer.h
+++ b/libtock/defer.h
@@ -1,5 +1,14 @@
 #pragma once
 
+// Implementation of the C `defer {}` feature.
+//
+// As of July 2025, upcoming versions of C will support `defer` as a way to run
+// code when a value goes out of scope. See:
+// https://thephd.dev/_vendor/future_cxx/technical%20specification/C%20-%20defer/C%20-%20defer%20Technical%20Specification.pdf
+//
+// This implements the same feature with a macro. This implementation used from
+// https://gustedt.wordpress.com/2025/01/06/simple-defer-ready-to-use/.
+
 #ifdef __cplusplus
 extern "C" {
 #endif
--- a/libtock/defer.h
+++ b/libtock/defer.h
@@ -4,14 +4,14 @@
 extern "C" {
 #endif
 
-#define __DEFER__(F, V) \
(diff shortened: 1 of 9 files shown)


bradjc:
Obviously I'm not a big fan of macros. Seems complicated compared to the goto version.

It also isn't very flexible for drivers that may need more cleanup.

ppannuto:
Okay, I found a much cleaner implementation of this, which matches the (current) [draft specification for the next release of C](https://thephd.dev/_vendor/future_cxx/technical%20specification/C%20-%20defer/C%20-%20defer%20Technical%20Specification.pdf), i.e., in the future `defer` becomes a keyword, we delete our header, and no code needs to change.

bradjc:
The latest version is nice, even if I don't understand the C code in the macros. But I'm worried about backwards compatibility.

ppannuto:
What's the worry w.r.t. backwards compatibility? This `defer` macro only exists in C files that explicitly include `libtock/defer.h`, it only changes code where it's used — so you simply write code with `defer` instead of `goto exitN`.

As for the macro, the fully expanded form basically does this:

```c
defer { foo(); bar(); }
   // ^^^^^^^^^^^^^^ this bit doesn't change at all

// The above expands to:

auto void __DEFER_FUNCTION0(int* __DEFER_VARIABLE0SENTINEL);  // This is just a function prototype
[[ gnu::cleanup(__DEFER_FUNCTION0) ]] int __DEFER_VARIABLE0; // This declares a variable which is never used, but the compiler can see when it goes out of scope, and will call __DEFER_FUNCTION0 when it goes out of scope
auto void __DEFER_FUNCTION0(__attribute__ ((unused)) int* __DEFER_VARIABLE0SENTINEL) { foo(); bar(); }
                                                                                  // ^^^^^^^^^^^^^^^^^ here's our unchanged code, where it's now the body of a function that's being defined
```

ppannuto:
> I rebased this on master so the CI can pick up a compiler newer than 2021.

Turns out I didn't need to do that, just needed to do [0b840d9](https://github.com/tock/libtock-c/pull/541/commits/0b840d9139aa8678d7547ed8d94d7838e259ee1c). I added a "legacy-compilers" CI run here that runs on the default LTS compilers for both ARM and RISCV and it seems to work without issue.

(Now, I acknowledge that #539 has some lingering issues that I'm working on, but they may actually be unrelated to compiler version; don't let issues over there poison the well here)

bradjc:
Oh this does work on jammy?

bradjc (dismissed):

.github/workflows/ci.yml
```diff
@@ -45,3 +45,29 @@ jobs:
...
+          release: '10.3-2021.07'
+      - run: arm-none-eabi-gcc --version
+      - name: setup-riscv-toolchain
+        run: sudo apt-get install -y gcc-riscv64-unknown-elf
+      - name: riscv-toolchain-version
+        run: riscv64-unknown-elf-gcc --version
+      #- name: ci-build
+      #  run: pushd examples; ./build_all.sh || exit; popd
+      - name: ci-debug-build
+        run: pushd examples/blink; make debug RAM_START=0x20004000 FLASH_INIT=0x30051 || exit; popd
```
brghena, line 73: (I must have forgotten to hit the button on this comment) Could we split this out to a separate PR? I _think_ it's separate from the rest of the changes here.

brghena (changes requested):

bradjc (dismissed):

bradjc (dismissed):

libtock/defer.h
```diff
@@ -0,0 +1,18 @@
...
+#endif
+
+#define __DEFER__(F, V)                \
+        auto void F(int* V##SENTINEL); \
+        [[gnu::cleanup(F)]] int V;     \
+        auto void F(__attribute__ ((unused)) int* V##SENTINEL)
+
+#define defer __DEFER(__COUNTER__)
+#define __DEFER(N) __DEFER_(N)
+#define __DEFER_(N) __DEFER__(__DEFER_FUNCTION_##N, __DEFER_VARIABLE_##N)
```
brghena, line 14: We should cite the source for this code. Probably also put a note that this mirrors the behavior of the C23 `defer` primitive.

(If I was looking at this without context I'd be like "what did those weird Tock people do???")

bradjc: I wrote something.

brghena (approved):

libtock/defer.h
brghena: Thanks!

PR #542, opened 2025-07-16
libtock: Do not include `_syscalls.h` automatically for each driver

bradjc:
Currently if an app uses a `libtock` driver, the `driver_syscalls.h` header is included automatically. This undoes that, requiring apps to explicitly include the syscalls header if the want to use the low-level syscalls.

Benefits:
- `#include`s become a signal that apps are using low-level APIs, which may not be the intention.
- libtock becomes more explicit in how it is being used, in-line with other languages (eg Rust).

Impacts:
- Need to duplicate the `_exists()` function in syscalls (which needs the driver num), in libtock (for async apps) and in libtock-sync (for sync apps).

### DIFF
--- a/doc/guide.md
+++ b/doc/guide.md
@@ -170,13 +170,13 @@ There must be a function to check for syscall driver existence.
 Signature:
 
 ```
-bool libtock_[name]_exists(void);
+bool libtock_[name]_driver_exists(void);
 ```
 
 Example:
 
 ```c
-bool libtock_[name]_exists(void) {
+bool libtock_[name]_driver_exists(void) {
   return driver_exists(DRIVER_NUM_[NAME]);
 }
 ```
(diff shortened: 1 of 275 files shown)


ppannuto:
Rebased; `u8g2` builds fine locally now

ppannuto:
Okay, got this to happy green checkmarks.

@bradjc , you should verify that [this change makes sense](https://github.com/tock/libtock-c/pull/542/commits/17e91e5acf5e419c13022e8549449d50c6f37b37#diff-af7987d41acdc6be7a2dcb6d03b7e82e31b896de837f73eecae9d998abca96f4R78), I honestly don't understand how the original code was ever accepted (or what the `irq` parameter was doing here), but I find the whole lr1110 stack a bit confusing, so it may just be me being confused; the rest of the changes were more rote.

ppannuto (approved):

alevy (approved):

PR #544, opened 2025-07-22
lib: improve robustness of fetch scripts.

ppannuto:
In particular, use `set -e` so that top-level builds stop if fetching fails, rather than trying to continue if, say, something unexpected like `unzip` is missing on a machine.

### DIFF
--- a/lib/fetch-libc++.sh
+++ b/lib/fetch-libc++.sh
@@ -1,5 +1,8 @@
 #!/usr/bin/env bash
 
+set -e
+set -u
+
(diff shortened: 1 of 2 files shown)


lschuermann:
@ppannuto This is what I tried to do in #477, seems like this PR is running into the same issues. Haven't gotten around to developing an approach that works out of the box on macOS...

ppannuto:
Ugh, well, I did have exactly the problem you had (surprise no `unzip` on a new install). I'll close this in favor of figuring out the other one...

PR #546, opened 2025-07-24
ci: Explicitly build for oldest and newest compliers

ppannuto:
This splits the "build" CI rule into two: `build-latest` runs on the newest available OS image with the newest available copmilers, `build-legacy` runs on the oldest active LTS release and uses the 'default' cross compiler toolchains from that LTS.

There's also two little general CI cleanups tacked on the end that clean up the logs some.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,9 +19,10 @@ jobs:
 
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
         with:
           submodules: false # LVGL makefile manually installs the submodule
+          show-progress: false
       - name: Disable wget progress output
         run: |
           echo "verbose = off" >> $HOME/.wgetrc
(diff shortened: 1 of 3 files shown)


brghena (approved):
This is fantastic. Definitely good to check both new and old compilers. Thanks!!

bradjc (approved):

lschuermann:
> Updated to use the build matrix

Awesome, thanks! Yeah, it's a bit more duplication than I anticipated still (because the setup is pretty different), but it'll be nice that we're less likely to have these two go out of sync.

lschuermann:
I ported the legacy branch protection rules of this repo into a ruleset, and changed the required status checks, so this should now be able to move into the merge queue. We might need to close and re-open the PR though...

image

ppannuto:
I think your third check pointed to the wrong thing (the `ci-build-latest` job which doesn't exist now, rather than the `ci-build (ubuntu-latest)` matrix entry). I think I fixed it:
image

lschuermann:
@ppannuto Right, missed that! Third time's the charm.

PR #547, opened 2025-08-08
gpio-async: convert to ywf

ppannuto:
Compile-tested only, fairly straight conversion.

There's a mild deviation from the style guide with the `wait_for_generic_command` instead of 6 different `wait_for_[set|clear|...]` methods. It mirrors how the driver itself is implemented, and I think is a cleaner design and easier to follow than lots of redundant syscall methods.

More generally, I suspect as we port more of these we may find it makes sense to rethink the guidance around ywf methods, but I'm not a that level of confident with just this one.

### DIFF
--- a/libtock-sync/peripherals/syscalls/gpio_async_syscalls.h
+++ b/libtock-sync/peripherals/syscalls/gpio_async_syscalls.h
@@ -1,0 +1,20 @@
+#pragma once
+
+#include <libtock/peripherals/syscalls/gpio_async_syscalls.h>
+#include <libtock/tock.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// Wait for a GPIO-Async operation with no values to finish.
+//
+// In lieu of redundant `wait_for_[set|clear|...]`, generic_command can be used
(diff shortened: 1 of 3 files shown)


libtock-sync/peripherals/syscalls/gpio_async_syscalls.h
```diff
@@ -0,0 +1,20 @@
+#pragma once
+
+#include <libtock/peripherals/syscalls/gpio_async_syscalls.h>
```
bradjc, line 3:
```suggestion
```

So we don't leak libtock in libtock-sync.

ppannuto: It's there to avoid duplicating the `DRIVER_NUM_GPIO_ASYNC`, presumably for the same reason you have the same include here ( https://github.com/tock/libtock-c/pull/553/files#diff-28af97b25b50c3d619c7a3ba099e6a2ff6ebcf972c1fc84800fa932c3c1159d7R3 ) and elsewhere in #553.

bradjc:
Ok I think this is ready to go, pending CI.

PR #551, opened 2025-08-12
lib: fetch: $BASHPID -> $$

bradjc:
$BASHPID wasn't added until bash 4. Mac is older, so use a different pid method.


lschuermann (approved):
I assume you've tested this and verified that it works.

PR #561, opened 2026-02-12
Update `screen` to YWF

bradjc:
Moves types to types.h file, uses defer.

### DIFF
--- a/libtock-sync/display/screen.c
+++ b/libtock-sync/display/screen.c
@@ -1,173 +1,155 @@
+#include <libtock/defer.h>
+#include <libtock/display/screen.h>
 #include <libtock/display/syscalls/screen_syscalls.h>
 
 #include "screen.h"
 
-struct screen_done {
-  returncode_t ret;
-  bool fired;
-};
-struct screen_format {
-  libtock_screen_format_t format;
-  returncode_t ret;
-  bool fired;
-};
-struct screen_rotation {
-  int rotation;
-  returncode_t ret;
-  bool fired;
-};
-
-static struct screen_done result;
-static struct screen_format result_format;
-static struct screen_rotation result_rotation;
-
-
-static void screen_cb_done(returncode_t ret) {
-  result.ret   = ret;
-  result.fired = true;
+static returncode_t screen_set_color(uint8_t* buffer, int buffer_len, int position, size_t color) {
+  // TODO color mode
+  if (position < buffer_len - 2) {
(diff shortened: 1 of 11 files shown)


libtock-sync/display/syscalls/screen_syscalls.c
```diff
@@ -0,0 +1,29 @@
...
+
+  *format = (libtock_screen_format_t) ret.data1;
+
+  int status = tock_status_to_returncode((statuscode_t) ret.data0);
+  return (returncode_t) status;
+}
+
+returncode_t libtocksync_screen_yield_wait_for_rotation(libtock_screen_rotation_t* rotation) {
+  yield_waitfor_return_t ret;
+  ret = yield_wait_for(DRIVER_NUM_SCREEN, 0);
```
brghena, line 23: Shouldn't these three `yield_wait_for` calls be different subscribe numbers?

libtock-sync/display/screen.c
```diff
@@ -1,173 +1,155 @@
...
-  returncode_t ret;
-  bool fired;
-};
-struct screen_rotation {
-  int rotation;
-  returncode_t ret;
-  bool fired;
-};
+static returncode_t screen_set_color(uint8_t* buffer, int buffer_len, int position, size_t color) {
+  // TODO color mode
```
brghena, line 8: Is this TODO something that should be handled now, or a note for the future?

brghena (changes requested):
Two issues here. Otherwise seems fine.

libtock-sync/display/syscalls/screen_syscalls.c
brghena: Or maybe we just have one upcall with contextually different upcall registers?

brghena: We definitely do https://github.com/tock/tock/blob/master/doc/syscalls/90001_screen.md#subscribe

libtock-sync/display/screen.c
bradjc: Ah, no, that is from the existing driver: https://github.com/tock/libtock-c/blob/efc9197bdd479e610fac7a7563d7fd9b8e534b50/libtock/display/screen.c#L195

brghena: Gotcha. The diff just didn't make that clear to me

brghena (approved):

PR #568, opened 2026-04-01
YWF: Convert many more drivers

bradjc:
Mostly done with Claude.

Want to see what happens with CI, and then try to test some of these changes.

### DIFF
--- a/libtock-sync/interface/button.h
+++ b/libtock-sync/interface/button.h
@@ -10,6 +10,17 @@ extern "C" {
 
 bool libtocksync_button_exists(void);
 
+// Set `count` to the number of buttons.
+returncode_t libtocksync_button_count(int* count);
+
+// Read the current button state into `button_value`.
+//
+// ## Arguments
+//
+// - `button_num`: The index of the button.
+// - `button_value`: Will be set to 1 if button is pressed, 0 otherwise.
+returncode_t libtocksync_button_read(int button_num, int* button_value);
+
 // Wait for a specific button to be pressed.
 //
 // This blocks until the button has been pressed. Note, this will not return if
--- a/libtock-sync/interface/button.h
+++ b/libtock-sync/interface/button.h
@@ -19,6 +19,10 @@ returncode_t libtocksync_button_count(int* count);
 //
 // - `button_num`: The index of the button.
 // - `button_value`: Will be set to 1 if button is pressed, 0 otherwise.
+//
+// ## Return Value
+//
+// A returncode indicating whether the button wait was completed successfully.
 returncode_t libtocksync_button_read(int button_num, int* button_value);
 
 // Wait for a specific button to be pressed.
--- a/libtock-sync/net/ieee802154.c
+++ b/libtock-sync/net/ieee802154.c
@@ -1,46 +1,14 @@
 #include <stdio.h>
+#include <string.h>
 
 #include <libtock-sync/services/alarm.h>
+#include <libtock/defer.h>
+#include <libtock/net/ieee802154.h>
 #include <libtock/net/syscalls/ieee802154_syscalls.h>
 
 #include "ieee802154.h"
-struct ieee802154_receive_data {
-  bool fired;
-  int pan;
-  int src_addr;
-  int dest_addr;
-};
 
-static struct ieee802154_receive_data receive_result = { .fired = false };
-
-
-struct ieee802154_send_data {
-  bool fired;
-  bool acked;
-  returncode_t ret;
-};
-
-static struct ieee802154_send_data send_result     = { .fired = false };
-static struct ieee802154_send_data send_result_raw = { .fired = false };
-
-static void ieee802154_receive_done_cb(int pan, int src_addr, int dest_addr) {
-  receive_result.fired     = true;
-  receive_result.pan       = pan;
-  receive_result.src_addr  = src_addr;
-  receive_result.dest_addr = dest_addr;
-}
-
-static void ieee802154_send_done_cb(returncode_t ret, bool acked) {
-  send_result.fired = true;
-  send_result.acked = acked;
-  send_result.ret   = ret;
-}
-
-static void ieee802154_send_raw_done_cb(returncode_t ret, bool acked) {
-  send_result_raw.fired = true;
-  send_result_raw.acked = acked;
-  send_result_raw.ret   = ret;
-}
+#include "syscalls/ieee802154_syscalls.h"
 
 bool libtocksync_ieee802154_exists(void) {
   return libtock_ieee802154_driver_exists();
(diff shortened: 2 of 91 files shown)


bradjc:
I reverted the changes to alarm.c as alarm needs more thought.

bradjc:
Ok I've tested a bunch of these, including the sensors using the hail app and the kv test app.

alevy (dismissed):

libtock-sync/interface/button.h
```diff
@@ -10,6 +10,17 @@ extern "C" {
...
+// Set `count` to the number of buttons.
+returncode_t libtocksync_button_count(int* count);
+
+// Read the current button state into `button_value`.
+//
+// ## Arguments
+//
+// - `button_num`: The index of the button.
+// - `button_value`: Will be set to 1 if button is pressed, 0 otherwise.
+returncode_t libtocksync_button_read(int button_num, int* button_value);
```
brghena, line 26:
```suggestion
// ## Arguments
//
// - `button_num`: The index of the button.
// - `button_value`: Will be set to 1 if button is pressed, 0 otherwise.
//
// ## Return Value
//
// A returncode indicating whether the button wait was completed successfully.
returncode_t libtocksync_button_read(int button_num, int* button_value);
```

This is a bit pedantic, but the other function in this file has full documentation, so we should probably match it here.

libtock-sync/net/ieee802154.c
```diff
@@ -52,42 +20,62 @@ returncode_t libtocksync_ieee802154_send(uint16_t         addr,
...
+  bool acked;
+
+  // Setup the CFG buffer with security parameters.
+  uint8_t buf_cfg[11];
+  buf_cfg[0] = level;
+  buf_cfg[1] = key_id_mode;
+  int bytes = libtock_ieee802154_key_id_bytes(key_id_mode);
+  if (bytes > 0) {
+    memcpy(buf_cfg + 2, key_id, bytes);
+  }
```
brghena, line 33: Leaving a note for myself here.

This is all valid from: https://github.com/tock/libtock-c/blob/master/libtock/net/ieee802154.c

That file uses `buf_cfg[27]`, but this seems correct that it'll only need to be 11 bytes max locally, and send expects it to be fixed at 11 bytes.

libtock-sync/net/udp.c
```diff
@@ -1,59 +1,36 @@
...
-  recv_sync_result.ret   = ret;
-}
-
 bool libtocksync_udp_exists(void) {
   return libtock_udp_driver_exists();
 }
 
 returncode_t libtocksync_udp_send(void* buf, size_t len,
-                                  sock_addr_t* dst_addr) {
+                                  __attribute__ ((unused)) sock_addr_t* dst_addr) {
```
brghena, line 14: This is pretty confusing.

[`libtock_udp_send()`](https://github.com/tock/libtock-c/blob/master/libtock/net/udp.c#L70) copies this `dst_addr` into a local `BUF_TX_CFG` array, but then doesn't use the array at all.

The array _used_ to be allowed to the kernel, but that was quite some time ago: https://github.com/tock/libtock-c/commit/f803963ba6564197a6effe652de83e38ffaf2c74#diff-7df54242ea0f2d5450fcd7ebf92350216fc1a416ce59b8e2fa2bf9f45e1018caL54

My guess is that the remaining pointless code and argument just accidentally never got removed.

But apps sure think that they're sending a message to a destination: https://github.com/tock/libtock-c/blob/master/examples/tests/udp/udp_virt_app_kernel/main.c#L40 Which does seem like a sensible interface. So maybe the bug is in `libtock_udp_send()` itself?

I'm super-duper confused how any of this works.

libtock-sync/storage/app_state.c
```diff
@@ -21,13 +13,8 @@ bool libtocksync_app_state_exists(void) {
...
-
-  err = libtock_app_state_save(app_state_cb);
+  err = libtock_app_state_save(app_state_noop_cb);
   if (err != RETURNCODE_SUCCESS) return err;
 
-  // Wait for the callback.
-  yield_for(&result.fired);
-
-  return result.ret;
+  return libtocksync_app_state_yield_wait_for();
```
brghena, line 19: This seems wrong to me. There's a return code that's getting dropped here. Shouldn't that  be in the ywf data?

brghena (changes requested):
Okay, I read through all of this.

One tiny nit. Also two places where I'm very confused about the resulting behavior and I think something is wrong (although not necessarily this PR's fault or responsibility to fix).

libtock-sync/net/udp.c
brghena: Maybe this should become an issue for udp-knowledgable-people to look into?

brghena: @tyler-potyondy or @hudson-ayers Any thoughts here?

hudson-ayers: I think this breakage actually was introduced in [1548c85c](https://github.com/tock/libtock-c/commit/1548c85ce246849f509679f84c5c92424af97a80). Before that, BUF_TX_CFG was a module-level static, so the buffer was already allowed when we were writing to it. I think we should merge this PR as-is since this breakage already existed, and file an issue to fix the underlying bug

hudson-ayers: Issue filed: https://github.com/tock/libtock-c/issues/576

brghena: Thank you!

libtock-sync/storage/app_state.c
bradjc: I believe nothing is dropped because there is no return code.

https://github.com/tock/tock/blob/master/capsules/extra/src/app_flash_driver.rs#L160C33-L160C48

https://github.com/tock/libtock-c/blob/master/libtock/storage/app_state.c#L14

brghena: Okay, so this won't likely result in a bug. But it still seems weird to me that we're both registering for a callback and calling yield_wait_for.

Shouldn't we just be calling `libtock_app_state_command_save()` directly? https://github.com/tock/libtock-c/blob/master/libtock/storage/app_state.c#L61 (I guess with the init check before it)

brghena: I can't suggest here, but I think the code would be:

```
returncode_t libtocksync_app_state_save(void) {
  returncode_t err;

  if (!_app_state_inited) {
    err = app_state_init();
    if (err != RETURNCODE_SUCCESS) return err;
  }

  err = libtock_app_state_command_save((uint32_t) _app_state_flash_pointer);
  if (err != RETURNCODE_SUCCESS) return err;

  return libtocksync_app_state_yield_wait_for();
}
```

bradjc: But `app_state_init` is static. Using the libtock function with a noop callback avoids copying or exposing the init function.

brghena (approved):

libtock-sync/storage/app_state.c
brghena: Good point. I think that a refactor would improve the app state driver, but that's out-of-scope for this PR

PR #570, opened 2026-04-23
libtock: rtc: use bitshifts

bradjc:
@brghena 

### AI

All claude


brghena (approved):
You're the hero we needed but never deserved.

PR #572, opened 2026-06-16
Add support for RISC-V 64 bit targets

bradjc:
This updates the Configuration.mk makefile to add support for building apps for the `qemu-system-riscv64` virt rv64imac platform.

The major work here (at least initially) is recompiling newlib and libc++ with `-mcmodel=medany`.

From there, we need to handle all of issues that come up when code that was written assuming 32-bit is compiled for 64-bit:

1. The ADC driver passed a pointer in an upcall. That has been replaced with just a buffer index number.
2. printf() statements assumed that `uint32_t` would be `long unsigned ints`. That isn't true on 64-bit. So we use different specifiers that are portable. A lot of this mechanical work was done by claude.
3. Some libraries (nrf serialization, lwip, lvgl) cast `uint32_t`s (or similar) to pointers. That doesn't work on 64-bit, so we just leave those libraries as 32-bit only.
4. Some apps also cast pointers from non register-sized types, and we also just leave those only for 32-bit.
5. The (arbitrary) RAM size we picked wasn't big enough for openthread 64-bit apps. This doubles the RAM size to enable those apps to build on 64-bit.

Some notes:
- crt0.c needs to update as it assumes 32 bit addresses
- other libtock needs to update with the larger addresses
- When I build libc++ on gcc 5.2.0 in docker I only get two rv64 libraries. Is that good enough for now?

### DIFF
--- a/Configuration.mk
+++ b/Configuration.mk
@@ -72,6 +72,12 @@ QEMU_RV32_VIRT_TOCK_TARGETS := rv32imac|rv32imac.0x80100080.0x80300000|0x8010008
                                rv32imac|rv32imac.0x80130080.0x80330000|0x80130080|0x80330000\
                                rv32imac|rv32imac.0x80180080.0x80380000|0x80180080|0x80380000
 
+# Specific addresses useful for the QEMU rv64i "virt" machine memory map.
+QEMU_RV64_VIRT_TOCK_TARGETS := rv64imac|rv64imac.0x80100080.0x80300000|0x80100080|0x80300000\
+                               rv64imac|rv64imac.0x80110080.0x80310000|0x80110080|0x80310000\
+                               rv64imac|rv64imac.0x80130080.0x80330000|0x80130080|0x80330000\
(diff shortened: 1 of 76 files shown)


alevy:
@bradjc The PR description and title need to be completely rewritten. I 100% believe that the motivation behind this PR is to "Add support for RISC-V 64 bit targets" but at least half of the changes are only indirectly related to that: adding a Makefile variable to basically every example app to do something 32 bit IPC and changing every use of `printf` to print an integer to be (I think?) explicitly 32 bit integer?

I can see why those would be useful/necessary to support the first non-32bit target in libtock-c, but there is nothing RISC-V 64 specific about, it's neither mentioned nor justified anywhere, and is the bulk of the reviewing.

bradjc:
Done.

> I can see why those would be useful/necessary to support the first non-32bit target in libtock-c, but there is nothing RISC-V 64 specific about, it's neither mentioned nor justified anywhere, and is the bulk of the reviewing.

It turns out to be a bit difficult to compile all of our examples for riscv64.

alevy (approved):

PR #581, opened 2026-08-05
Update the libtock syscall code to use the new RV64I format from the upcoming TRD

bradjc:
Matches https://github.com/tock/tock/pull/5036

I also fixed the return type to be `returncode_t` for the relevant functions.


ppannuto (dismissed):
Would we be happier in the future / this be clearer if we have one #define TRD104 / TRD_RISCV64 at the top and then we use those throughout to choose instead of evaluating the arch constants throughout?

brghena (approved):
I looked through this all and it makes sense to me.

brghena:
(opened/closed again to try to reset checks)

PR #583, opened 2026-08-14
Add a system call return test that tries all syscall types

bradjc:
Matches https://github.com/tock/tock/pull/5083

To accomplish this, this PR also:
- Adds the helper functions for all of the remaining command return types.
- Implements the converters for rv64 support.
- Fixes the riscv command implementation which was using only u32 types.
- Adds a test which calls each command/subscribe/allow and checks the return values.

### DIFF
--- a/libtock/tock.h
+++ b/libtock/tock.h
@@ -180,6 +180,41 @@ returncode_t tock_command_return_u32_u32_to_returncode(syscall_return_t, uint32_
 // Convert a `syscall_return_t` with a `u64` value to a `returncode_t`.
 returncode_t tock_command_return_u64_to_returncode(syscall_return_t command_return, uint64_t* val);
 
+// Convert a `syscall_return_t` with three `u32` values to a `returncode_t`.
+//
+// This expects exactly three `u32`s to be returned (i.e. the only success case
+// is `TOCK_SYSCALL_SUCCESS_U32_U32_U32`). Do not use with other expected
+// SyscallReturn variants.
+returncode_t tock_command_return_u32_u32_u32_to_returncode(syscall_return_t, uint32_t*, uint32_t*, uint32_t*);
+
+// Convert a `syscall_return_t` with a `u32` and a `u64` value to a `returncode_t`.
+//
+// This expects a `u32` and a `u64` to be returned (i.e. the only success case
+// is `TOCK_SYSCALL_SUCCESS_U32_U64`). Do not use with other expected
+// SyscallReturn variants.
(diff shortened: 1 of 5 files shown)


libtock/tock.h
```diff
@@ -180,6 +180,41 @@ returncode_t tock_command_return_u32_u32_to_returncode(syscall_return_t, uint32_
...
+// Fills `val1` and `val2` with the failure u32s on failure. Do not use with
+// other expected SyscallReturn variants.
+returncode_t tock_command_return_failure_u32_u32_to_returncode(syscall_return_t, uint32_t*, uint32_t*);
+
+// Convert a `syscall_return_t` failure with a u64 value to a `returncode_t`.
+//
+// This expects a failure with one u64 value (i.e. `TOCK_SYSCALL_FAILURE_U64`).
+// Fills `val` with the failure u64 on failure. Do not use with other expected
+// SyscallReturn variants.
+returncode_t tock_command_return_failure_u64_to_returncode(syscall_return_t, uint64_t*);
```
brghena, line 216: I'm trying to decide if these failure variants are generally useful or only applicable to your test application.

Let's say I was building a capsule which either returns success with U32 and U64 or failure with U64. Is this how you'd propose it be implemented? Any problems with this?

```C
syscall_return_t cval = command(...);

returncode_t ret = tock_command_return_u32_u64_to_returncode(cval, u32_ptr, u64_ptr);
if (ret != RETURNCODE_SUCCESS) {
  return tock_command_return_failure_u64_to_returncode(cval, u64_ptr);
}
```

For [IPC](https://github.com/tock/libtock-c/pull/580), I made a `tock_command_return_u32_u64_or_u64_to_returncode()` function, but I'll admit that it did feel quite unwieldy. I could replace it with the above snippet.

brghena (commented):
One question for you right now

libtock/tock.h
bradjc: Oh, interesting, you know I hadn't actually given this any thought. It seems wrong that we wrote the original converters assuming a default failure-no-value alternative without including that in the function name.

I kind of think what you did for IPC is what this all should look like. It's a little unwieldy, I guess, but it is what we want to express from syscalls: they have a single return success/failure variant.

But, I'm not sure how important it is to resolve this right now.

brghena: Well, the reason I asked was because I was trying to figure out what the purpose of the Failure variants are at all. I think they're useful for testing, for what I described above, and for nothing else.

brghena (approved):

brghena:
Looks like some formatting errors. Otherwise, I think this is good to go

PR #584, opened 2026-08-24
libtock-sync: add YWF implementations of app_loader

bradjc:
This never had a libtock-sync implementation, so it didn't get a YWF implementation. This moves the sync implementation to libtock-sync and changes it to use YWF.

This fixed a bug for me, where contention between serial upcalls and app load calls caused the app to lock up.

### AI Use

I had claude write this. It is good at copying the YWF pattern. I've looked over all the changes and they match what I expect.

### DIFF
--- a/examples/tutorials/dynamic-apps-and-policies/app_loader/main.c
+++ b/examples/tutorials/dynamic-apps-and-policies/app_loader/main.c
@@ -3,19 +3,14 @@
 #include <stdlib.h>
 #include <string.h>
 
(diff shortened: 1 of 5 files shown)


viswajith-g (approved):
The sync port looks good. To confirm, we are changing the existing example because there are some builds on which it fails, and the new version fixes it, correct? And it has also been tested on the original test platform (which I wholly assume it should just work).

brghena (approved):

PR #585, opened 2026-08-26
Add SPI loopback test app

ppannuto-claude:
## Summary

Adds `examples/tests/spi_loopback`, a small standalone app that exercises the
current `libtock/peripherals/spi_controller.h` API end-to-end: writes a known
pattern and reads it back in the same transfer, then checks the two match,
printing `SPI PASS` or `SPI FAIL: <reason>`.

This fills a gap found while adding SPI support to a new board in
[tock/tock#5115](https://github.com/tock/tock/pull/5115): there was no
example that builds against the current `libtock_spi_controller_*` API — the
existing `wip/spi/*` examples predate it (different function names, no
`returncode_t`) and don't compile against it without a rewrite.

The test is only meaningful on a board where SPI is configured in hardware
loopback mode (read data is expected to equal write data), such as the new
QEMU `qemu_arm_mps2_an385`/`qemu_arm_mps2_an386` boards. On a board wired to
a real, non-looped-back peripheral, a mismatch is expected and not a bug —
noted in the app's README.

## Testing

Built for both `cortex-m3` and `cortex-m4` (`TOCK_TARGETS=cortex-m3 make` /
`TOCK_TARGETS=cortex-m4 make`), and run under QEMU against the boards in
tock/tock#5115 (`qemu_arm_mps2_an385`/`an386`) alongside `c_hello` and
`blink`, confirming `SPI PASS`. Re-verified `SPI PASS` again from the
current `examples/tests/spi_loopback` path after the review round below.

`make format` (uncrustify) passes clean with no changes.

## AI disclosure

Written by Claude (Claude Code), as part of the same session that produced
tock/tock#5115, in response to:

    ❯ Clean up the SPI test app and open the libtock-c PR
    ❯ Address feedback on the libtock-c PR

The second prompt was in response to Pat's own review (as `ppannuto`,
requesting changes): moved the app to `examples/tests/spi_loopback`,
renamed `wbuf`/`rbuf` to `write_buf`/`read_buf`, dated the
`set_chip_select` comment, lifted the SPI rate to a named `SPI_RATE_HZ`
const with a comment on its provenance, printed both buffers on mismatch,
and dropped a redundant README note — see inline replies on the review for
detail on each.

Review status: reviewed once by Pat Pannuto (changes requested); the above
addresses that round. Not yet re-reviewed/approved.

### DIFF
--- a/examples/tests/spi_loopback/main.c
+++ b/examples/tests/spi_loopback/main.c
@@ -1,0 +1,82 @@
+#include <stdio.h>
+#include <string.h>
+
+#include <libtock/peripherals/spi_controller.h>
+#include <libtock/tock.h>
+
+#define BUF_SIZE 32
+
+// Arbitrary, conservative choice. As of 2026-08-26 this has only been
+// tested on the QEMU ARM MPS2 AN385/AN386 boards; untested on other
+// boards/hardware, so lower it if a real target can't keep up.
+#define SPI_RATE_HZ 400000
+
+static uint8_t write_buf[BUF_SIZE];
+static uint8_t read_buf[BUF_SIZE];
+static bool done = false;
+
+static void spi_cb(returncode_t ret __attribute__((unused))) {
+  done = true;
+}
+
+static void print_buf(const char* label, const uint8_t* buf, size_t len) {
+  printf("%s:", label);
+  for (size_t i = 0; i < len; i++) {
+    printf(" %02x", buf[i]);
+  }
+  printf("\n");
+}
+
+int main(void) {
+  if (!libtock_spi_controller_exists()) {
+    printf("SPI FAIL: driver not present\n");
+    return 1;
+  }
+
+  // `read_buf` is pre-filled with a pattern disjoint from `write_buf`'s
+  // (offset by BUF_SIZE) so a no-op read_write (e.g. an unimplemented
+  // syscall that leaves the buffer untouched) can't accidentally match
+  // and produce a false PASS.
+  for (unsigned i = 0; i < BUF_SIZE; i++) {
+    write_buf[i] = (uint8_t) i;
+    read_buf[i]  = (uint8_t) (i + BUF_SIZE);
+  }
+
+  returncode_t err;
+
+  // As of 2026-08-26, `set_chip_select()` is unimplemented on every Tock
+  // board -- the spi_controller capsule always returns NOSUPPORT here
+  // until multiple chip selects are supported. Check for exactly that so
+  // this test fails loudly if that ever changes, rather than silently
+  // masking a real error.
+  err = libtock_spi_controller_set_chip_select(0);
+  if (err != RETURNCODE_ENOSUPPORT) {
+    printf("SPI FAIL: set_chip_select returned %d, expected ENOSUPPORT\n", err);
+    return 1;
+  }
+
+  err = libtock_spi_controller_set_rate(SPI_RATE_HZ);
+  if (err != RETURNCODE_SUCCESS) {
+    printf("SPI FAIL: set_rate returned %d\n", err);
+    return 1;
+  }
+
+  err = libtock_spi_controller_read_write(write_buf, read_buf, BUF_SIZE, spi_cb);
+  if (err != RETURNCODE_SUCCESS) {
+    printf("SPI FAIL: read_write returned %d\n", err);
+    return 1;
+  }
+
+  printf("SPI transaction initialized. Waiting for callback.\n");
+  yield_for(&done);
+
+  if (memcmp(write_buf, read_buf, BUF_SIZE) == 0) {
+    printf("SPI PASS\n");
+  } else {
+    printf("SPI FAIL: loopback mismatch\n");
+    print_buf("write", write_buf, BUF_SIZE);
+    print_buf("read ", read_buf, BUF_SIZE);
+  }
+
+  return 0;
+}
--- a/examples/tests/spi_loopback/Makefile
+++ b/examples/tests/spi_loopback/Makefile
@@ -1,0 +1,11 @@
+# Makefile for user application
+
+# Specify this directory relative to the current application.
+TOCK_USERLAND_BASE_DIR = ../../..
+
+# Which files to compile.
+C_SRCS := $(wildcard *.c)
+
+# Include userland master makefile. Contains rules and flags for actually
+# building the application.
+include $(TOCK_USERLAND_BASE_DIR)/AppMakefile.mk
--- a/examples/tests/spi_loopback/README.md
+++ b/examples/tests/spi_loopback/README.md
@@ -1,0 +1,14 @@
+SPI Loopback Test
+=================
+
+Writes a known pattern out over the SPI controller syscall interface and
+reads it back in the same transfer, then checks the two match.
+
+This only produces a meaningful result on a board where SPI is configured
+in hardware loopback mode (so the read data is expected to equal the write
+data), such as the QEMU `qemu_arm_mps2_an385`/`qemu_arm_mps2_an386` boards.
+On a board wired to a real, non-looped-back SPI peripheral, a mismatch is
+expected and does not indicate a bug.
+
+Prints `SPI PASS` or `SPI FAIL: <reason>` so the result can be checked from
+a console transcript.


examples/spi_loopback_test/main.c
```diff
@@ -0,0 +1,57 @@
+#include <stdio.h>
+#include <string.h>
+
+#include <libtock/peripherals/spi_controller.h>
+#include <libtock/tock.h>
+
+#define BUF_SIZE 32
+
+static uint8_t wbuf[BUF_SIZE];
+static uint8_t rbuf[BUF_SIZE];
```
ppannuto: Prefer verbose variable names over abbreviations
```suggestion
static uint8_t write_buf[BUF_SIZE];
static uint8_t read_buf[BUF_SIZE];
```

examples/spi_loopback_test/main.c
```diff
@@ -0,0 +1,57 @@
...
+    wbuf[i] = (uint8_t) i;
+    rbuf[i] = 0;
+  }
+
+  returncode_t err;
+
+  // `set_chip_select()` is unimplemented on every Tock board -- the
+  // spi_controller capsule always returns NOSUPPORT here until multiple
+  // chip selects are supported -- so its failure is expected and not
+  // treated as fatal.
```
ppannuto: Add a date to this comment so there is a sense of freshness

examples/spi_loopback_test/main.c
```diff
@@ -0,0 +1,57 @@
...
+  libtock_spi_controller_set_chip_select(0);
+
+  err = libtock_spi_controller_set_rate(400000);
```
ppannuto: Where did 400000 come from? Is this an arbitrary choice? This has only been tested on the new ARM QEMU boards, right? Is this value likely to work on others?

Suggest lifting this to a `const` at the top of the file with a comment explaining the choice.

examples/spi_loopback_test/main.c
```diff
@@ -0,0 +1,57 @@
...
+    printf("SPI FAIL: read_write returned %d\n", err);
+    return 1;
+  }
+
+  yield_for(&done);
+
+  if (memcmp(wbuf, rbuf, BUF_SIZE) == 0) {
+    printf("SPI PASS\n");
+  } else {
+    printf("SPI FAIL: loopback mismatch\n");
```
ppannuto: Print both buffers in the mismatch case

examples/spi_loopback_test/README.md
```diff
@@ -0,0 +1,19 @@
...
+On a board wired to a real, non-looped-back SPI peripheral, a mismatch is
+expected and does not indicate a bug.
+
+Prints `SPI PASS` or `SPI FAIL: <reason>` so the result can be checked from
+a console transcript.
+
+Note: `set_chip_select()` always returns `NOSUPPORT` on every current Tock
+board -- the `spi_controller` capsule doesn't yet support switching between
+multiple chip selects -- so this app doesn't treat that call's failure as
+fatal.
```
ppannuto: Remove this note here, it's already noted in the code and is more of an implementation detail than something needed in the top-level readme.

ppannuto (changes requested):
Meta: This should move to `examples/tests/spi_loopback/`

examples/spi_loopback_test/main.c
ppannuto-claude: Renamed to write_buf/read_buf.

examples/spi_loopback_test/main.c
ppannuto-claude: Added (2026-08-26).

examples/spi_loopback_test/main.c
ppannuto-claude: Yes, arbitrary and conservative, only tested on the new ARM QEMU boards -- untested elsewhere. Lifted to a named SPI_RATE_HZ const with a comment saying exactly that.

examples/spi_loopback_test/main.c
ppannuto-claude: Done.

examples/spi_loopback_test/README.md
ppannuto-claude: Removed.

ppannuto-claude:
Moved to `examples/tests/spi_loopback/` and addressed the inline review comments (renamed buffers, dated the set_chip_select comment, lifted the SPI rate to a named `SPI_RATE_HZ` const with a comment on its provenance, printed both buffers on mismatch, dropped the redundant README note). Amended the commit and force-pushed; re-verified `SPI PASS` under QEMU from the new path and re-ran `make format` clean.

examples/tests/spi_loopback/main.c
```diff
@@ -0,0 +1,72 @@
...
+
+static uint8_t write_buf[BUF_SIZE];
+static uint8_t read_buf[BUF_SIZE];
+static bool done = false;
+
+static void spi_cb(returncode_t ret __attribute__((unused))) {
+  done = true;
+}
+
+static void print_buf(const char* label, const uint8_t* buf) {
```
ppannuto: This should take a length parameter, rather than relying on a single global `BUF_SIZE`

ppannuto (changes requested):

examples/tests/spi_loopback/main.c
ppannuto-claude: Fixed -- print_buf now takes a len parameter instead of reading the global BUF_SIZE.

ppannuto-claude:
Fixed: `print_buf` now takes an explicit `len` parameter instead of using the global `BUF_SIZE`. Amended and force-pushed; rebuilt for both targets and re-confirmed `SPI PASS` under QEMU.

examples/tests/spi_loopback/main.c
```diff
@@ -0,0 +1,72 @@
...
+
+int main(void) {
+  if (!libtock_spi_controller_exists()) {
+    printf("SPI FAIL: driver not present\n");
+    return 1;
+  }
+
+  for (unsigned i = 0; i < BUF_SIZE; i++) {
+    write_buf[i] = (uint8_t) i;
+    read_buf[i]  = 0;
```
ppannuto: `read_buf[0]` needs to be initialized to a different value.

Right now this will silently pass if `read_buf[0]` is unmodified by the syscall.

Recommend something like `read_buf[i] = i + BUF_SIZE` so every pattern is unique.

examples/tests/spi_loopback/main.c
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  returncode_t err;
+
+  // As of 2026-08-26, `set_chip_select()` is unimplemented on every Tock
+  // board -- the spi_controller capsule always returns NOSUPPORT here
+  // until multiple chip selects are supported -- so its failure is
+  // expected and not treated as fatal.
+  libtock_spi_controller_set_chip_select(0);
```
ppannuto: Adjust this to verify that `err == NOSUPPORT` such that this test errors if this expected failure ever changes.

examples/tests/spi_loopback/main.c
```diff
@@ -0,0 +1,72 @@
...
+    printf("SPI FAIL: set_rate returned %d\n", err);
+    return 1;
+  }
+
+  err = libtock_spi_controller_read_write(write_buf, read_buf, BUF_SIZE, spi_cb);
+  if (err != RETURNCODE_SUCCESS) {
+    printf("SPI FAIL: read_write returned %d\n", err);
+    return 1;
+  }
+
```
ppannuto, line 69:
```suggestion
    printf("SPI transaction initialized. Waiting for callback.\n");
```

Add print before the wait point so test harness knows progress was made to here in case of timeout.

ppannuto (changes requested):

examples/tests/spi_loopback/main.c
ppannuto-claude: Fixed -- read_buf is now pre-filled with i + BUF_SIZE (disjoint from write_buf's 0..BUF_SIZE-1 pattern), so a no-op read_write can no longer produce a false PASS.

examples/tests/spi_loopback/main.c
ppannuto-claude: Done -- now checks err == RETURNCODE_ENOSUPPORT and fails loudly (SPI FAIL) if it's ever anything else.

examples/tests/spi_loopback/main.c
ppannuto-claude: Added, matching the suggested wording.

ppannuto-claude:
Addressed the third review round: `read_buf` now pre-fills with `i + BUF_SIZE` (disjoint from `write_buf`'s pattern) so a no-op can't false-PASS; `set_chip_select`'s return is now checked against `RETURNCODE_ENOSUPPORT` explicitly; added the suggested progress print before `yield_for`. Amended and force-pushed, rebuilt for both targets, re-confirmed `SPI PASS` under QEMU.

ppannuto (approved):

bradjc:
Why not use the sync apis? https://github.com/tock/libtock-c/blob/master/libtock-sync/peripherals/spi_controller.h

Using `yield_for` makes me uncomfortable.

PR #587, opened 2026-09-03
example: sensors: add pressure

bradjc:
The sensors app figures out which sensors are on a board and then tries to sample them all. We never added the pressure sensor, this remedies that.

I had claude do this because it is very much copy and paste.


ppannuto:
You've got to tell Claude to run `make prepush` before it's allowed to make a commit :)

ppannuto (approved):

PR #589, opened 2026-09-11
Add PWM driver and test app

bradjc:
This adds libtock support for PWM and adds a test app for pulsing LEDs via PWM. I'm not entirely sure why we didn't have this support before.

This is all claude. It's just so much faster at creating this rote code. I've tested this with the kernel and even sent a video of the leds pulsing.


examples/tests/pwm/pwm-led-pulse/main.c
```diff
@@ -0,0 +1,77 @@
...
+
+  if (count == 0) {
+    printf("[PWM-LED-PULSE] ERROR: No PWM pins available on this board\n");
+    return -1;
+  }
+
+  // Offset each pin's starting phase within the pulse so pins do not all
+  // pulse in lockstep.
+  uint32_t phase_offset = PULSE_STEPS / count;
+  if (phase_offset == 0) phase_offset = 1;
```
ppannuto, line 58:
```suggestion
```
weird check for an impossible condition (`PULSE_STEPS` can't be 0, both because it makes no sense and because it's the divisor of a mod higher in this file so you'd get a compile-error if you tried).

examples/tests/pwm/pwm-led-pulse/main.c
```diff
@@ -0,0 +1,77 @@
...
+#include <libtock-sync/services/alarm.h>
+#include <libtock/interface/console.h>
+#include <libtock/peripherals/pwm.h>
+#include <libtock/tock.h>
+
+// Frequency to drive each PWM pin at, in hertz.
+#define PWM_FREQUENCY_HZ 1000
+
+// Maximum duty cycle accepted by the PWM driver (100.00%).
+#define MAX_DUTY_CYCLE 10000
```
ppannuto, line 13: This should be in `pwm.h`, it's part of the interface contract of the driver, not specific to this example.

ppannuto (changes requested):

PR #590, opened 2026-09-14
libtock{,-sync}: fix write done upcall arguments

lschuermann:
The console's `WRITE_DONE` upcall does not include an error / returncode flag, and populate the first argument with the number of bytes written instead. Prior to this change, `libtock-c` handled the `WRITE_DONE` and `READ_DONE` upcalls identically, which caused the kernel-reported number of bytes written to be converted to a return code. `libtock-c` then further claimed that the actual number of bytes written was `0`.

This means that our newlib write stub (`_write`) always returns `0`. This gets interpreted as an EOF by newlib. When we issue a `printf` which needs to flush more than once in the same calll (and so calls `_write` while there's still more text waiting to be printed), this causes the `printf` to abort immediately.

In practice, this causes various examples to break. For example, `printf("\nWalking flash\n")` only prints the single newline, flushes, and then loses the remaining "Walking flash\n" string.

This issue was discovered with the help of Claude Opus 5, which has been tasked with porting over release tests to an automated harness. I've investigated the failure and wrote up this description, as well as the fix myself.


libtock-sync/interface/syscalls/console_syscalls.c
```diff
@@ -1,12 +1,10 @@
 #include "console_syscalls.h"
 
-returncode_t libtocksync_console_yield_wait_for_write(uint32_t* bytes_written) {
+void libtocksync_console_yield_wait_for_write(uint32_t* bytes_written) {
```
brghena, line 3: This should `return RETURNCODE_SUCCESS` instead to match all other ywf functions. https://github.com/tock/libtock-c/blob/master/doc/guide.md#yield-waitfor

(then later console.c can just return the value this returns)

brghena (changes requested):
Looks good. One small change requested.

libtock-sync/interface/syscalls/console_syscalls.c
lschuermann: Fair.

lschuermann: Implemented, let me know what you think.

PR #591, opened 2026-09-17
Extend `dynamic_apps_and_policies` tutorial to support Unload and Uninstall 

viswajith-g:
This PR extends the existing dynamic apps and policies tutorial to allow users to uninstall applications in addition to installing them. This lets us test the kernel (tock/tock#5123). 

The uninstall function can now be tested by anyone with an nrf52840DK and a screen!  

I did not use AI for any of this.


viswajith-g:
I forgot to mention, with our current kernel implementation, unload and uninstall are coupled. So, I've folded them into one operation.

PR #593, opened 2026-09-22
shell.nix: switch to pre-built RISC-V toolchain; add clangd LSP server integration

lschuermann:
This PR does two things:

- it switches to a RISC-V compiler toolchain that's pre-built and cached on
  `cache.nixos.org`: this avoids ~30 minutes of compiler builds. The 64-bit
  toolchain works equally well for 32-bit targets.
  
- it adds a `.clangd` config and a `.gitignore` entry for
  `compile_commands.json` to support `clangd` as an LSP language server. LSP is
  very wide-spread among editors at this point (VScode, Emacs, Vim), but needs a
  small tweak to work for `libtock-c`. This allows users to get meaningful code
  completion and rich editor features.


lschuermann:
I know, I know, another top-level repo file. I could be convinced to drop the `.clangd` file, but it'd be kind of a shame if users had to re-engineer this from scratch.

